PR #21, opened 2026-06-20
Fix global session listing and activity time

ergs0204:
## Summary

This fixes session list behavior when `opencode serve` is launched from a broad/default directory such as `~` and sessions exist in other project folders.

The previous remote UI only listed `/session` from the server's default workspace. That could hide sessions created in selected folders, or switch to the detail view after creation without a usable selected session in app state.

This PR changes the session list flow to:

- discover sessions across projects with `/experimental/session` when available
- fall back to `/session` for older OpenCode servers
- hydrate discovered sessions from scoped `/session?directory=...` lists so card metadata such as summaries comes from the project-scoped endpoint
- query statuses per session directory with `/session/status?directory=...`
- keep a newly created session in local state if refresh does not immediately return it

It also fixes the displayed/sorted "Updated" timestamp. The app previously used `session.time.updated`, but OpenCode can bump that row timestamp for non-chat session metadata/maintenance. That made old sessions, for example an unused `/home/opc/kesto` session, appear as updated today. The session list now uses the latest message timestamp when available, falling back to the session row timestamp only when no message exists or the lookup fails.

## Details

- Added `api.listGlobalSessions()` for `/experimental/session`.
- Added `api.loadLatestMessage()` for a cheap `/session/:id/message?limit=1` activity lookup.
- Added global session hydration from per-directory scoped session lists.
- Added a small in-memory latest-message timestamp cache keyed by session id and server `session.time.updated`.
- Extended the `Session` type for optional global-session `project` metadata.
- Added regression assertions covering global discovery, scoped hydration, and latest-message activity timestamps.

## Verification

Local verification passed:

- `npm run test:ui`
- `npm run test:model`
- `npm run test:settings`
- `npm run build`

GitHub Actions signed APK build passed:

- https://github.com/ergs0204/opencode-remote-android/actions/runs/27855606991

Tested manually on Android after the signed APK build.

### DIFF
--- a/web/src/App.tsx
+++ b/web/src/App.tsx
@@ -3,7 +3,7 @@ import ReactMarkdown from "react-markdown"
 import remarkGfm from "remark-gfm"
 import { api } from "./api"
 import { createTranslator, languageOptions, normalizeLanguage, type LanguageCode } from "./i18n"
-import type { CommandInfo, DiffFile, FileEntry, FileStatusEntry, MessageEnvelope, ModelOption, ModelSelection, PathInfo, ProjectDashboard, ServerConfig, Session, SessionView, TodoItem } from "./types"
+import type { CommandInfo, DiffFile, FileEntry, FileStatusEntry, MessageEnvelope, ModelOption, ModelSelection, PathInfo, ProjectDashboard, ServerConfig, Session, SessionStatus, SessionView, TodoItem } from "./types"
 import {
   SettingsIcon,
   FolderIcon,
@@ -112,13 +112,17 @@ function isProjectDirectory(pathInfo: PathInfo): boolean {
   return pathInfo.worktree !== "/"
 }
 
-function toSessionView(session: Session): SessionView {
+function messageActivityTime(message: MessageEnvelope): number {
+  return Math.max(message.info.time.created, message.info.time.completed ?? 0)
+}
+
+function toSessionView(session: Session, status?: SessionStatus, activityTime = session.time.updated): SessionView {
   return {
     id: session.id,
     title: session.title,
     directory: session.directory,
-    updated: session.time.updated,
-    status: "idle",
+    updated: activityTime,
+    status: status?.type ?? "idle",
     files: session.summary?.files ?? 0,
     additions: session.summary?.additions ?? 0,
     deletions: session.summary?.deletions ?? 0,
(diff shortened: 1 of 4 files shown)


gervaso-assistant (changes requested):
Thanks for the PR — the direction looks good and I could reproduce why this is useful: when `opencode serve` is launched from a broad/default directory, `/session` misses project-scoped sessions while `/experimental/session` finds them.

I tested the PR locally on top of latest `upstream/main` (`99b8448`) and the checks passed:

- `npm run test:ui`
- `npm run test:model`
- `npm run test:settings`
- `npm run test:i18n`
- `npm run build`
- `npx cap sync android`
- `./gradlew assembleDebug`

Before merging, I think two small follow-ups are needed:

1. **Global session discovery should page through results.**
   The live OpenCode `/doc` for `/experimental/session` exposes `limit`, `cursor`, and `start`. `api.listGlobalSessions()` currently calls `/experimental/session` once, so installs with many sessions can still hide anything beyond the first page/default limit.

2. **Do not cache failed latest-message lookups.**
   `loadSessionActivityTimes()` converts a failed `/session/:id/message?limit=1` request into `[]`, then caches `session.time.updated` for that session/update pair. A transient failure or temporarily bad directory can therefore keep the old mutable row timestamp until the session row changes again. Better to only cache successful latest-message responses, or mark failures as retryable on the next refresh.

With those fixed, I’m happy with the approach.

ergs0204:
Updated the PR for the two review follow-ups.

Changes:

- `api.listGlobalSessions()` now follows `/experimental/session` pagination using `x-next-cursor`, so it does not stop at the first/default page.
- Response headers are normalized for both browser fetch and Capacitor native HTTP, so cursor handling works in APK/native mode too.
- Failed `loadLatestMessage()` requests are no longer cached as `session.time.updated`; failures fall back for the current render only and retry on the next refresh.
- Added regression assertions for global pagination and retryable latest-message failures.

gervaso-assistant (approved):
Tested updated PR on latest main and verified the previous blockers are fixed.

Checks run locally:
- npm ci
- npm run test:ui
- npm run test:model
- npm run test:settings
- npm run test:i18n
- npm run build
- npx cap sync android
- ./gradlew assembleDebug

Android APK smoke-tested by Giulio and confirmed working. Ready to merge.

PR #30, opened 2026-07-24
fix: don't close the app when pressing back from an open session

birabittoh:
I kept closing this app by mistake so I made this fix out of spite.

If the user is in the "Sessions" tab, going back will close the app. In all other cases, the app will go back to the "Sessions" tab instead of closing.

### DIFF
--- a/web/package-lock.json
+++ b/web/package-lock.json
@@ -9,6 +9,7 @@
       "version": "1.5.0",
       "dependencies": {
         "@capacitor/android": "^8.3.4",
+        "@capacitor/app": "^8.1.1",
         "@capacitor/cli": "^8.3.4",
         "@capacitor/core": "^8.3.4",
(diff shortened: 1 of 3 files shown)


giuliastro:
Landed in #35 — thanks @birabittoh, this was a real annoyance and you were right to fix it.

I reworked three things rather than merging directly:

- `exitApp()` was called inside the `setView` updater. Updaters have to be pure because React can invoke them more than once; under StrictMode that meant `exitApp` could fire twice.
- The cleanup captured `listenerHandle` before `addListener` resolved, so a listener registered after teardown was leaked. It now removes itself either way.
- Back also dismisses the delete dialog, an inline rename, or an open detail sheet before it changes view — otherwise back jumped straight past an open sheet.

Closing this since the change is in `main`.

PR #31, opened 2026-07-24
feat: allow queuing prompts while a session is working

birabittoh:
Keep the composer enabled and the send button available while the assistant is busy, so a new message can be queued instead of forcing the user to wait or only offering to stop. The stop button still shows whenever the composer is empty while the session is working.

### DIFF
--- a/web/src/App.tsx
+++ b/web/src/App.tsx
@@ -1647,19 +1647,17 @@ function App() {
               onKeyDown={(event) => {
                 if (event.key === "Enter" && !event.shiftKey) {
                   event.preventDefault()
-                  if (!isWorking) {
-                    send().catch(() => undefined)
-                  }
+                  send().catch(() => undefined)
                 }
               }}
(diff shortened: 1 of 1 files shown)


giuliastro:
Landed in #35 — thanks @birabittoh. The composer being dead while the agent worked was the right thing to go after.

One catch worth flagging: as written this broke the OMP backend, which landed after you opened this. ACP allows one turn per session, and the bridge rejected a second prompt with `The OMP session is already running` — so on OMP the user would have got a red error exactly where they expected a queued message.

So #35 makes the queue real on the bridge side: the prompt is recorded immediately so it is visible while it waits, the session keeps reporting `busy` between turns, a queued model change is applied on dequeue rather than under the running turn, and stopping the session discards the queue along with the messages recorded for it. Your composer change is in as-is, with `showStopAction` keeping stop available only while there is nothing to send.

Closing this since the change is in `main`.

PR #32, opened 2026-07-25
feat: major UI overhaul

birabittoh:
## Summary

Includes both #30 and #31 in order to avoid merge conflicts, so you might want to handle those before this one. Major refactor of the entire UI: streamed message rendering, grouped/formatted agent actions (tool calls, diffs, reasoning, to-dos, questions), interactive question answering, and several session-list and layout fixes.

It's a lot of commits so you might wanna squash and merge. These are substantial changes so I'm open to feedback and willing to make more changes in order to get it merged.

- Stream assistant replies token-by-token instead of waiting for the full message, and keep the message list correctly pinned to the bottom while it grows (without fighting a user who's scrolled away).
- Group each turn's thinking/tool-calls/edits into a single collapsible "action" row (e.g. *"Thought for 12s, read 3 files, edited 2 files"*) instead of showing every tool call inline, with a modal for full detail.
- Render tool calls, diffs, reasoning, to-do updates, and question prompts with purpose-built, readable formatting instead of raw JSON/command dumps — including colored unified diffs with a expandable full-diff modal.
- Let the user answer the assistant's interactive questions (`AskUserQuestion`) directly from the chat.
- Localize all of the newly added strings across English, Italian, and Traditional Chinese.
- Assorted layout/mobile fixes

## Changes

- **Streaming & rendering**: apply streamed `message.part` events for gradual token rendering; stop periodic refresh from wiping in-progress reasoning; never let a reasoning update shrink already-shown thinking text; keep pinned to the latest message while opening a session and while content is still loading.
- **Diffs**: render colored unified diffs for patch parts; truncate long diff/tool previews inline and expand to a full modal on tap; don't wrap long lines in tool call command/output.
- **Action grouping**: group a turn's reasoning/tool/patch parts into a single summarized row; give `question` and `todowrite` tool calls dedicated, readable summaries and detail views instead of raw JSON.
- **Questions**: let the user answer the assistant's `AskUserQuestion` prompts from the chat UI.
- **Queueing**: allow queuing prompts while a session is working (`queue` branch).
- **Session list**: keep rename/delete on one row on mobile; simplify session card actions to just Rename and Delete; move the session status pill onto the stats row; don't close the app on Android back-press from an open session (`main` branch); navigate to a newly created session immediately.
- **Layout**: top nav bar scrolls with the page instead of staying pinned; composer send button stays beside the composer on mobile and drops its redundant text label; modal spacing, stray divider, and chat horizontal overflow fixes; live-updates status text sits next to the connection status instead of on its own line.
- **i18n**: add and localize all strings introduced by the above (agent action labels/counts, question/todo formatting, modal close button) across `en`, `it`, and `zh-TW`.
- **Polish**: truncate long tool-call modal titles instead of dumping the full command into the title bar; capitalize the joined action-group summary regardless of which piece (reasoning vs. tool count) comes first.
- **Performance**: memoize the message list (`MessagesPane`) and each message (`MessageArticle`) so composer typing and unrelated state updates no longer re-render the whole conversation; keep unchanged messages referentially stable across streamed part updates and the periodic refetch (`mergeFetchedMessages`) instead of allocating a fresh object per message every time, so per-message rendering/diffing only reruns for messages that actually changed; hoist `remarkPlugins` to a stable reference; stop the 3.5s polling fallback from doing a redundant full refetch while the SSE event stream is already live.

## TODO

`App.tsx` needs to be split into multiple, independent component files. Let me know if it's in scope for this issue or if you want to handle it later.

### DIFF
--- a/web/package-lock.json
+++ b/web/package-lock.json
@@ -9,6 +9,7 @@
       "version": "1.5.0",
       "dependencies": {
         "@capacitor/android": "^8.3.4",
+        "@capacitor/app": "^8.1.1",
         "@capacitor/cli": "^8.3.4",
         "@capacitor/core": "^8.3.4",
         "react": "18.3.1",
@@ -31,6 +32,15 @@
       "license": "MIT",
       "peerDependencies": {
         "@capacitor/core": "^8.3.0"
+      }
+    },
+    "node_modules/@capacitor/app": {
+      "version": "8.1.1",
+      "resolved": "https://registry.npmjs.org/@capacitor/app/-/app-8.1.1.tgz",
+      "integrity": "sha512-xM2ZTX5jK60tFtmjsmJv+Cdj1Qsb6NcavkqmdEnCPQBeya9oKhamZJxYOnQNj1ZvcJM7sWfG6BEtSLx42pisbA==",
+      "license": "MIT",
+      "peerDependencies": {
+        "@capacitor/core": ">=8.0.0"
       }
     },
     "node_modules/@capacitor/cli": {
--- a/web/package.json
+++ b/web/package.json
@@ -18,6 +18,7 @@
   },
   "dependencies": {
     "@capacitor/android": "^8.3.4",
+    "@capacitor/app": "^8.1.1",
     "@capacitor/cli": "^8.3.4",
     "@capacitor/core": "^8.3.4",
     "react": "18.3.1",
--- a/web/src/App.tsx
+++ b/web/src/App.tsx
@@ -1,4 +1,6 @@
 import { useEffect, useMemo, useRef, useState } from "react"
+import { App as CapacitorApp } from "@capacitor/app"
+import type { PluginListenerHandle } from "@capacitor/core"
 import ReactMarkdown from "react-markdown"
 import remarkGfm from "remark-gfm"
 import { api } from "./api"
@@ -906,6 +908,24 @@ function App() {
   useEffect(() => {
     localStorage.setItem(LANGUAGE_STORAGE_KEY, language)
   }, [language])
+
+  useEffect(() => {
+    let listenerHandle: PluginListenerHandle | undefined
+    CapacitorApp.addListener("backButton", () => {
+      setView((current) => {
+        if (current === "sessions") {
+          CapacitorApp.exitApp()
+          return current
+        }
+        return "sessions"
+      })
+    }).then((handle) => {
+      listenerHandle = handle
+    })
+    return () => {
+      listenerHandle?.remove()
+    }
+  }, [])
 
   useEffect(() => {
     const mediaQuery = window.matchMedia("(prefers-color-scheme: dark)")
--- a/web/src/App.tsx
+++ b/web/src/App.tsx
@@ -1647,19 +1647,17 @@ function App() {
               onKeyDown={(event) => {
                 if (event.key === "Enter" && !event.shiftKey) {
                   event.preventDefault()
-                  if (!isWorking) {
-                    send().catch(() => undefined)
-                  }
+                  send().catch(() => undefined)
                 }
               }}
-              disabled={!selectedSession || isWorking}
-            />
-            <button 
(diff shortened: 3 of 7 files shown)


giuliastro:
Thanks for this @birabittoh — the streamed rendering and the grouped tool/diff/reasoning output are exactly where this app needed to go. I'd like it in, but `main` moved a long way under it today and I think the rebase is better done by you than by me. Here's the full picture so it's a mechanical job rather than an archaeology one.

## What changed in `main`

Released as [v2.0.0](https://github.com/giuliastro/harness-remote/releases/tag/v2.0.0): the project is now **Harness Remote**, the app is harness-agnostic, and **Oh My Pi** joined OpenCode as a selectable backend through a local ACP bridge (`bridge/`). #30 and #31 also landed, reworked, in #35 — so you can drop those commits from this branch.

## The conflicts

Merging this into `main` today conflicts in `web/src/App.tsx` only, 10 hunks. Four of them sit on top of things that must survive:

| Hunk | In `main` | Please keep |
|---|---|---|
| 2 | `defaultConfig(backend)` and the per-backend storage keys | the whole multi-backend config layer |
| 4 | `loadedMessagesRef` snapshot logic | replaces `setMessages((current) => …)`; guards against a background refresh clobbering streamed output |
| 5 | `isChatNearBottom` | you have `isNearMessagesBottom` doing the same job — pick one |
| 7 | the autosave effect | **the important one, see below** |

Hunks 9 and 10 are the session-card actions and the messages pane, where `main` hides rename/delete for OMP and you extract `MessagesPane`.

## Please don't lose this one

Hunk 7 lands on the fix for a bug that could brick the app:

```js
if (draftConfig.host.trim() && !isValidServerConfig(draftConfig)) return
```

Settings now save automatically. Typing a host such as `http://192.168.1.64` one character at a time passes through `http:` and `http://`, which cannot be turned into a URL — `streamURL` threw inside a render effect, React unmounted the tree, and the half-typed host had already been persisted, so every later launch reproduced the empty root. The only way out was clearing app data. Resolving hunk 7 toward this branch brings that back.

Related, and please keep: the five `isValidServerConfig(config)` guards, the `ErrorBoundary` in `main.tsx`, and `web/src/server-config.test.mjs` (`npm run test:config`), which fails if the guard goes missing.

## Work with no conflict markers

This branch predates OMP, so nothing in it knows that backend exists. After the 10 textual conflicts there is still semantic work:

- **hunk 3** adds a `questions` fetch to `loadSelected`. The OMP bridge does not implement that endpoint — it needs gating on `config.backend`, like the existing `config.backend !== "omp"` checks around agents, rename/delete and diff.
- same for any other new call the overhaul introduces: OMP has no agents, no server slash commands, no VCS/diff, and no interactive questions.
- `web/src/i18n.ts` gained a `settings.hostHint` key in all three languages, and `npm run test:i18n` enforces parity — worth re-checking after merging your +248 lines.

## Before pushing

CI now runs these before Android packaging, so it's quickest to run them locally:

```bash
cd web && npm run build && npm run test:i18n && npm run test:config && npm run test:ui && npm run test:settings && npm run test:model && npm run test:events
cd ../bridge && npm test
```

Squashing is fine by us. If the rebase turns out nastier than this makes it sound, say so and we'll split it — the streamed rendering could land ahead of the question answering, which is where most of the backend-specific work sits.

giuliastro:
One clarification on how this will land, since it matters for credit: your commits will be **merged**, not re-implemented. Anything that still needs changing afterwards goes in separate commits on top, so the authorship stays yours.

For #30 and #31 I reworked your commits instead of merging them, which was the wrong call — same code, but the authorship became mine. I have since added `Co-authored-by:` trailers to both, so you are in the contributors graph for that work. Merging properly from the start is still the better route, and that is what this one gets.

So: no need to squash on our account. Keep whatever commit structure makes the rebase easiest for you.

birabittoh:
New PR: #37

Issue #36, opened 2026-07-25
Add PI as a supported harness


giuliastro:
[PI](https://pi.dev/) is the minimal agent harness that Oh My Pi derives from ([earendil-works/pi](https://github.com/earendil-works/pi), MIT, free). It is listed as planned in the README and this issue is the work. **Not assigned — comment if you want to take it.**

## Why this is mostly reuse

The OMP backend already talks to an agent over ACP through the bridge in `bridge/`. Almost nothing in there is OMP-specific: it speaks plain ACP JSON-RPC over stdio and translates it to the HTTP/SSE API the app consumes. The methods it uses are `session/new`, `session/load`, `session/list`, `session/prompt`, `session/cancel` and `session/set_config_option`.

## Two routes, with a real trade-off

**Route A — the community ACP adapter.** PI's maintainer [declined native ACP support](https://github.com/earendil-works/pi/issues/175), suggesting an adapter on top of PI's RPC mode instead. Several exist; [`@victor-software-house/pi-acp`](https://github.com/victor-software-house/pi-acp) looks the most maintained (308 tests, semantic-release, active) and PI is in the [ACP agent registry](https://zed.dev/acp/agent/pi). It launches over stdio with `npx -y @victor-software-house/pi-acp`, which is the same shape the bridge already spawns, and it supports every method listed above including model selection via `session/set_config_option`.

Caveats worth knowing before starting: its development is centred on Zed, so behaviour may differ from the spec elsewhere — we learned with OMP that real agents deviate in ways no test double predicts. It also does not emit `agent_plan`, so the todo/plan panel will stay empty (see below). MCP servers are not wired through, which does not affect us: the bridge always passes `mcpServers: []`.

**Route B — PI's RPC mode directly.** First-party, documented in `packages/coding-agent/docs/rpc.md` in the PI repo, no third-party dependency. Costs a new transport in the bridge instead of reusing the ACP client.

Route A is the cheaper first cut and the one I would start with. Route B is the better long-term answer if the adapter turns out flaky.

## What actually needs changing

Two hard-coded OMP assumptions will block Route A immediately:

1. **`bridge/src/acp-client.js:40`** spawns `[ompBin, "acp"]` — the `acp` subcommand is baked in. PI's adapter is a different executable that takes no such argument. This needs to become a configurable command plus arguments (`--agent-cmd` alongside the existing `--omp-bin`, or a per-backend descriptor).
[...]

PR #37, opened 2026-07-25
Streamed rendering, grouped tool/diff/reasoning formatting, interactive questions (rebased on v2.0.0)

birabittoh:
## Summary

Rebase of #32 onto current `main` (v2.0.0 / Harness Remote / Oh My Pi backend), per the review on that PR. Same feature set:

- Stream assistant replies token-by-token instead of waiting for the full message, and keep the message list pinned to the bottom while it grows without fighting a user who's scrolled away.
- Group each turn's thinking/tool-calls/edits into a single collapsible "action" row, with a modal for full detail.
- Render tool calls, diffs, reasoning, to-do updates, and question prompts with purpose-built formatting instead of raw JSON/command dumps, including colored unified diffs with an expandable full-diff modal.
- Let the user answer the assistant's interactive questions (`AskUserQuestion`) directly from the chat.
- Localize all newly added strings across English, Italian, and Traditional Chinese.
- Assorted layout/mobile fixes, plus a few small polish items (trimmed stop-button label, removed redundant session "Open" button, Refresh/New Session sizing).
- Performance: memoize the message list and each message so composer typing and unrelated state updates don't re-render the whole conversation; keep unchanged messages referentially stable across streamed updates and periodic refetch; skip the polling fallback while live updates are connected.

## What changed vs #32

- Dropped the commits from #30/#31, since those landed separately (reworked) in #35.
- Resolved the 10-hunk `App.tsx` conflict against `main`: kept the multi-backend config layer, `loadedMessagesRef` snapshot logic, and the autosave host-validation guard that prevents a half-typed host from bricking the app; kept this branch's `isNearMessagesBottom` scroll tracking, `MessagesPane` extraction, and session-card layout, merged with the OMP-aware gating and `Open` button.
- Gated the `questions` fetch in `loadSelected` behind `config.backend`, since the OMP bridge doesn't implement that endpoint.
- Updated `ui-regression.test.mjs`'s auto-scroll assertion to verify the `stickToBottomRef` gate instead of forbidding `messageScrollSignature` outright, preserving pinned-scroll-while-streaming behavior.

Squashing is fine if you'd rather merge it as one commit.

## Test plan

- [x] `npm run build`
- [x] `npm run test:i18n`
- [x] `npm run test:config`
- [x] `npm run test:ui`
- [x] `npm run test:settings`
- [x] `npm run test:model`
- [x] `npm run test:events`
- [x] `cd bridge && npm test`


giuliastro:
Merged, and released as [v2.1.0](https://github.com/giuliastro/harness-remote/releases/tag/v2.1.0). Thank you — this is the biggest single improvement the app has had, and it shows: the chat went from a wall of raw output to something you can actually follow on a phone. The grouped action rows and the inline question answering in particular are the kind of thing I would not have got to on my own.

The rebase was also exactly what I asked for and then some. You kept every point from the review — the multi-backend config layer, the `loadedMessagesRef` snapshot logic, the autosave host-validation guard — gated the `questions` fetch behind the backend so the OMP bridge is not asked for an endpoint it does not have, and dropped the #30/#31 commits. Your 34 commits went in as a merge, so the authorship is yours in the contributors graph.

The one thing I changed was on my side, not yours. Your stall watchdog is a good idea and its premise holds for OpenCode, which beats every 10s — but the bridge only wrote when a session event occurred, so an idle OMP session looked like a dead connection. Measured in a browser against a real OMP install, counting requests to `/global/event`:

| | requests in ~115s idle | started at |
|---|---|---|
| before | 4 | 0s, 0s, 31s, 62s |
| after | 1 | 0s |

Worth knowing because the churn also suppressed the polling fallback you skip while live updates report connected, so each reconnect window left a gap with neither running. Fixed in #38 by making the bridge beat every 10s like OpenCode. Nothing for you to do.

Also verified against real OMP 17.1.3, not just the fakes: session list, open with full history replay, prompt with streamed output, model selection.

If you fancy another one, PI support is open in #36 with the groundwork mapped out — the bridge should be largely reusable through a community ACP adapter.

PR #40, opened 2026-07-26
Fix issues from last PR

birabittoh:
Some fixes that I missed on my last PR:
* Way better grouping logic
* Keep nav pinned on desktop, only let it scroll away on mobile

Here's what the new grouping looks like:
immagine

### DIFF
--- a/web/src/App.tsx
+++ b/web/src/App.tsx
@@ -764,6 +764,7 @@ function buildMessageTimeline(parts: MessagePart[]): TimelineItem[] {
   }
   for (const part of parts) {
     if (part.type === "step-start" || part.type === "step-finish") continue
+    if (part.type === "text" && !part.text) continue
     if (ACTION_GROUP_TYPES.has(part.type)) {
       buffer.push(part)
     } else {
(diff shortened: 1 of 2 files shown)


giuliastro:
Merged — thanks. The continuous timeline across message boundaries is the right model: an action group that happened to be split by a message boundary is still one action, and treating it as one reads much better. The screenshot sold it.

One thing I changed on top, in #41. The fallback formed a run whenever the buffer held more than one message, including when nothing in it was groupable:

```js
if (buffer.length === 1 && buffer[0].parts.every((part) => !ACTION_GROUP_TYPES.has(part.type))) {
```

Since `ACTION_GROUP_TYPES` is `reasoning`/`tool`/`patch` and OMP only ever produces `text` parts, a run should never form on that backend — but two consecutive replies did, and merged into one bubble. It shows up in the queued-prompt flow that landed in #35: while a queued prompt runs the live order is user, user, assistant, assistant, so both answers shared a bubble with nothing to say which prompt each answered. Reopening the session split them again, since OMP's persisted history is strictly interleaved, so the same conversation rendered two ways.

A run now forms only when the buffered messages actually contain a groupable part. Your behaviour for the case you wrote it for is untouched.

Not your bug to have predicted — the bridge-side prompt queue is mine and its live ordering is a quirk of how a queued prompt is recorded before it runs. Worth knowing about if you touch that area again.

Also worth recording, since it briefly looked like a bug in your change: while testing this I saw two assistant bubbles with identical text. That turned out to be the model, not the code — a session with a dozen near-identical `Say only: X` turns in context and a small free model. On a fresh session the queue dispatches correctly.

PR #42, opened 2026-07-26
Feat/pi support

Baylar55:


Baylar55:
### Title                                                                    
                                                                              
 ```text                                                                      
   feat: add PI ACP backend support                                           
 ```                                                                          
                                                                              
 ### Description                                                              
                                                                              
 ```markdown                                                                  
   ## Summary                                                                 
                                                                              
   Adds PI as a supported Harness Remote backend through the community        
   `@victor-software-house/pi-acp` ACP adapter.                               
                                                                              
   ## Changes                                                                 
                                                                              
   - Generalizes ACP process launching to support a command plus argument     
 list.                                                                        
   - Adds bridge backend selection:                                           
     - `--backend omp`                                                        
     - `--backend pi`                                                         
   - Adds PI defaults using:                                                  
     - `npx -y @victor-software-house/pi-acp`                                 
     - port `4097`                                                            
     - username `pi`                                                          
   - Handles ACP authentication methods other than the OMP-specific `agent`   
 ID.                                                                          
   - Adds Windows-safe launching for `.cmd` ACP adapters.                     
   - Renames the shared OMP service to the backend-neutral `AcpService`.      
   - Adds backend-aware service routing and health responses.                 
   - Adds PI to the web backend selector and persistent connection settings.  
   - Gates unsupported PI features such as agent selection, rename/delete,    
 and diff/VCS.                                                                
   - Adds PI setup instructions and limitations to the README.                
   - Adds bridge and web regression coverage.                                 
                                                                              
   ## Validation                                                              
                                                                              
   Passed:                                                                    
                                                                              
   - `cd bridge && npm test` — 29 tests                                       
   - `cd web && npm run build`                                                
   - `npm run test:i18n`                                                      
   - `npm run test:config`                                                    
   - `npm run test:ui`                                                        
   - `npm run test:settings`                                                  
   - `npm run test:model`                                                     
   - `npm run test:events`                                                    
   - Browser smoke test confirmed the PI selector, port `4097`, and username  
 `pi`.                                                                        
                                                                              
   ## Limitation                                                              
                                                                              
   The real `pi-acp` smoke test could not complete on the development         
 workstation because the current adapter package launches through `bun`,      
 which was not installed. The README documents Bun as a PI prerequisite.      
                                                                              
   The implementation is based on the ACP adapter’s documented session,       
 streaming, authentication, cancellation, and model-selection behavior and    
 should be tested against a real PI installation during review.               
 ```                                                                          
                                                                              
 You can also add this short comment after opening the PR:                    
                                                                              
 ```markdown                                                                  
   The bridge and web regression suites pass. The remaining review item is a  
 real PI adapter smoke test on a machine with Bun and PI credentials          
 available.                                                                   
 ```

gervaso-assistant:
Thanks for putting this together. We developed an independent PI implementation in parallel and have opened it as draft #43 so the approaches can be compared or consolidated rather than silently competing.

The draft uses the same `@victor-software-house/pi-acp` adapter, but also includes:

- a real PI adapter prompt/streaming/status smoke test;
- handling for agent-initiated `session/request_permission` requests (`allow_once`) instead of returning unsupported;
- centralized OMP/PI profiles and per-backend capabilities;
- OMP persisted-history, restart, and desktop-to-app continuation fixes discovered while extracting the shared ACP driver.

The comparison and validation evidence are in #43. This is not a request to discard this PR, and no code was copied from it. I am happy to consolidate the missing pieces into whichever architecture the maintainer prefers, while preserving credit for the independent work here.

giuliastro:
Merged — thank you, and welcome. This is the first outside answer to #36 and it arrived within hours of the issue going up, which is more than I expected.

You hit both blockers the issue named, exactly: the hardcoded `["acp"]` argument and the auth check that threw when no method with id `agent` was offered. Falling back to any advertised method and skipping authentication entirely when none is offered is the right shape.

The part I did not anticipate, and which is the most valuable thing in this PR, is the Windows handling. `npx` is a `.cmd` shim there and cannot be spawned directly. I verified it on my machine:

```
spawn("npx", ["--version"])       -> ERROR ENOENT
spawn("cmd.exe", [/d /s /c npx])  -> exited 0
```

Without your `ComSpec` wrapping the PI backend simply would not start on Windows, and the default `npx.cmd` on `win32` with `npx` elsewhere is the correct call. There is a parallel PI implementation in #43 that lacks this, so your branch is what makes PI actually launchable on Windows.

Three things were lost when you merged `main` into the branch — `main` moved a fair bit while this was open, so the conflicts were unavoidable rather than careless. Repaired in #44:

- `type Translator` was dropped, so `npm run build` failed with a dozen `TS2304` errors. Worth knowing: the web regression suites passed the whole time, because they assert against source text and never compile. `npm run build` is what catches it, and CI runs it.
- The SSE heartbeat was dropped, which reintroduces a 30s reconnect loop against the bridge. The heartbeat test caught that one by timing out.
- The session card `Open` button came back with a `PlayIcon` reference whose import no longer exists. It was removed on purpose as redundant with the card itself being clickable, so I removed the button rather than restoring the import.

Nothing for you to do — your commits are in as yours.

One thing still outstanding, and I would rather be honest about it than quietly announce PI as done: I could not smoke-test the adapter here, because PI is not installed on this machine and I did not want to pull a third-party package to find out. If you have run it against a real PI, saying so on #36 would let me announce it with confidence. Tool calls and permission prompts are the parts I would most want confirmed.

PR #46, opened 2026-07-26
refactor: centralize harness profiles and capabilities

gervaso-assistant:
## Summary

First reviewable slice extracted from #43, rebased onto `main` after #42 and #44.

- adds one bridge harness-profile registry for OMP and PI defaults/capabilities
- serves backend-specific capabilities instead of one shared static table
- makes web behavior capability-driven for models, agents, todos, diff, questions, rename, and delete
- scopes model selection by backend/session and ignores stale model-catalog requests
- introduces `HARNESS_REMOTE_*` environment names while retaining every `OMP_BRIDGE_*` name as a one-release compatibility alias

## Windows compatibility

This deliberately keeps #42's Windows launch path unchanged: PI still resolves to `npx.cmd`, and `AcpClient` still wraps `.cmd`/`.bat` through `ComSpec`. The profile layer sits on top of the merged `acp-service.js`, `--backend`, `--acp-command`, and `--acp-arg` implementation rather than replacing it.

## Scope

No permission-policy or OMP history/snapshot changes are included. Those are separate follow-ups, as requested in #43.

## Validation

- `cd bridge && npm test` — 32/32
- `cd web && npm run build`
- `npm run test:i18n`
- `npm run test:ui`
- `npm run test:config`
- `npm run test:settings`
- `npm run test:model`
- `npm run test:events`

Relates to #36 and follows the maintainer plan in #43.


giuliastro:
Rebased onto current `main` and merged as part of #50 — your commit is in with your authorship. Closing this since the content has landed.

The conflicts were mine, not yours: #45 and #49 landed on `bridge/src/acp-client.js`, `config.js` and their tests after you opened this, so I did the rebase rather than bouncing it back.

One note on #47 specifically: I had already implemented permission granting in #49 while unblocking PI on Node, before this extraction existed. Yours is the better version and I resolved toward it — `allow_once` rather than `allow_always`, so a grant covers the current call instead of persisting in the harness, and opt-in per harness via `permissionMode` rather than unconditional. I kept only the `permission` event from mine, so a grant stays visible in the log.

Verified after the rebase against real PI 0.82.1: `granted tool permission (allow_once)` and the tool call writes its file. OMP re-checked for regressions.

PR #47, opened 2026-07-26
feat: allow local ACP tool permissions once

gervaso-assistant:
## Summary

Second focused slice requested in #43, stacked on #46.

- handles `session/request_permission` as an ACP request instead of returning `-32601`
- prefers `allow_once`, then `allow_always`, and otherwise cancels cleanly
- keeps the default policy deny-safe inside `AcpClient`; local OMP/PI profiles opt into allow
- keeps unknown agent-initiated methods answered as unsupported so they cannot stall the agent

## Validation

- `cd bridge && npm test` — 32/32
- unit regression verifies `allow_once` wins even when `allow_always` is offered first
- real PI adapter `0.17.1`: session created, prompt completed, exact concatenated response `PI permission works`, final status `idle`

## Dependency

Stacked on #46 using the temporary `review/profiles-capabilities-base` base branch so this diff contains only the four permission-policy files. Retarget to `main` after #46 merges.

Relates to #36.

### DIFF
--- a/bridge/src/acp-client.js
+++ b/bridge/src/acp-client.js
@@ -8,6 +8,7 @@ export class AcpClient extends EventEmitter {
   #command
   #args
   #spawn
+  #permissionMode
   #child
   #buffer = ""
   #nextID = 1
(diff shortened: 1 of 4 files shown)


giuliastro:
Rebased onto current `main` and merged as part of #50 — your commit is in with your authorship. Closing this since the content has landed.

The conflicts were mine, not yours: #45 and #49 landed on `bridge/src/acp-client.js`, `config.js` and their tests after you opened this, so I did the rebase rather than bouncing it back.

One note on #47 specifically: I had already implemented permission granting in #49 while unblocking PI on Node, before this extraction existed. Yours is the better version and I resolved toward it — `allow_once` rather than `allow_always`, so a grant covers the current call instead of persisting in the harness, and opt-in per harness via `permissionMode` rather than unconditional. I kept only the `permission` event from mine, so a grant stays visible in the log.

Verified after the rebase against real PI 0.82.1: `granted tool permission (allow_once)` and the tool call writes its file. OMP re-checked for regressions.

PR #48, opened 2026-07-26
fix: reconcile OMP sessions across clients

gervaso-assistant:
## Summary

Third focused slice requested in #43, stacked after #46 and #47.

- reads persisted OMP user/assistant history without loading a second ACP process just to display it
- persists the last successful bridge message/todo snapshot across navigation and restarts
- reconciles empty/partial ACP replay without erasing newer cached messages or completed todo state
- keeps desktop-created and pre-restart sessions writable; first app prompt loads and continues them
- merges newly persisted desktop/app branches during polling
- marks external sessions for an explanatory UI hint, without disabling the composer
- adds relocatable per-backend state via `--state-dir` / `HARNESS_REMOTE_STATE_DIR`

## Windows teardown correction

The failing ownership regression now calls `await restarted.flushSnapshots()` before removing its temporary directory. No asynchronous snapshot writer remains when `rm(..., { recursive: true })` runs, addressing the reported Windows `ENOTEMPTY` race at its source.

## Validation

- `cd bridge && npm test` — 39/39
- `cd web && npm run build`
- all six web regression commands pass
- OMP bridge smoke: health returned backend `omp`, version `17.0.8`; OMP capability profile returned models/todos enabled
- a fresh live-agent turn was attempted, but the installed OMP account currently returns `usage_limit_reached`; no real-agent success is claimed for this rebased branch

## Dependency

Stacked on #47 using `review/acp-permission-base` so this PR shows only OMP reliability. Retarget to `main` after the preceding slices merge.

Relates to #36.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,7 +94,7 @@ If remote/mobile cannot connect, open TCP 4096 in your OS firewall and network f
 
 ### Oh My Pi Bridge Setup
 
-Harness Remote connects to OMP through the bridge included in this repository. The bridge starts `omp acp` on the same computer, translates its ACP stdio protocol to the app's HTTP/SSE API, and never reads or modifies OMP's internal databases.
+Harness Remote connects to OMP through the bridge included in this repository. The bridge starts `omp acp` on the same computer and translates its ACP stdio protocol to the app's HTTP/SSE API. To show sessions created by another OMP process without loading and interrupting them, it reads the append-only user/assistant transcript under OMP's session directory; it does not modify OMP state.
 
 #### Prerequisites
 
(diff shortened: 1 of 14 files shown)


giuliastro:
Rebased onto current `main` and merged as #51 — your commit is in with your authorship. Closing this since the content has landed.

The conflicts were mine again: #49 landed on `acp-service.js` after you opened this. Your `#chunkMessageIDs` replaces the `#streaming` map I had added there — same problem, better solution, keyed by `sessionId:role` and handling replay explicitly on the reasoning that a replay update is a whole persisted block rather than part of a live turn. I removed mine; keeping both left a duplicate `messageID` declaration.

Verified against real OMP 17.1.3 on a session created by an earlier bridge process: history read without spawning a second agent, `external` flag set, title derived after load, and a new prompt answered with the history preserved. Also worth noting the `ENOTEMPTY` teardown failure I reported on #43 passes here — 42/42.

That completes the split you did of #43: #46, #47 and #48 are all in. Splitting it was the right call — reviewing three focused PRs against a moving `main` was tractable in a way the 27-file version would not have been, and it let me take your permission policy over mine on its merits rather than as a package deal.

PR #67, opened 2026-07-27
fix: prevent long session titles from breaking layout

Baylar55:
## Why
Returning from a session to the sessions tab could leave a very long session title wider than its card, causing the mobile layout to overflow and break the UI.

## What changed
- Constrain session card content so it can shrink in narrow layouts.
- Render long session titles as a single-line ellipsis while preserving the full title in the existing title attribute.
- Add a regression guard for the CSS invariant.

## Verification
- Reproduced the overflow at 390px with a 240-character title before the fix.
- Confirmed ellipsis containment at 390px and 900px in the running Vite app.
- npm run build
- npm run test:i18n
- npm run test:config
- npm run test:ui
- npm run test:settings
- npm run test:model
- npm run test:events
- bridge: npm test (43 passing)

Not tested against a real harness; this is a UI-only change.

### DIFF
--- a/web/src/styles.css
+++ b/web/src/styles.css
@@ -526,8 +526,14 @@ p {
   grid-template-columns: minmax(0, 1fr);
   align-items: start;
   gap: var(--space-3);
+  min-width: 0;
 }
 
+.session-card h3 {
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
+}
 .session-card-main p,
 .session-card > p {
   margin-top: var(--space-1);
--- a/web/src/ui-regression.test.mjs
+++ b/web/src/ui-regression.test.mjs
@@ -10,6 +10,8 @@ const refreshButton = app.match(/<button onClick=\{refreshSessionsWithIndicator\
 assert.ok(refreshButton, 'sessions refresh button should call refreshSessionsWithIndicator')
 assert.ok(refreshButton[0].includes('RefreshIcon'), 'idle sessions refresh button should render a non-spinning RefreshIcon')
 assert.ok(refreshButton[0].includes('refreshingSessions ? <LoadingIcon'), 'refresh button should spin only during an active manual refresh')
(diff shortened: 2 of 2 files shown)


giuliastro:
Thanks for this — merged, and the diagnosis was right. `min-width: 0` on `.session-card-main` is the actual fix: a grid child cannot shrink below its content without it, which is exactly what let a long title widen the card and push the mobile layout sideways. Good catch, and thank you for the repro details and for running the full suite before opening it.

I've adjusted one thing in #68, and I want to explain rather than just change it under you.

The `white-space: nowrap` next to it goes a bit further than the overflow needed: it collapses every multi-line card title to a single ellipsised line. The mobile cards have always wrapped — the README screenshots show two-line titles — so after the merge the app no longer matched its own documentation. `overflow-wrap: break-word` contains the title just as well, keeps the wrapping, and also handles the one case plain wrapping can't: a title that is a single unbroken token.

I measured both against the real stylesheet at a 390px viewport, with a long spaced title and a 128-character unbroken one:

| | `nowrap` | `overflow-wrap: break-word` |
|---|---|---|
| long spaced title | 1 line, truncated | 4 lines, wrapped |
| unbroken token | 1 line, truncated | 4 lines, wrapped |
| overflows card / viewport | no | no |

Both contain the overflow; only the second keeps the information. The desktop sidebar still shows one ellipsised line, which is deliberate there — it sets its own `nowrap` and keeps the full title in the tooltip.

One note on the regression guard, which is a repo convention rather than anything wrong with your patch: CONTRIBUTING asks these tests to assert the invariant rather than the spelling of the code, since a test pinned to specific declarations blocks a legitimate change later. So it now asserts "a long title must break rather than widen its card" instead of requiring `nowrap` and `ellipsis` specifically. Worth knowing for next time — and your `min-width: 0` assertion was already exactly the right shape.

Incidentally, writing that guard bit me in a way you might hit too: a bare `.session-card h3` regex also matches the tail of `.sidebar-sessions .session-card h3`, so my first version read the sidebar's intentional `nowrap` as the mobile regression and failed. The selectors are anchored now.

Thanks again — this was a real bug on a small screen, and it's the kind of report that's easy to miss from a desktop.

PR #77, opened 2026-07-28
feat: add Claude Code backend support via ACP bridge

gervaso-assistant:
## Summary

Adds **Claude Code** as a supported backend in Harness Remote, following the same architecture used for OMP and PI — an ACP bridge profile that launches the official `@agentclientprotocol/claude-agent-acp` adapter.

### Bridge — `bridge/src/harness-profiles.js`
- New `claude` profile: launches `npx -y @agentclientprotocol/claude-agent-acp@0.59.0` (pinned for stability)
- Uses `permissionMode: "allow"` (auto-grants tool permissions, same as PI/OMP)

### Web app — types, capabilities, storage
- `web/src/types.ts`: Extended `BackendKind` with `"claude"`
- `web/src/backendCapabilities.ts`: Capability map for Claude Code
- `web/src/backendClient.ts`: Backend metadata
- `web/src/storageKeys.ts`: Isolated storage key

### Web app — UI (`web/src/App.tsx`)
- Backend selection, display name, default config, help section, README link

### Documentation — `README.md`
- Full Claude Code Bridge Setup section with prerequisites and limitations

## Architecture

```
App --HTTP/SSE--> bridge --ACP stdio--> @agentclientprotocol/claude-agent-acp --SDK--> Anthropic API
```

## Caveats

- Auto-grants tool permissions (same as PI pattern)
- Requires `claude login` or `ANTHROPIC_API_KEY` on the host
- Not yet tested end-to-end with a live Claude Code installation

## Verification

- TypeScript `tsc --noEmit`: passes
- Vite production build: 277 modules, succeeds
- Bridge tests: 0 failures

### DIFF
--- a/CLAUDE-CLI-INTEGRATION-PLAN.md
+++ b/CLAUDE-CLI-INTEGRATION-PLAN.md
@@ -1,0 +1,156 @@
+# Piano di integrazione Claude Code (CLI)
+
+## Identità prodotto
+
+**Harness Remote** è agnostico rispetto all'harness. OpenCode, OMP e PI sono backend selezionabili.
+Aggiungere Claude Code come backend significa aggiungere una voce nell'enum `BackendKind`, un profilo
+nel bridge e la sezione corrispondente nel README / help in-app.
+
+## Decisione architetturale
+
+Claude Code **non espone un server HTTP/SSE nativo** (≠ OpenCode) e **non supporta ACP nativamente**
+(issue #6686 chiusa come `not_planned`). Ha però:
+
+1. Un **Agent SDK ufficiale** (`@anthropic-ai/claude-agent-sdk`) TypeScript/Python
+2. Un **adattatore ACP ufficiale** (`@agentclientprotocol/claude-agent-acp`) mantenuto dal team ACP
+
+L'approccio più rapido e a basso rischio è **riutilizzare l'architettura bridge esistente** (stessa
+usata per OMP e PI), aggiungendo un profilo harness che lancia l'adattatore ACP ufficiale:
+
+```mermaid
+flowchart LR
+  A[App Android / Web] -->|HTTP + SSE| B[bridge nel repository]
+  B -->|ACP JSON-RPC su stdio| C[claude-agent-acp]
+  C -->|Claude Agent SDK| D[Claude Code / Anthropic API]
+```
+
+L'adattatore `@agentclientprotocol/claude-agent-acp` traduce ACP in chiamate al Claude Agent SDK,
+che a sua volta parla con l'API Anthropic. Il bridge non cambia affatto il suo layer ACP:
+`AcpClient` e `AcpService` funzionano già con qualsiasi processo ACP-compatible.
+
+## Modifiche necessarie
+
+### 1. Bridge — `bridge/src/harness-profiles.js`
+
+Aggiungere profilo `claude`:
+
+```js
+claude: {
+  id: "claude",
+  label: "Claude Code",
+  command: "npx",
+  // Pinned to avoid the `notarget` scenario that PI hit.
+  args: ["-y", "@agentclientprotocol/claude-agent-acp@0.59.0"],
+  permissionMode: "allow",
+  capabilities: {
+    ...COMMON_CAPABILITIES,
+    models: false,
+    todos: true,
+    commands: false
+  }
+}
+```
+
+L'adattatore ACP ufficiale supporta: streaming, tool calls, TODO, @-menzioni, immagini,
+terminali interattivi. `permissionMode: "allow"` auto-grant dei permessi (stesso pattern di OMP/PI).
+
+### 2. Web — `web/src/types.ts`
+
+Aggiungere `"claude"` al tipo `BackendKind`:
+
+```ts
+export type BackendKind = "opencode" | "omp" | "pi" | "claude"
+```
+
+### 3. Web — `web/src/backendCapabilities.ts`
+
+Aggiungere mappa per `claude`:
+
+```ts
+claude: {
+  sessions: true,
+  prompt: true,
+  abort: true,
+  streaming: true,
+  models: false,
+  agents: false,
+  todos: true,
+  diff: false,
+  filesystemBrowser: true,
+  questions: false,
+  commands: false,
+  sessionRename: true,
+  sessionDelete: true
+}
+```
+
+### 4. Web — `web/src/backendClient.ts`
+
+Aggiungere metadati:
+
+```ts
+claude: {
+  modelSelectionRequiresSession: true,
+  messageRefreshSupported: true
+}
+```
+
+### 5. Web — `web/src/storageKeys.ts`
+
+Aggiungere chiave:
+
+```ts
+claude: "opencode.remote.server.claude"
+```
+
+### 6. Web — `web/src/App.tsx`
+
+Punti da modificare:
+
+| Linee | Cosa |
+|-------|------|
+| `isBridgeBackend` | Aggiungere `backend === "claude"` |
+| `backendDisplayName` | Aggiungere `"Claude Code"` |
+| `defaultConfig` | Port 4097, username `claude` |
+| `parseStoredConfig` | Validare `"claude"` come backend valido |
+| `initialConfig` | Validare `"claude"` come backend valido |
+| `backend <select>` | Aggiungere `<option value="claude">Claude Code (ACP bridge)</option>` |
+| `placeholder` username | Cambiare condizione |
+| Help server section | Aggiungere blocco istruzioni per Claude |
+| README link | Aggiungere anchor per claude |
+
+### 7. Bridge — test `bridge/test/harness-profiles.test.js`
+
+Aggiungere test per il nuovo profilo.
+
+## Cose che funzionano già senza modifiche
+
+- `acp-client.js` — già ACP-compatible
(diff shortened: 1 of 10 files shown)


giuliastro:
Reviewed and tested this end to end, which the description lists as missing. It works — and there are four things I would want resolved before it lands.

## It works, and a subscription login is enough

Ran the bridge with `--backend claude` against a real Claude Code installation:

```
/v1/health  →  {"healthy":true,"backend":"claude","version":"0.59.0"}
session     →  created
prompt      →  [user] Reply with exactly: BRIDGE_OK
               [assistant] BRIDGE_OK
```

`ANTHROPIC_API_KEY` was **not** set. The Agent SDK accepted the stored `claude auth login` session, so a Claude subscription is sufficient and no metered API billing is involved. That resolves the open question in the caveats: `claude login` alone is enough, and the README can say so rather than offering the API key as an equal alternative.

Checks all pass on the branch: `npm run build`, the six web suites, and 45/45 bridge tests.

## 1. `--root` restricts neither the session list nor transcript reads

This is the one I would fix before merging. Started the bridge with `--root` pointing at an empty scratch directory. The session list came back with every Claude Code session on the machine, from unrelated projects. Picking one whose `cwd` was a different repo entirely:

```
target cwd:        <another repo, outside --root>
root passed:       …/scratchpad/claude-root
GET /session/<id>/message  →  96 messages returned
```

Strictly this is pre-existing bridge behaviour — `listSessions` only filters by directory when the caller passes one, and the README already says `--root` governs browsing and new-session cwd rather than being a sandbox. But the consequence changes character with this backend. OMP and PI histories are per-project and short-lived; Claude Code keeps a long global history spanning every repository the user has ever worked in. So on this backend, anyone holding the bridge credentials can enumerate and read every past conversation on the machine, and `--root` reads like it would prevent that.

Two ways out, either is fine: filter the listing by the configured roots, or state plainly in the Claude section that `--root` does not limit which sessions are visible and that the bridge therefore exposes the machine's whole Claude Code history.

## 2. Delete hides rather than deletes, and rename does not reach the harness

Both are bridge-local. After `DELETE /session/<id>` the session is gone from the listing, but the harness still holds it on disk (`~/.claude/projects/…/<id>.jsonl`, 17 KB in my run). Restarting the bridge with a fresh `--state-dir`:

```
REAPPEARED — title now: 'Verify bridge connection'
```

It comes back, and the rename is gone too — the title reverted to the one the harness generated, not the one set through the bridge and not the one the session was created with. So both operations live entirely in `~/.harness-remote/<backend>/` and vanish with it.

That is a defensible design — a client-side archive and nickname — but it is not what the app's **Delete** button implies, and someone who deletes a session from their phone and later opens Claude Code will find it still there. Worth either naming it for what it is in the UI and the README, or implementing it against the harness where the protocol allows.

## 3. The capability flip covers OMP and PI, not just Claude

`COMMON_CAPABILITIES` moves `sessionRename` and `sessionDelete` from `false` to `true`, so this PR enables rename and delete for OMP and PI as well, and `backendCapabilities.ts` mirrors it. The README currently documents both as intentionally unavailable on those backends.

If that is the spillover from the separate rename/delete work, it belongs there rather than here, so this PR stays reviewable as "adds a backend" and the capability change gets discussed on its own terms. If the two are meant to land together, say so and the README sections for OMP and PI need updating in the same change.

## 4. Smaller things

- The adapter is pinned to `0.59.0`; current is `0.63.0`. Pinning is right and matches the reasoning already applied to `pi-acp` — just noting the gap is four minors.
- `@agentclientprotocol/claude-agent-acp` declares `engines: node >=22`, while CONTRIBUTING.md still says Node 20 or newer for the bridge. Same situation as the PI adapter; the prerequisite line should be precise now that two backends need 22.
- `CLAUDE-CLI-INTEGRATION-PLAN.md` adds 156 lines. There is a precedent in `OMP-INTEGRATION-PLAN.md`, so this is a choice rather than a problem — worth making deliberately.

One thing worth calling out as correct: the CORS change to allow `PATCH` and `DELETE` in the preflight. The same change was proposed in #59 and I closed it there because the bridge implemented neither method, so the preflight would have promised something that 404s. Here the handlers exist, so the header is now accurate.

## Not covered

I did not exercise `todos: true` or confirm that `models: false` matches what the adapter actually offers.

gervaso-assistant:
Thanks for the thorough E2E test and review. I agree with all four points. Here is how I plan to address them:

### Changes made (already pushed)

**1. Session visibility and `--root`**
Documented explicitly in the Claude Code section: `--root` does not filter the session list. The bridge enumerates all Claude Code sessions on the machine, potentially spanning every repository the user has ever worked in.

**2. Rename/delete are bridge-local**
Added the same disclaimer already present in the OMP section: renames persist in `~/.harness-remote/claude/` only; deletion hides the session from this bridge but does not erase Claude Code's own history. Deleted sessions reappear if the bridge starts from a fresh state directory.

**3. sessionRename/sessionDelete no longer spill to OMP and PI**
Reverted `COMMON_CAPABILITIES` and `backendCapabilities.ts` OMP/PI entries back to `sessionRename: false, sessionDelete: false`. Only the `claude` profile has them `true`. The OMP and PI README sections already say rename/delete are unavailable, which is now accurate again.

**4. Version pin and Node requirement**
- Pinned to `@agentclientprotocol/claude-agent-acp@0.63.0`
- README now states Node.js 22 or newer for the Claude Code adapter (matching the PI adapter's requirement)

### What I have not changed

- The CORS preflight headers (`PATCH`, `DELETE`) were already present as pre-existing changes in my working tree before this PR. They are not part of my intended diff but I can revert them if you prefer them discussed separately.
- The `preserveListedTimestamps` and `reloadOnHistoryRefresh` profile fields — added to `pi` and `claude` profiles because they were already in the working tree. Again, revertible on request.

giuliastro:
Thanks — `b787e64` closes two of the four: the capability flip is back off for OMP and PI, so this reads as "adds a backend" again, and the adapter now tracks `0.63.0`.

Two remain, and the first is the one I would not merge without:

- **`--root` restricts neither the session list nor transcript reads.** With this backend that means the bridge exposes every Claude Code conversation on the machine, across all projects, to anyone holding its credentials. Either filter the listing by the configured roots, or say plainly in the Claude section that `--root` does not limit which sessions are visible.
- **Delete hides rather than deletes, and rename never reaches the harness.** Both survive a bridge restart but not a change of `--state-dir`, and Claude Code keeps the session either way.

That second point now also applies to #78, which carries a byte-identical `renameSession` / `deleteSession` and enables them for OMP and PI. The two changes therefore collide on `acp-service.js`, `server.js` and the CORS header.

I would let #78 go in first — it owns rename/delete for the existing backends — and rebase this on top afterwards. What is left here is then just the `claude` profile, its capability entry and the README section, which is a much easier thing to review and to keep.

Also still open from the earlier pass: CONTRIBUTING.md says Node 20 or newer for the bridge while this adapter needs 22, and `CLAUDE-CLI-INTEGRATION-PLAN.md` is a deliberate choice worth confirming rather than a problem.

gervaso-assistant:
Agreed on all three points. Here is what I have done and what I propose:

### Done (pushed in 071ed33)

**Session visibility and `--root`** — Now stated plainly in the Claude Code section: "Session visibility is not restricted by `--root`. The bridge enumerates all Claude Code sessions on the machine, potentially spanning every repository the user has ever worked in. Anyone holding the bridge credentials can list and read every past conversation."

**Rename/delete scope** — Replaced the misleading "persistent local rename/delete" with a clear disclaimer matching the OMP section: renames are bridge-local, deletion hides from this bridge only, sessions reappear after a fresh `--state-dir`.

**Node 22** — README prerequisites now say Node.js 22 or newer. Also updated CONTRIBUTING.md to say Node 22.

### Proposed: let #78 land first

I agree that #78 should go in first — it owns rename/delete for OMP/PI and touches `acp-service.js`, `server.js`, and CORS. Once it lands, I will rebase this PR on top of it. What remains here will then be just the `claude` profile, its capability/storage/UI entries, and the README section, which is much easier to review.

### Plan file

`CLAUDE-CLI-INTEGRATION-PLAN.md` follows the existing `OMP-INTEGRATION-PLAN.md` precedent. I am happy to keep it.

giuliastro:
Rebased onto the merged #78 and fixed two things the rebase exposed, so this is down to 10 files and mergeable.

**The rebase quietly undid #78.** Replaying this branch's own "revert sessionRename/Delete for OMP/PI" on top of the change that had just enabled them left the app's capability defaults at `false` for OMP and PI while the bridge profile still advertised `true`. The handshake would have papered over it at runtime, which is why it is worth naming: the two layers disagreeing is invisible in testing. Both are `true` again.

**The harness table stopped being a table.** Adding the Claude Code row gave the separator a fourth column in a three-column table. Fixed, along with "all three harnesses" (four now) and the rename/delete entry, which names Claude Code alongside OMP and PI since it is bridge-local there too.

**Verified on the rebased code, not just on the branch as it was:**

```
/v1/health      →  {"healthy":true,"backend":"claude","version":"0.63.0"}
session         →  created, title 'Remote session'
prompt          →  [user] Reply with exactly: REBASE_OK
                   [assistant] REBASE_OK
```

Plus a clean build, 6/6 web suites and 48/48 bridge tests. The `BackendKind` guard that landed with #78 passes here, which is the machine-checked form of "Claude Code can actually be selected" — the thing that was missing when these two changes were briefly entangled.

PR #78, opened 2026-07-28
Fix remote OMP and PI session flows

gervaso-assistant:
## Summary
- add bridge-local rename and delete controls for OMP and PI sessions
- persist the bridge nickname/archive metadata in its state directory without altering harness history
- stabilize PI session creation, model loading, and ordering when ACP omits update timestamps
- use localized Remote session titles for newly created remote sessions
- allow credentialed browser CORS preflights for PATCH and DELETE session actions

## Session metadata scope
ACP provides no physical session-delete operation. Rename is a bridge-local nickname and delete hides a session from that bridge only; desktop OMP/PI history remains untouched. Both records are kept in `--state-dir`, so clearing or relocating it restores native titles and hidden sessions to the bridge listing.

## Verification
- `npm test` in `bridge` — 48 passing
- `npm run test:i18n`, `npm run test:ui`, `npm run test:model`, and `npm run build` in `web` — passing
- browser smoke tests: OMP model selector and PI new-session composer

### DIFF
--- a/OMP-INTEGRATION-PLAN.md
+++ b/OMP-INTEGRATION-PLAN.md
@@ -134,7 +134,12 @@ Completato:
 - bridge `v0.1.7`: snapshot persistiti separati dallo streaming live; il polling non forza più `session/load`;
 - smoke reale contro OMP `v17.0.8` completato su due sessioni e sei cicli di polling.
 
-Ancora intenzionalmente non supportato: rinomina/eliminazione persistente di una sessione OMP, comandi server OpenCode, agenti OMP configurabili, diff/VCS e accesso filesystem fuori dalle root consentite.
+2: Ancora intenzionalmente non supportato: comandi server OpenCode, agenti OMP configurabili, diff/VCS e accesso filesystem fuori dalle root consentite. Rinomina ed eliminazione sono metadati locali del bridge: il rename è un nickname e il delete nasconde la sessione solo da quel bridge. Sono salvati nello state directory; se viene svuotato o sostituito, titoli nativi e sessioni nascoste riappaiono. ACP non offre un'operazione di eliminazione fisica, quindi la cronologia nativa dell'harness resta intatta.
+3: Rename and delete use the same controls as OpenCode, but are bridge-local metadata: a rename is a nickname and delete hides the session only from this bridge. Both are stored under its state directory, so clearing or moving `--state-dir` restores the harness title and makes hidden sessions visible again. ACP does not define physical session deletion; the native OMP history remains untouched and visible to desktop clients.
+4: queued follow-up prompts, model selection, and bridge-local rename/delete.
+The nickname and hidden-session records live under the bridge state directory:
+clearing or moving it restores PI's native title and listing. ACP does not define
+physical session deletion, so deleted sessions remain in PI's own history.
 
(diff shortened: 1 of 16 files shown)


giuliastro:
Tested this against a real OMP install (17.1.5) through the bridge, with a throwaway session in an isolated `--root`, plus the full check set. Everything you list passes on my side too: **48/48** bridge tests, `npm run build`, and all six web suites. Creating a session with no title gives the new default, and both new operations work:

```
POST /session                    →  title 'Remote session'
PATCH /session/<id> {title}      →  'Renamed on PR 78'
DELETE /session/<id>             →  true, gone from the listing
```

Three things before it lands.

## 1. "Persistent" is narrower than it sounds

The summary says *persistent rename and delete*. Measured, it means "survives a bridge restart as long as the state directory survives". Restarting with a fresh `--state-dir`:

```
REAPPEARED — title now: 'Session 019fa92a'
```

The deleted session comes back and the rename is gone — the title falls back to the generated `Session <id>` form. Both operations live entirely in `#titles` / `#deletedSessions` and their snapshot under `~/.harness-remote/<backend>/`; the harness is never told. So deleting a session from a phone leaves it untouched in desktop OMP, and renaming it there shows the old title.

That is a reasonable design — a client-side archive and nickname — and given ACP exposes no session deletion it may be the only one available. The issue is only that the word promises more than the code does. Either scope the claim ("hidden from this bridge, retained in the harness") or, if ACP does offer a way through, use it.

It also means a user who relocates or clears `--state-dir` silently gets their deleted sessions back. Worth a line in the README next to the existing `--state-dir` note.

## 2. The README still says these are unavailable

The change flips `sessionRename` and `sessionDelete` to `true` for OMP and PI, but no documentation moves with it. The OMP section still reads *"Agent selection, persistent session rename/delete, server slash commands, and VCS/diff are intentionally unavailable"*, and the PI section says the same. CONTRIBUTING.md asks for the harness section to state what a backend cannot do — the inverse holds when it gains something, and here the wording is now actively wrong in both sections.

## 3. This collides with #77, and the order matters

#77 adds the same `renameSession`, `deleteSession`, `#deletedSessions` and `#requireSession` to `acp-service.js`, the same `PATCH` / `DELETE` handlers to `server.js`, and the same `Access-Control-Allow-Methods` line. The implementations are identical; #77 simply enables them for `claude` only while this one enables them for OMP and PI. Whichever merges first, the other conflicts.

Suggestion: land this one first, since it owns rename/delete for the existing backends along with the PI stability work, then rebase #77 on top. #77 then reduces to what its title claims — a harness profile, a capability entry and a README section — and the duplicated infrastructure disappears from it.

## Not covered

I did not exercise the PI path (`preserveListedTimestamps`, `reloadOnHistoryRefresh`) against a live PI adapter, so the ordering and timestamp fixes rest on your own smoke tests and the new unit coverage.

gervaso-assistant:
Thanks — addressed the documentation and scope issues in `2d04597`.

- The README and OMP integration plan now describe rename/delete as **bridge-local metadata**, explain that ACP has no physical delete, and state the `--state-dir` reset/relocation behavior explicitly.
- The PR description now uses the same scoped terminology rather than claiming unrestricted persistence.
- I agree on merge order: this PR should land before #77; #77 can then rebase on `main` and keep only Claude-specific profile/capability/documentation work.

I also reran the complete local suite after restoring the OMP/PI capability declarations on this branch: 48/48 bridge tests, all web regression suites, and the production build pass. My live PI smoke coverage remains limited to the configured adapter flow (create session, prompt, model selection, refresh/order stability), not an authenticated CI integration.

PR #97, opened 2026-07-30
fix(web): editable port field and single-read error bodies, plus PNA preflight support

giuliastro:
## What changed

- Kept the existing automatic configuration save, while preventing the settings modal from closing through backdrop clicks.
- Made the port field clearable and paste-friendly, with numeric keyboard input but no increment/decrement steppers.
- Fixed error handling for non-JSON bridge responses so the app does not attempt to consume a response body twice.
- Added Private Network Access CORS support for the public Pages app connecting to local bridges.

## Root cause

The desktop settings modal could be dismissed by a click that reached its backdrop, which made it appear to close while configuration state was updating. The modal now closes only via its explicit `×` button. The native numeric port control also forced an intermediate `0`, preventing normal replacement or paste. When a bridge returned a non-JSON error, the browser client attempted `json()` and then `text()` on the one-shot stream, masking the original server error.

## Validation

- `npm run test:settings`
- `npm run test:i18n`
- `npm run test:profiles`
- `npm run test:config`
- `npm run test:ui`
- `npm run build`
- `bridge: npm test`


giuliastro:
Merging with the modal backdrop change reverted. The other three stand.

The backdrop edit landed on `Modal` — the shared detail view for thoughts, tool calls and diffs — instead of `DesktopModalOverlay`, which is what wraps settings and help. Checked in the running app on this branch: the settings modal still closed on a backdrop click. Meanwhile the one modal where the dismissal costs nothing lost it, and `Modal` has no Escape handler to fall back on, so its only exit became the Close button.

The premise does not hold either. The debounced save is an effect in `App`, not in the modal, and closing settings does not unmount `App`. A keystroke followed 60ms later by a backdrop click still persisted (port 4099 written to the active profile). No edit was being lost.

The accompanying check read as if it pinned the settings modal, but it matched any backdrop without a click handler: swapping the two — settings non-dismissable, `Modal` dismissable — left it passing.

Kept:

- **`api.ts` one-shot body.** Real bug, right fix: the old `json()` then `text()` pair always threw on the second read, so the catch swallowed it and every non-JSON error collapsed to `HTTP nnn`. `responseDetail` already parses a string and falls back to it verbatim, so reading `text()` once covers both shapes.
- **PNA preflight header.** Correctly placed after the `allowedOrigin` guard, so it only goes to origins listed in `--cors`, and only on the preflight. Worth knowing the limit: https Pages to a private IP over http is still blocked as mixed content whatever this header says — it buys the localhost case, where Chrome treats the origin as trustworthy.
- **Port field.** Verified: clearable, digits only, letters rejected, a pasted `  8080  ` trimmed. `pattern="[0-9]*"` is inert without a form — the `onChange` guard is what enforces it.
- **`type="button"`.** No forms on the page, so no behaviour change, but it matches every other button in the panel.

One nit left alone: `assert.equal(api.includes('const text = await response.text()'), false)` bans a string rather than the shape, so any future legitimate `text()` read in `api.ts` trips it.

PR #99, opened 2026-07-31
Add OMP extension-backed actions

### LINKED ISSUE #90: Feature request: message actions
A user should be able to access some menu by right-clicking or long-pressing a message. All messages should have a "Copy as markdown" option. Agent messages should also have revert options, like "Revert conversation only" and "Revert code and conversation".

Baylar55:
## Summary

Adds the OMP-only proof of concept discussed in #90 for discovering and invoking optional host extension actions through a generic bridge API.

- capture session-scoped ACP `available_commands_update` catalogs
- expose generic `GET /session/:id/action` and `POST /session/:id/action/:name` endpoints
- recognize Undo/Redo only when the active `omp acp` runtime advertises both extension commands
- invoke actions without recording `/undo` or `/redo` as user prompts
- refresh authoritative conversation/session state after execution
- keep initial Redo availability in bridge-local, session-specific state
- integrate the existing message menu with the structured action endpoints
- keep `@baylarsadigov/omp-undo-redo` as an optional host-side plugin, not a frontend dependency

## State model and limitations

This deliberately implements the agreed proof-of-concept boundary:

- Redo becomes enabled only after an Undo produces an observed history change.
- Redo is cleared after Redo, a new prompt, or another observed session-history change.
- State is bridge-local, so restarting the bridge or invoking commands elsewhere can temporarily lose synchronization.
- The extension provides sequential Undo/Redo; this does not add arbitrary targeted revert-to-message or a selectable conversation-only mode.
- PI remains unchanged and can be investigated separately.

The action-provider registry and HTTP contract are generic so other known bridge extensions can be added without redesigning the frontend API.

## Verification

- `cd bridge && npm test` — 55 tests passed
- Web i18n, UI, settings, model, event, profile, and configuration regression checks passed
- `cd web && npm run build` passed
- Real OMP smoke test with `@baylarsadigov/omp-undo-redo@1.0.32`: first action discovery returned Undo enabled and Redo disabled; the browser message menu showed Undo and correctly hid Redo

Closes #90 only partially: this contributes extension-backed sequential Undo/Redo, while the issue's targeted revert variants remain separate work.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -88,10 +88,12 @@ The bridge implements a deliberate subset of the app's API:
 **Implemented:** `/v1/health`, `/global/health`, `/v1/capabilities`, `/v1/events`, `/global/event`,
 `/session` (list and create), `/v1/sessions`, `/experimental/session`, `/session/status`, `/path`,
 `/file`, `/command` (empty), `/agent` (empty), `/config/providers`, and on a session:
-`message`, `todo`, `diff` (empty), `prompt_async`, `abort`.
+`message`, `todo`, `diff` (empty), `prompt_async`, `abort`, rename, delete, plus generic action
+discovery and invocation at `action` and `action/{name}`. OMP currently maps the optional
+`omp-undo-redo` extension onto the generic action API only after ACP advertises both commands.
 
 **Not implemented — anything else 404s**, including `/question` and its replies, `/project/current`,
-`/vcs`, `/file/status`, `/session/{id}/command`, and renaming or deleting a session.
+`/vcs`, `/file/status`, and `/session/{id}/command`.
 
 When you add a call, pick one of two patterns already used in the codebase:
 
--- a/README.md
+++ b/README.md
@@ -76,6 +76,9 @@ Depending on the harness:
 - review changed files and their diffs — OpenCode
 - rename and delete sessions — OpenCode changes them in the harness; on OMP, PI and Claude Code the
   same controls keep a bridge-local nickname and hide the session from that bridge only
+- use Undo and Redo for OMP sessions when the host has loaded the optional
+  `@baylarsadigov/omp-undo-redo` extension; the bridge discovers and invokes it without bundling it
+  into the app
 
 ## Desktop Mode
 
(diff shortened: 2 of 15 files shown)


giuliastro:
Thanks, this is well aligned with the proposed design. I found one issue to fix before merging: `invokeAction()` compares the full serialized message objects, but ACP replay regenerates local timestamps, so an Undo that makes no history change can still return `applied: true`. I reproduced it by running a successful Undo followed by another Undo at the boundary; the second one returned `true` instead of `false`.

Could you compare a stable semantic history representation that excludes timestamps/volatile IDs, add a regression test for the no-op case, and show a visible message in the frontend when `applied` is `false`? After that, this should be ready.

Baylar55:
Thanks for reproducing this. You’re right: the current comparison includes replay-generated timestamps and IDs, so it can report a semantic no-op as applied.

I’ll update it to compare a stable history representation containing only ordered conversation semantics—message roles and meaningful part content—while excluding timestamps, message/part/call IDs, and other replay-generated fields.

I’ll also add:

- A regression test where ACP replays identical history with regenerated timestamps/IDs and a second Undo at the boundary returns `applied: false`.
- A localized informational message in the conversation UI when an action is not applied, such as “Nothing to undo in this session” or “Nothing to redo in this session.”

I’ll push the fix to this PR and let you know when it is ready for another look.

Baylar55:
Fixed in `7170791` and pushed to this PR.

- Replaced full message serialization with a stable semantic history signature. It compares message roles and ordered meaningful part content, canonicalizes object keys, and excludes replay-generated message/part/call IDs and timestamps while preserving semantic tool input/output.
- Added a regression that performs a successful Undo, regenerates replay IDs/timestamps, then performs another Undo at the boundary and asserts `applied: false`. Redo remains enabled from the first successful Undo.
- Added localized informational feedback for no-op actions: “Nothing to undo in this session.” / “Nothing to redo in this session.” The detail view renders this as an info notice rather than an error.

Verification:

- Bridge: 55 tests passed
- Web i18n, UI, settings, model, events, profiles, and config checks passed
- TypeScript/Vite production build passed
- Browser smoke check confirmed the no-op response renders the visible informational notice with no error

giuliastro:
Merged—thank you, @Baylar55! With `@baylarsadigov/omp-undo-redo` installed on the OMP host, the integration can now be tested on GitHub Pages: https://giuliastro.github.io/harness-remote/

giuliastro:
Thanks again for this contribution. While testing it with the extension installed, I found a bridge-side state-sync issue: we currently infer whether Undo/Redo succeeded from the reconstructed transcript, which is not always enough to represent OMP session-tree navigation or file-only changes.

I opened #101 with the details and a proposed next step: let the extension expose authoritative action availability/result state for the bridge to forward. Your input there would be very welcome.

Baylar55:
I opened new pull request #102 for #101

Issue #101, opened 2026-07-31
fix(bridge): make OMP extension action state authoritative


giuliastro:
## Problem

The OMP Undo/Redo integration currently derives whether an extension action was applied by comparing the bridge's reconstructed message history before and after sending `/undo` or `/redo`.

This is not an authoritative signal. `@baylarsadigov/omp-undo-redo` can successfully navigate the active session tree and restore file state while the visible text history is unchanged or incomplete. The prompt boundary is intentionally retained, and tool/file-only turns may have no text change for the bridge to compare.

As a result:

- Undo can succeed in OMP but the UI does not consistently remove the expected content.
- The bridge may return `applied: false`, so its bridge-local Redo state remains disabled and **Redo is never offered**.
- On reload, externally loaded OMP sessions can be reconstructed from the append-only JSONL history rather than the active session-tree branch, making the displayed conversation inconsistent after navigation.

## Current implementation

- `bridge/src/acp-service.js` sets `applied` from a semantic message-history comparison.
- `bridge/src/extension-actions.js` enables Redo only after that inferred success.
- `bridge/src/omp-session-history.js` reads the append-only session log, which does not encode the currently selected tree branch.

## Expected behavior

The extension should be the authority for:

- whether an action was applied or was a no-op;
- the current availability of Undo and Redo for the session;
- a revision/token identifying the active session-tree state.

Harness Remote should render the active ACP session context after navigation and use the extension-provided action state directly, instead of inferring success from the visible message transcript.

## Proposed direction

Define a small optional bridge-extension action protocol that exposes structured action status/results, for example:

```json
{
  "actions": [
    { "id": "undo", "enabled": true },
    { "id": "redo", "enabled": false }
  ],
  "sessionRevision": "..."
}
```

Invoking an action should return the updated state and whether it was applied. Until that contract exists, the current integration should not treat a transcript diff as proof of success or failure.

Related: #90, #99.

PR #105, opened 2026-08-01
fix(bridge): support OMP undo/redo outside Git

### LINKED ISSUE #103: fix: OMP Undo/Redo bridge integration fails for session-only (non-Git) sessions
Hi @Baylar55 — thank you again for the quick follow-up in #102. I tested the merged fix against the installed extension, but the integration is still not usable for a session-only OMP workspace.

## Observed after updating and restarting the bridge

- **Undo does not reliably operate on the last turn**; the rendered session can jump to an earlier/wrong branch.
- **Redo is never offered** after Undo.
- The failure remains visible from Harness Remote after a full restart on the current `main`.

## Concrete environment finding

The OMP session being tested has a scratch-directory working tree that is **not a Git repository**:

```
C:\Users\giuli\AppData\Local\Temp\...\scratchpad\ompproj
git rev-parse --is-inside-work-tree
# fatal: not a git repository
```

The extension intentionally supports session-only undo outside Git, but the new bridge state loader cannot read authoritative state in that mode:

- `createOmpUndoRedoActionStateLoader()` first calls `git rev-parse --git-common-dir`.
- On this session it returns `undefined` (verified directly with the active session id).
- The durable schema-1 sidecar is under the Git common directory, so there is no state file for a non-Git session.

That leaves the bridge on its fallback action state: Undo is advertised from its default, while Redo stays disabled. After an action there is still no authoritative revision or action result to enable Redo.

## Why the rendered history can be wrong

Without `activeSessionLeaf`, `omp-session-history.js` falls back to `records.at(-1)` from the append-only JSONL file. A session-tree navigation does not rewrite that file, so the last physical record is not an authoritative representation of the currently selected branch after Undo/Redo. The bridge can therefore reconstruct and show the wrong branch even if OMP navigated its live session tree.

## Test gap

The new loader test mocks a valid Git common directory and a schema-1 sidecar. It does not cover the extension's supported session-only/no-Git mode, nor an end-to-end branched JSONL session where Undo changes the active leaf without a Git sidecar.

## Proposed next step

We need a runtime action-state contract that works in **both** extension modes, ideally exposed directly by the extension/ACP session rather than only through a Git-sidecar file:

- authoritative `undo` / `redo` availability;
- active session leaf and revision;
- action result/no-op result.

Until such state is available, the bridge should not synthesize usable Undo/Redo controls from defaults, and it should avoid reconstructing an “active” branch from the last JSONL record.

Could you please take a look? I am happy to test a patch against this exact non-Git session setup.

Related: #101, #102.

Baylar55:
## Summary

Fixes the remaining OMP Undo/Redo integration gap for session-only workspaces outside Git.

- consume the extension's PID-scoped schema-2 runtime action state from `@baylarsadigov/omp-undo-redo` v1.1.0+
- validate ACP PID, runtime ID, session hash, action availability, revision, active leaf, and invocation token
- prefer live runtime state while retaining the durable Git schema-1 sidecar as fallback
- hide actions when no authoritative state exists instead of synthesizing availability from defaults
- avoid treating the final append-only JSONL record as the active branch; use ACP replay when no authoritative leaf exists
- reject stale state from another process/runtime

## Why

In a non-Git scratch workspace, `git rev-parse --git-common-dir` cannot locate the old sidecar. That left Undo on fallback defaults, never enabled Redo, and could render an abandoned JSONL branch. Extension v1.1.0 now publishes equivalent runtime state independently of Git, so the bridge can use the actual live session tree.

## Testing

- `cd bridge && npm test`
- 63 tests passed
- added runtime non-Git, stale-runtime rejection, ACP PID propagation, unavailable-state replay, and authoritative-branch coverage

Closes #103. Related: #101, #102.

### DIFF
--- a/bridge/src/acp-client.js
+++ b/bridge/src/acp-client.js
@@ -33,6 +33,10 @@ export class AcpClient extends EventEmitter {
     return this.#agentInfo
   }
 
+  /** PID identifies extension runtime state published by this exact ACP process. */
+  get processID() {
+    return Number.isInteger(this.#child?.pid) ? this.#child.pid : undefined
+  }
 
   async start() {
     if (this.#child) return
--- a/bridge/src/acp-service.js
+++ b/bridge/src/acp-service.js
@@ -379,7 +379,7 @@ export class AcpService {
     const state = await loadExtensionActionState(
       this.#actionProviders,
       requireCommands ? this.#commandCatalogs.get(sessionID) ?? [] : undefined,
-      { sessionID, directory: session.cwd }
+      { sessionID, directory: session.cwd, processID: this.#acp.processID }
     )
     if (state) this.#authoritativeActionStates.set(sessionID, state)
     else this.#authoritativeActionStates.delete(sessionID)
--- a/bridge/src/extension-actions.js
+++ b/bridge/src/extension-actions.js
@@ -21,13 +21,22 @@ function availableProviders(providers, commands) {
 }
 
 export function listExtensionActions(providers, commands, state, busy = false, authoritativeState) {
-  return availableProviders(providers, commands).flatMap((provider) => provider.actions.map((action) => ({
-    id: action.id,
-    source: provider.id,
-    enabled: !busy && (authoritativeState?.source === provider.id
-      ? authoritativeState.actions.find((candidate) => candidate.id === action.id)?.enabled ?? false
-      : state.get(action.id) ?? action.enabledByDefault)
-  })))
+  return availableProviders(providers, commands).flatMap((provider) => {
+    if (authoritativeState?.source === provider.id) {
+      return provider.actions.map((action) => ({
+        id: action.id,
+        source: provider.id,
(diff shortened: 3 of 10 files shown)


Baylar55:
Implementation note: this consumes the runtime contract published by `@baylarsadigov/omp-undo-redo` v1.1.0 (`1d69ac8`). Git snapshot/restore behavior remains unchanged. In non-Git workspaces, the extension navigates OMP session history while correctly reporting that files are not restored. Runtime state now makes Undo/Redo availability, active leaf, revision, and no-op results authoritative in Harness Remote.

Baylar55:
If you want to support undo and redo for both conversation and file changes in non-git workspace, that will be a new feature. So for this I should update my extension first.

giuliastro:
Confirmed working in the real non-Git scratch-session setup after updating to `@baylarsadigov/omp-undo-redo@1.1.0` and restarting OMP and the bridge. Undo now follows the active branch and Redo is exposed correctly. Thank you for the thorough fix and the runtime-state contract, @Baylar55!

PR #107, opened 2026-08-01
Serve the harness command catalog so OMP slash commands reach the app

noamsiegel:
## What

`GET /command` returned a hardcoded `[]` and the OMP profile advertised `commands: false`, so the app hid the command surface for OMP entirely. Meanwhile OMP advertises **80 commands over ACP, 30 of them `skill:` entries**, and the bridge was already capturing them: `available_commands_update` lands in `#commandCatalogs` (`acp-service.js:754`), but only `actions()` ever read it.

This serves that catalog.

## Changes

- **`AcpService#commands(sessionID)`** maps the catalog to the app's `CommandInfo` shape and tags `skill:`-prefixed names as `source: "skill"`. The app already has a skill-only tab in Help -> Commands (`App.tsx:2095`), so this needed no UI work. Called without a `sessionID` it answers from the newest catalog, because the app loads commands before a session is selected.
- **`POST /session/:id/command`** sends `/name args` as prompt text. ACP defines no command channel; a slash-prefixed prompt is exactly what the composer would have sent, and OMP executes it. This route previously 404'd, so flipping the capability without it would have broken every command the app offered.
- **OMP profile reports `commands: true`**, plus the matching default in `web/src/backendCapabilities.ts`.

Worth flagging: the `pi` profile **already** declared `commands: true`, so PI users were being offered a picker that `/command` answered with an empty list. They now get the real catalog if their adapter advertises one. I could not test PI - I do not have it installed - so that half is reasoned from the profile table, not measured.

## Docs

README's "server slash commands ... intentionally unavailable" line for OMP, the feature bullet, and the CONTRIBUTING route inventory (`/command` "(empty)", `/session/{id}/command` listed as not implemented) all described the old behaviour and are updated.

## Tests

Two new tests in `bridge/test/server.test.js`, using a `SkillCommandAcp` extending the existing `ExtensionActionAcp`:

- catalog is served with skills tagged, and the session-less call returns the same list
- a picked command arrives as `/skill:memory status`; a command with no arguments sends `/model` with no trailing space; a body without `command` is a 400

Both were mutation-checked: dropping the skill tagging and always appending the argument separator each fail exactly these two tests. One existing assertion, `ompCapabilities.commands === false`, is updated since it pinned the behaviour this PR changes.

`node --test` in `bridge/`: 60 pass, 0 fail. All 7 `web/` suites pass, and `tsc -b && vite build` is clean.

## Verified against a live harness

OMP 17.2.2, bridge reachable over Tailscale:

```
GET /command?sessionID=...  ->  79 entries: 30 skills, 49 commands
GET /command                ->  79 entries
```

and the deployed PWA at giuliastro.github.io/harness-remote, pointed at the patched bridge, lists all 30 `/skill:*` entries under Help -> Commands. Before the patch that page was empty for OMP.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -87,13 +87,14 @@ The bridge implements a deliberate subset of the app's API:
 
 **Implemented:** `/v1/health`, `/global/health`, `/v1/capabilities`, `/v1/events`, `/global/event`,
 `/session` (list and create), `/v1/sessions`, `/experimental/session`, `/session/status`, `/path`,
-`/file`, `/command` (empty), `/agent` (empty), `/config/providers`, and on a session:
-`message`, `todo`, `diff` (empty), `prompt_async`, `abort`, rename, delete, plus generic action
-discovery and invocation at `action` and `action/{name}`. OMP currently maps the optional
+`/file`, `/command` (the harness catalog from `available_commands_update`), `/agent` (empty),
+`/config/providers`, and on a session:
+`message`, `todo`, `diff` (empty), `prompt_async`, `command`, `abort`, rename, delete, plus generic
+action discovery and invocation at `action` and `action/{name}`. OMP currently maps the optional
 `omp-undo-redo` extension onto the generic action API only after ACP advertises both commands.
 
 **Not implemented — anything else 404s**, including `/question` and its replies, `/project/current`,
-`/vcs`, `/file/status`, and `/session/{id}/command`.
+`/vcs`, and `/file/status`.
 
 When you add a call, pick one of two patterns already used in the codebase:
 
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ Depending on the harness:
 
 - answer the questions the agent asks, options or free text, without leaving the app — OpenCode
 - follow todo/plan updates as the agent works — OpenCode, OMP, Claude Code
-- send server `/commands` — OpenCode
+- send server `/commands` — OpenCode, OMP
 - choose the agent a session runs as — OpenCode
 - review changed files and their diffs — OpenCode
 - rename and delete sessions — OpenCode changes them in the harness; on OMP, PI and Claude Code the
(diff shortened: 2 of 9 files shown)


giuliastro (changes requested):
Thanks for the thoughtful contribution and the thorough test coverage, @noamsiegel — much appreciated!

This is close, but the initial catalog load needs a fix: on startup the app calls /command before any session has received available_commands_update, so it gets []. Opening a session populates the catalog through actions(), but the UI never calls loadCommands() again, leaving Help → Commands empty. Please reload the catalog after loadSelected() (or initialize it server-side without a sessionID) and add coverage for the initial-load flow.

noamsiegel:
Good catch, and you were right that my own evidence was a race rather than a working path: the picker I saw populated only because an earlier session had already warmed the bridge's in-memory catalog.

Fixed in 9761a14 by reloading the catalog after `loadSelected()`, the app-side option you offered.

**Why there and not server-side.** Initializing without a `sessionID` would mean the bridge issuing an ACP `session/load` at startup against an arbitrary session, or creating a throwaway one, just to learn a list that is identical for every session on the machine. `loadSelected` is the shared seam all ten of its callers pass through, so a session opened from the list, restored on mount, or freshly created all recover with one line. It is gated on `capabilities.commands && commands.length === 0`, so a harness without commands and a catalog already loaded both cost nothing.

**Reproduced your scenario first**, cold bridge with a wiped `--state-dir`, cleared `localStorage`, phone-width layout:

- without the fix: Help → Commands reads **"No server commands available"** while `GET /command` on the same bridge returns **80** entries
- with the fix: the same flow lists `/security`, `/model`, and the Skills tab fills

Removing the line reproduces the empty page on demand, which is what the new `ui-regression` assertion pins, and it is checked both ways rather than only in the passing direction.

**Coverage for the initial-load flow**, as asked:

- `bridge/test/server.test.js` - "fills the command catalog once a session is loaded, so a cold bridge recovers": `/command` is `[]` before any load, then returns the catalog after the session loads. Mutating the session-less fallback to `[]` fails it.
- `web/src/ui-regression.test.mjs` - asserts the refetch exists *and* that it lives inside `loadSelected` rather than in individual callers. Deleting the line fails it.

`node --test` in `bridge/`: 61 pass, 0 fail. All 7 web suites pass, `tsc -b && vite build` clean.

One thing I could not verify and do not want to imply I did: I have no PI or Claude Code install, so the `pi` profile's pre-existing `commands: true` and whether those adapters advertise a catalog remain untested by me.

PR #110, opened 2026-08-05
docs: document non-Git file undo/redo support

Baylar55:
## Summary

- document conversation and supported file restoration in non-Git workspaces with `@baylarsadigov/omp-undo-redo` v1.2.0+
- record that v1.2.0 retains the schema-2 runtime action-state contract introduced in v1.1.0
- clarify that file restoration remains extension-owned, so Harness Remote needs no alternate bridge action path or package dependency

## Verification

- `cd bridge && npm test`
- 66 tests passed

Related to #103 and #105.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,10 +76,11 @@ Depending on the harness:
 - review changed files and their diffs — OpenCode
 - rename and delete sessions — OpenCode changes them in the harness; on OMP, PI and Claude Code the
   same controls keep a bridge-local nickname and hide the session from that bridge only
-- extend bridge-backed harnesses through optional host extensions: bridge discovers compatible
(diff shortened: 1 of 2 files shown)


Baylar55:
Implementation note: I verified this against local `omp-undo-redo` commit `483e73b`, published as `@baylarsadigov/omp-undo-redo` v1.2.0. Non-Git file restoration is implemented inside the extension through blob checkpoints. Runtime marker schema 1 and action-state schema 2 remain unchanged, so Harness Remote's existing action invocation and authoritative session reload already consume the new behavior. This PR therefore updates dependency and feature documentation only; no bridge or web code change is required.

PR #121, opened 2026-08-09
feat: attach images to a prompt on the OMP bridge

luccapinto:
## Why

The app can drive OMP from a phone but not show it anything. A screenshot of a stack trace, a photo of a whiteboard, a design mock — all of it has to be described in words, or the session has to move back to a desktop. OMP itself is ready for this: `initialize` advertises `agentCapabilities.promptCapabilities.image` (measured against `omp acp` 17.2.10), and the bridge was discarding that capability.

Two things were missing, in two layers:

- `bridge/src/server.js` read `parts.find(part => part.type === "text")` and dropped every other part, and `acp-service.js` built `prompt: [{ type: "text", text }]` unconditionally;
- `web/src/` had no way to pick a file. `attachment?: boolean` in `api.ts` was only the model capability flag, and `MessagePartView` rendered `text`, `reasoning`, `tool` and `patch` — never a file part.

The wire format already carried `parts[]`, so this adds no protocol. It stops throwing parts away.

## What it does

Paperclip in the composer → OS picker (camera, gallery or file browser) → chips above the input, removable → sent as `file` parts in the shape the app already speaks to OpenCode:

```json
{ "type": "file", "mime": "image/jpeg", "filename": "error.jpg", "url": "data:image/jpeg;base64,..." }
```

The bridge validates and forwards them as ACP `image` blocks. An image with no caption is a complete prompt — a bare screenshot is the most common thing to send from a phone. Attachments also render in the transcript, and survive reopening a session.

Images are resized client-side to a 1568px longest edge. A phone photo is several megabytes of pixels no model needs, and it is paid for twice: once on a mobile uplink, once in tokens. A PNG that needed no resizing stays PNG so screenshot text stays crisp; everything else is encoded as JPEG, which also normalises whatever the picker returns, HEIC included.

Limits: 8 attachments per prompt, 5MB each, 15MB total, `image/png|jpeg|webp|gif`. Rejections name what is accepted and **preserve the typed text** — losing a long prompt because an attachment was refused is the kind of failure that makes someone stop using a tool.

## Every change lives on two backends

Gated, not assumed. The bridge accepts an attachment only when the agent advertised `promptCapabilities.image`; PI, Claude Code and Codex CLI do not, so `/v1/capabilities` reports it off, the composer hides the picker, and the bridge refuses with an explicit message instead of letting the agent fail mid-turn. `bridge/test/server.test.js` covers that path with a fake that advertises no image support.

## Both layouts

Checked either side of 781px: phone at 420px and desktop at 1280px. The composer now uses the `.composer-bar` / `.composer-chips` / `.composer-actions` / `.composer-send` classes that were already in `styles.css` but unwired, so the attach button sits left of a right-aligned send button in both. Thumbnails are bounded (`max-width: 100%`, `max-height: 22rem`) — an unbounded photo would widen the page, which is the failure the reasoning and tool blocks already guard against.

## The bridge is a network service

- **Auth and `--root` are untouched.**
- **CORS is untouched.**
- **`readBody` rises from 1MB to 25MB**, which is the one security-relevant change here. It was necessary: a single phone photo in base64 exceeds 1MB. The real ceilings are the attachment limits above, enforced before anything reaches the agent, and payloads are validated per attachment rather than trusted by total size alone.
- **Nothing is written to disk.** Images travel as base64 inside the ACP prompt, so there is no temp directory to clean up or leak.

## How this was verified

Against a **real harness**, not only the fakes — `omp acp` 17.2.10 on Ubuntu 26.04, Node 22.

- Sent a solid magenta PNG with "reply with only the colour name". OMP answered `Magenta`, which is the only proof that matters: the image reached the model.
- Repeated the same flow through the built PWA in a real browser, driving the OS picker, so the canvas resize and the upload path were exercised end to end, not just the API.
- Intercepted the outgoing request to confirm the payload shape, and confirmed the composer clears its chips on success and restores both text and attachments on failure.
- Reopened the session against a bridge with a pristine state directory and confirmed the thumbnail returns, stable across repeated polls.
- `bridge/test`: 84 pass. Web suites: 10 pass (including the new `test:attachments`). `npm run build`, `npm run build:electron`, `npm run test:desktop` (14) all pass.

Measuring the real harness produced four findings that are now in `docs/DEPENDENCIES.md`, because none of them could be guessed: OMP **re-encodes images to WebP** (a PNG upload replays as `image/webp`, so the mime must come from the record, never from the upload), it **stores no filename**, it **mints a fresh `messageId` on every `session/load`**, and it replays a stored image as a `user_message_chunk` sharing the text chunk's `messageId`.

Image chunks are mapped **only while replaying**: a live turn already records its own attachment in `#recordPrompt`, so accepting one there would draw the same thumbnail twice.

## One caveat worth stating

While testing I saw a user message appear twice in a session that several bridge processes had loaded with divergent state. It reproduces on `v2.10.3` conditions too and comes from `mergeExternalHistory`, which is untouched here: it dedupes by `info.id` — the JSONL loader uses record ids while ACP replay mints UUIDs, so they never match — or by signature within a 30s window, while the loader's timestamps are hours older than a fresh replay's. Under clean startup both `v2.10.3` and this branch return the same stable history. This change cannot introduce it; it only makes the two copies visually distinguishable when it happens, since one carries the thumbnail. Happy to open a separate issue with the trace if useful.

## Not included

`test:attachments` should join the suite list in `.github/workflows/gh-pages.yml`, but pushing a workflow change needs an OAuth scope this fork's token does not have, so I left CI alone rather than half-applying it. One line, wherever you prefer it in that list:

```yaml
          npm run test:attachments
```

Audio is deliberately out of scope: ACP exposes no audio prompt block, so voice would have to be transcribed before it becomes a prompt. That is a different design.


giuliastro:
Thanks for the thorough work here — the backend validation and replay handling look solid. One issue I noticed before merging: the PR description says PI, Claude Code and Codex CLI hide the paperclip when the harness does not advertise image support, but the current UI renders the attachment button unconditionally (it is only disabled when there is no selected session or the max attachment count is reached).

Could you gate the paperclip on the actual image/prompt capability exposed by the bridge, so unsupported harnesses do not offer an attachment flow that will only fail at send time with `This harness does not accept images`?

I think this should be a small frontend/capability plumbing fix, but it would make the implementation match the behavior described in the PR and avoid a confusing UX on non-image-capable ACP backends.

luccapinto:
Correct diagnosis, and the description was wrong rather than the code being subtly off: only the server-side refusal existed. `/v1/capabilities` served `profile.capabilities` verbatim, that table has no attachment field, and the button's only condition was `!selectedSession || attachments.length >= ATTACHMENT_MAX_COUNT`. So on PI, Claude Code and Codex CLI the flow was exactly the one you describe — pick an image, send, get `This harness does not accept images`.

Fixed in `e801793`.

**Bridge.** `/v1/capabilities` now derives the flag from the agent's own handshake instead of the profile table:

```js
await acp.start()
writeJSON(response, 200, { ...profile.capabilities, attachments: Boolean(acp.promptCapabilities?.image) })
```

`AcpClient` retains `agentCapabilities.promptCapabilities` from `initialize`, alongside `agentInfo`. Deriving it rather than adding `attachments: true` to the OMP profile is deliberate: it is the agent that decides, so an OMP release that stopped advertising images would stop offering the control instead of failing at send time — which is the same failure mode you just caught, one layer down. `await acp.start()` matches what `/v1/health` already does, and the client memoises startup, so this costs no extra process.

**App.** The picker and its hidden input are wrapped in `capabilities.attachments && (…)`, the same gate style already used for rename, delete, models and todos. `HarnessCapabilities` gained the field and all five entries in `DEFAULT_HARNESS_CAPABILITIES` state it explicitly — `false` everywhere, including OpenCode, which has no attachment path in this PR — so a failed capability fetch cannot enable the control by omission. Staged attachments are also cleared when the connection changes, since a staged image belongs to the server it was staged on and the next one may not take images at all.

**Verified both directions, not just the happy one.** Against real `omp acp` 17.2.10: `attachments: true`, paperclip and file input present. Against a stub ACP agent advertising `promptCapabilities: { embeddedContext: true }` and no image: `attachments: false`, and both the button and the `input[type=file]` absent from the DOM — checked at 420px and 1280px, with the send button still present in both cases. Server-side refusal still returns `400 This harness does not accept images` for a client that ignores the flag.

`bridge/test/server.test.js` covers the reporting in both directions using a fake with and without the capability, and the web suite asserts the gate exists and that every backend default states its attachment support. Full run: bridge 85, web 10 suites, `npm run build`, `npm run build:electron`, `npm run test:desktop` 14.

Still outstanding from the description: `npm run test:attachments` is not in `.github/workflows/gh-pages.yml`, because pushing a workflow change needs an OAuth scope this fork's token lacks. One line in that list whenever you want it.

giuliastro:
Thanks for the contribution and for turning the feedback around so quickly. The capability gating looks good now, and the extra testing notes were very helpful. I've merged the PR and also added the attachment regression test to the Pages workflow. Appreciate the thorough work on this.

PR #125, opened 2026-08-10
fix(web): raise message and session action menus above the composer

### LINKED ISSUE #124: Copy menu (long-press / right-click) is covered by the composer — taps land on the input box
## Bug

Both layouts are affected — long-press on the mobile layout, right-click on the desktop/web layout: the message context menu can be covered by the composer (input box). When the menu opens over the bottom of the transcript it overlaps the composer, and because both elements stack at the same z-index and the composer comes later in the DOM, the composer paints on top of it: the top menu items ("Copy text" / "Copy Markdown") are hidden behind the input box and taps/clicks land on the composer instead of the menu.

Image

## Cause

Both elements stack at the same level:

- `.composer` — `z-index: var(--z-sticky)` (20); `position: fixed` on mobile (floating above the bottom nav), `position: relative` in flow on desktop
- `.message-context-menu` — hardcoded `z-index: 20`, `position: fixed`, rendered inside the message bubble (earlier in the DOM than the composer)

With equal z-index the browser falls back to DOM order, and the composer comes after the message list, so it wins the tie and covers the top of the menu.

## Steps to reproduce

1. Open a session with a long conversation
2. Scroll to a message near the bottom of the transcript
3. Long-press the message (mobile layout) or right-click it (desktop/web layout) — the menu opens over the composer and its top items are covered

## Expected vs actual

- Actual: the top of the menu (the copy items) is covered by the input box — see the screenshot above
- Expected: the menu renders above the composer

Image

## Suggested fix

Replace the hardcoded `z-index: 20` with the existing `--z-menu` token (30) on `.message-context-menu` (and, for consistency, on `.session-actions-menu`, which uses the same hardcoded value). A PR with that change follows this issue.

joyfish666:
## What

Raise the message context menu (long-press on a phone, right-click on desktop) above the composer.

The composer stacks at `z-index: var(--z-sticky)` (20) — `position: fixed` floating over the transcript on mobile, `position: relative` below the transcript on desktop. The message context menu hardcoded `z-index: 20` and, being rendered earlier in the DOM, lost the tie: whenever the menu opened over the bottom of the transcript (mobile bottom sheet, or a right-click near the composer on desktop), its top items ("Copy text" / "Copy Markdown") were covered by the input box and taps/clicks landed on the composer instead of the menu.

## Changes

- `web/src/styles.css`
  - `.message-context-menu`: `z-index: 20` → `z-index: var(--z-menu)` (30), so it stacks above the composer and every sticky element
  - `.session-actions-menu`: same change for consistency — the header actions dropdown was on the same hardcoded value
- `web/src/ui-regression.test.mjs`: update the assertion that pinned the literal `z-index: 20`

## Verification

- `npm run test:ui` — UI regression tests pass
- `npm run build` — tsc + vite production build pass
- Manual check on a phone-width window and on desktop: the message menu now renders above the composer in both layouts; the header "⋮" menu is unchanged

Fixes #124

### DIFF
--- a/web/src/styles.css
+++ b/web/src/styles.css
@@ -2582,7 +2582,7 @@ a:focus-visible,
 
 .message-context-menu {
   position: fixed;
-  z-index: 20;
+  z-index: var(--z-menu);
   display: grid;
   min-width: 14rem;
   gap: 1px;
@@ -2662,7 +2662,7 @@ a:focus-visible,
   position: absolute;
   top: calc(100% + var(--space-1));
   right: 0;
-  z-index: 20;
+  z-index: var(--z-menu);
   display: grid;
   min-width: 14rem;
   gap: 1px;
--- a/web/src/ui-regression.test.mjs
+++ b/web/src/ui-regression.test.mjs
@@ -523,7 +523,7 @@ assert.match(
 )
 assert.match(app, /session-actions-menu/, 'the header actions menu should have its own styles')
 assert.match(styles, /\.session-actions-menu\s*\{[\s\S]*?position:\s*absolute/, 'the header actions menu must overlay the conversation rather than push its layout')
-assert.match(styles, /\.session-actions-menu\s*\{[\s\S]*?z-index:\s*20/, 'the header actions menu must stack above the message list')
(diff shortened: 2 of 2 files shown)


giuliastro:
Thanks @joyfish666! Nice clean fix — using the shared `--z-menu` token here is exactly the right approach, and thanks for updating the regression test as well. Merged! 🙌

PR #140, opened 2026-08-12
feat(core): introduce normalized AgentRun model

### LINKED ISSUE #131: feat(core): introduce a normalized AgentRun and attention model
## Why

Harness Remote currently exposes sessions primarily through backend-specific session state. That is sufficient for a remote client, but the next product layer needs a backend-agnostic concept representing **work being performed by an agent**.

The UI should eventually be able to answer questions such as:

- Which agents are currently working?
- Which run needs the user's attention?
- Which runs failed or completed recently?
- Which project and machine does a run belong to?

Those questions should not require every screen to understand OpenCode, OMP, PI, Claude Code or Codex-specific session semantics.

This issue introduces the normalized model only. It should remain compatible with the current session UI and derive its state from data already available today.

Depends on / should coordinate with #130.

## Proposed model

The exact TypeScript shape is intentionally open to implementation, but the abstraction should cover concepts equivalent to:

```ts
type AgentRunStatus =
  | "idle"
  | "working"
  | "waiting"
  | "retrying"
  | "completed"
  | "failed"
  | "stopped"

type AttentionReason =
  | "permission"
  | "question"
  | "failure"
  | "completion"
  | "waiting"

type AgentRun = {
  id: string
  backend: BackendKind
  sessionId: string
  status: AgentRunStatus
  attention?: AttentionReason
  projectId?: string
  machineId?: string
  startedAt?: number
  updatedAt?: number
}
```

Names and fields may change if the implementation finds a cleaner model.

## Scope

- Define a backend-agnostic `AgentRun`-level representation derived from existing session/backend data.
- Normalize current session states such as idle, busy, retry and waiting.
- Define a small attention model that can later drive an inbox.
- Keep a stable reference back to the underlying backend/session so existing actions continue to work.
- Add tests covering normalization across OpenCode and ACP-backed sessions.
- Design the model so optional future `projectId` and `machineId` fields can be added/populated without changing its purpose.

## Important design rule

`AgentRun` is **not** a replacement for the underlying session payload.

Backend/session-specific information should remain available for detailed views and actions. `AgentRun` is the normalized operational layer used for cross-agent status, sorting and attention.

## Non-goals

- no project discovery yet
- no multi-machine daemon yet
- no task launcher
- no push notifications
- no agent routing
- no requirement to synthesize permissions/questions from harnesses that do not expose them
- no major UI redesign in this issue

## Acceptance criteria

- [ ] A normalized run model exists independently of backend-specific session types.
- [ ] Existing `idle`, `busy`, `retry` and `waiting` semantics map consistently into it.
- [ ] A run can identify whether it currently needs user attention when the underlying harness exposes enough information.
- [ ] The model retains the backend/session identity required to open and control the existing session.
- [ ] Normalization is tested against OpenCode and at least one ACP-backed harness.
- [ ] Existing UI behavior remains functional.
- [ ] The design can accommodate future project and machine identity without coupling this issue to their implementation.

## Follow-up

The first consumer of this model should be the Agent Inbox: a cross-backend view organized by `Needs attention`, `Working`, and recent terminal states.

giuliastro:
## Summary

Introduces the backend-agnostic `AgentRun` projection planned in #131, without changing the existing session UI.

### What changed

- add `AgentRunStatus`, `AgentAttention`, `AgentRun` and `AgentRunSignals`
- normalize current harness session states (`idle`, `busy`, `retry`, `waiting`) into operational states
- retain the backend + session identity needed to open/control the underlying session
- surface actionable question/permission attention with the original request id
- support explicit future terminal signals (`completed`, `failed`, `stopped`) without inventing them from current harness data
- reserve optional project/machine/start metadata for the later control-plane phases
- add normalization coverage across OpenCode and ACP-backed harness examples
- run the new suite in PR checks

## Design notes

`AgentRun` is deliberately a projection rather than a replacement for `SessionView`. Backend-specific details and capabilities remain on the existing session path.

`waiting` is normalized as an operational state but is not treated as user attention: waiting for internal/subagent work does not itself require intervention.

When multiple actionable signals exist, permission takes precedence over question so a future Inbox can surface the most immediately blocking action. The request id is retained so that Inbox can later become actionable without reshaping this model.

## Validation

The PR workflow runs:

- TypeScript build/type-check
- the new `test:agent-runs` suite
- all existing web regression suites
- bridge tests
- debug APK build

Closes #131.

### DIFF
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -60,6 +60,7 @@ jobs:
           npm run test:profiles
           npm run test:markdown
           npm run test:attachments
+          npm run test:agent-runs
           npm run test:platform
         working-directory: web
 
--- a/web/src/agentRuns.ts
+++ b/web/src/agentRuns.ts
@@ -1,0 +1,111 @@
+import type { BackendKind, PermissionRequest, QuestionRequest, SessionView } from "./types"
+
+export type AgentRunStatus =
+  | "idle"
+  | "working"
+  | "waiting"
+  | "retrying"
+  | "completed"
+  | "failed"
+  | "stopped"
+
+export type AgentAttention =
+  | { reason: "permission"; requestId: string }
+  | { reason: "question"; requestId: string }
+  | { reason: "failure" }
+  | { reason: "completion" }
+
+export type AgentRun = {
+  id: string
+  backend: BackendKind
+  sessionId: string
+  title: string
+  directory: string
+  status: AgentRunStatus
+  attention?: AgentAttention
+  projectId?: string
+  machineId?: string
+  startedAt?: number
+  updatedAt?: number
+}
+
+export type AgentRunSignals = {
+  questions?: readonly Pick<QuestionRequest, "id" | "sessionID">[]
+  permissions?: readonly Pick<PermissionRequest, "id" | "sessionID">[]
+  terminalStatus?: Extract<AgentRunStatus, "completed" | "failed" | "stopped">
+  projectId?: string
+  machineId?: string
+  startedAt?: number
+}
+
+const WORKING_STATUSES = new Set(["busy", "working", "running"])
+const RETRYING_STATUSES = new Set(["retry", "retrying"])
+const WAITING_STATUSES = new Set(["waiting"])
+const COMPLETED_STATUSES = new Set(["completed", "complete", "done", "success", "succeeded"])
+const FAILED_STATUSES = new Set(["failed", "failure", "error"])
+const STOPPED_STATUSES = new Set(["stopped", "cancelled", "canceled", "aborted"])
+
+/**
+ * Normalize the status vocabulary exposed by individual harnesses into the operational states used
+ * by the control-plane layer. The currently supported session list uses idle/busy/retry/waiting;
+ * the additional aliases keep the boundary tolerant of richer lifecycle data without teaching UI
+ * consumers about backend-specific words later.
+ */
+export function normalizeAgentRunStatus(status: string, terminalStatus?: AgentRunSignals["terminalStatus"]): AgentRunStatus {
+  if (terminalStatus) return terminalStatus
+
+  const normalized = status.trim().toLowerCase()
+  if (WORKING_STATUSES.has(normalized)) return "working"
+  if (RETRYING_STATUSES.has(normalized)) return "retrying"
+  if (WAITING_STATUSES.has(normalized)) return "waiting"
+  if (COMPLETED_STATUSES.has(normalized)) return "completed"
+  if (FAILED_STATUSES.has(normalized)) return "failed"
+  if (STOPPED_STATUSES.has(normalized)) return "stopped"
+  return "idle"
+}
+
+function attentionFor(
+  sessionId: string,
+  status: AgentRunStatus,
+  signals: AgentRunSignals
+): AgentAttention | undefined {
+  const permission = signals.permissions?.find((request) => request.sessionID === sessionId)
+  if (permission) return { reason: "permission", requestId: permission.id }
+
+  const question = signals.questions?.find((request) => request.sessionID === sessionId)
+  if (question) return { reason: "question", requestId: question.id }
+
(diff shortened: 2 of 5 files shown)


web/src/agentRuns.ts
```diff
@@ -0,0 +1,111 @@
...
+  machineId?: string
+  startedAt?: number
+}
+
+const WORKING_STATUSES = new Set(["busy", "working", "running"])
+const RETRYING_STATUSES = new Set(["retry", "retrying"])
+const WAITING_STATUSES = new Set(["waiting"])
+const COMPLETED_STATUSES = new Set(["completed", "complete", "done", "success", "succeeded"])
+const FAILED_STATUSES = new Set(["failed", "failure", "error"])
+const STOPPED_STATUSES = new Set(["stopped", "cancelled", "canceled", "aborted"])
```
giuliastro: These three sets contradict the design note above them. The summary says terminal signals are supported "without inventing them from current harness data", but these map raw harness status strings straight onto terminal states — which is inventing them from exactly that data.

Nothing triggers it today: `acp-service.js` only ever emits `busy` or `idle`, and the OpenCode path adds `retry` and `waiting`, so none of `done`, `success`, `error`, `aborted` can currently appear. But the sets exist to be tolerant of richer vocabularies, so the failure arrives on its own the first time a harness reports one.

The case that worries me is `"error"` in `FAILED_STATUSES`. A harness reporting a transient, recoverable error would promote the run to a terminal `failed`, and `attentionFor` would then raise `{ reason: "failure" }` — a passing hiccup surfaced in the Inbox as a finished, failed run, with no path back to `working` once the next turn starts.

These are also the only mappings the suite does not cover: the test exercises `idle`, `busy`, `retry`, `waiting`, `BUSY` and an unknown value, never these aliases.

Two coherent options, and I would take the first:

1. Drop the three sets. Terminal states then come only from the explicit `terminalStatus` signal, which is what the summary describes, and an unrecognized harness word keeps falling through to `idle`.
2. Keep them, correct the summary to say terminal states can also come from the harness vocabulary, and add cases for each alias so the promotion is deliberate rather than incidental.

---

.github/workflows/pr-checks.yml
```diff
@@ -60,6 +60,7 @@ jobs:
           npm run test:profiles
           npm run test:markdown
           npm run test:attachments
+          npm run test:agent-runs
```
giuliastro, line 63: Worth adding the same line to `gh-pages.yml`, which runs its own copy of this list and is the gate on the Pages deploy. As it stands `test:agent-runs` protects pull requests but not the deploy, so a change that breaks it after merge would ship the PWA anyway.

That drift is what the comment a few lines above this block is about — "Keep the two in step: a suite that runs only after the merge is the situation this workflow exists to end" — and it works in this direction too: a suite that runs only *before* the merge leaves the deploy ungated.

One line in `gh-pages.yml`, next to `npm run test:attachments`.

---

giuliastro (commented):
Reviewed the four files against the current `main`. The shape is sound and the central claim holds: I checked the status vocabulary this normalizes against, and `App.tsx:224` treats exactly `busy`, `retry` and `waiting` as active with `idle` as the default, while `acp-service.js` only ever emits `busy` or `idle`. So the mapping matches what the app actually produces today.

The projection reads well. Keeping `backend` and `sessionId` on `AgentRun` means it stays a view rather than a replacement, permission-over-question precedence is the right call for a blocking-action Inbox and is covered by a test, `waiting` correctly raises no attention, and an unrecognized state fails to `idle` rather than throwing.

Two things below, neither blocking, and the PR is still a draft:

- the terminal-status alias sets contradict the design note in the summary, and are the only mappings the suite does not cover
- `test:agent-runs` is added to `pr-checks.yml` but not to `gh-pages.yml`, so the new suite gates pull requests but not the deploy

One observation on validating this one: since nothing imports `agentRuns.ts` yet, the debug APK from this branch is byte-for-byte the behaviour of `main`. Installing it proves nothing about the change — the test suite is the only evidence here, which raises the stakes on the untested aliases above.

---

Issue #143, opened 2026-08-12
P0: make one machine daemon reliable, universal and restart-stable


giuliastro:
## Goal

Make the daemon a reliable **one machine = one endpoint = one credential policy = all discovered local agents** runtime.

This is the first P0 stabilization issue. Nothing task/fleet-related should be considered product-ready until this works repeatedly on a real device.

### Stable identity and authentication
- machine identity persists across ordinary restarts;
- explicit `--username` + `--password` are stable because the user chose them and are never rewritten or persisted elsewhere by Harness;
- when credentials are omitted, the daemon generates an ephemeral random password for that process, prints it clearly, and does **not** persist it;
- restart with generated credentials intentionally rotates the password;
- environment-provided credentials remain supported;
- no generated client password is stored on disk.

### Universal agent discovery
- discover supported installed agents safely;
- expose all usable ACP agents, not only one ACP primary;
- include OpenCode without requiring obscure flags;
- never silently default to an uninstalled backend;
- one broken agent must not make the machine unusable.

### Agent lifecycle
- states: `configured/starting/available/unavailable/unauthenticated` where applicable;
- preserve human-readable failure reason in the machine snapshot;
- slow startup is asynchronous and retryable rather than one long blocking timeout;
- already-running compatible services may be adopted when safe.

### Machine API
`/v1/machine` is the authoritative source for:
- stable machine identity;
- discovered agents;
- backend/transport/capabilities;
- state + failure reason;
- enough information for the client to avoid declaring a backend before discovery.

### Backward compatibility
- legacy direct/single-backend servers remain supported;
- no existing direct URL semantics are changed by this milestone;
- old clients can continue using existing paths while the machine-first client is introduced.

## Folded-in issues

This issue now owns the work previously tracked separately in **#177** (wrong backend default, OpenCode startup/adoption, incomplete multi-agent exposure).

## Acceptance test

On a real Termux/Debian or equivalent machine with PI + at least one second supported agent:

1. start daemon without credentials → a random password is printed and nothing is persisted;
2. restart without credentials → a different random password is printed;
3. start with explicit username/password, restart with the same command → the same chosen credentials work;
4. machine identity remains stable in both cases;
5. `/v1/machine` lists all detected agents;
6. one unavailable agent reports a reason but does not break the others;
7. no startup path silently selects an uninstalled agent.

Only after this passes do we move to the machine-first client issue.

giuliastro:
[...]

PR #148, opened 2026-08-12
feat(cli): add one-command startup for the existing bridge

### LINKED ISSUE #147: feat(cli): add one-command startup for the existing bridge
## Why

Before the Universal Daemon refactor in #143, Harness should remove the largest adoption blocker using the architecture that already exists.

The current setup requires users to understand per-backend bridge commands, host/port choices, credentials and platform/network details before they can evaluate the product. That is too much friction in a category where comparable tools reach a working state from one command.

This issue is the **P0A adoption win** and should ship before the larger daemon work.

## Target experience

```bash
npx harness-remote
```

or an equivalently simple installed command.

For the first milestone, the launcher may target **one detected/specified backend at a time**. It does not need the multi-agent registry from #143.

Expected behavior:

1. detect supported installed agents conservatively where practical;
2. choose a sensible backend automatically when exactly one supported option is available, otherwise present/require an explicit choice;
3. choose an available local port or a documented default with clear collision handling;
4. generate or reuse secure credentials when non-loopback access requires them;
5. start the existing bridge/server path with the resolved configuration;
6. print concise connection/setup information a phone/web client can use;
7. provide actionable errors rather than falling back to the long manual setup path silently.

## Scope

- thin CLI/launcher above the **existing bridge**;
- no Universal Daemon registry required;
- safe agent executable discovery/detection;
- backend selection rules;
- port selection/collision handling;
- credential generation/reuse policy;
- bridge startup and lifecycle forwarding;
- concise terminal output with connection information;
- explicit flags for users who need to override auto-detection;
- documentation for the new shortest path while keeping advanced/manual setup available.

## Security

- do not weaken the existing rule against unauthenticated non-loopback exposure;
- generated credentials must use cryptographically appropriate randomness;
- never print or persist agent-provider tokens;
- discovery must not execute arbitrary PATH entries merely to identify them;
- explicit user overrides remain available for unusual environments.

## Non-goals

- no multi-agent daemon;
- no multi-machine fleet;
- no QR pairing requirement;
- no task launcher/worktrees;
- no hosted relay;
- no removal of existing manual bridge commands.

## Acceptance criteria

- [ ] A fresh supported environment can start the existing Harness bridge through one command or an equivalently low-friction command path.
- [ ] At least one supported backend can be detected or selected without reading backend-specific setup instructions.
- [ ] Port collisions fail over or produce a clear actionable error.
- [ ] Authentication remains required for non-loopback exposure according to existing security policy.
- [ ] The launcher prints concise connection/client setup information.
- [ ] Advanced/manual bridge invocation remains supported.
- [ ] Automated tests cover detection/selection/config generation without requiring real provider credentials.

## Relationship to #143

This issue deliberately **does not wait for #143**. It takes the cheap adoption win against the current bridge first.

#143 then evolves the architecture into one machine-level multi-agent daemon without carrying the deadline for fixing first-run friction.

giuliastro:
## Summary

- add a root `harness-remote` launcher entry point over the existing bridge
- detect OMP / PI / Claude Code / Codex CLI executables conservatively without running them
- auto-select an unambiguous backend, otherwise require `--backend`
- default quick start to authenticated LAN binding with generated Basic Auth credentials
- auto-select the first free port starting at 4097 and give clear errors for explicit port collisions
- print concise connection details before starting the existing bridge
- preserve advanced bridge flags and the existing manual bridge path
- add unit coverage for discovery/selection/config forwarding
- document a real repository-level one-command path: `npx github:giuliastro/harness-remote`
- document the current OpenCode exception

## Why

This is the P0A adoption win from the revised Harness 3 roadmap: remove first-run friction **before** the larger Universal Daemon refactor in #143.

It deliberately does not implement the daemon or a multi-agent registry. The goal is to make the architecture that already exists much easier to evaluate.

## Notes

OpenCode remains direct HTTP in the current architecture and is not wrapped by this launcher.

The root package is intentionally private for now. `npx github:giuliastro/harness-remote` executes the repository package directly; this PR does not claim that an npm registry package has already been published.

Closes #147.


bridge/src/launcher.js
```diff
@@ -0,0 +1,180 @@
...
+  }
+}
+
+export function buildBridgeArgs(args, { backend, host, port, username, password }) {
+  const result = [...args]
+  if (!hasOption(result, "--backend")) result.push("--backend", backend)
+  if (!hasOption(result, "--host")) result.push("--host", host)
+  if (!hasOption(result, "--port")) result.push("--port", String(port))
+  if (!hasOption(result, "--username") && username) result.push("--username", username)
+  if (!hasOption(result, "--password") && password) result.push("--password", password)
```
giuliastro: The generated password goes to the child on its argv, which puts it in `ps` output and in `/proc/<pid>/cmdline` for any other local user for as long as the bridge runs. A user who typed `--password` themselves chose that; a launcher that generates one and passes it this way makes it the default for everyone who follows the quick start.

The fix is nearly free, because the bridge already accepts it another way — `config.js:40-57` reads `HARNESS_REMOTE_PASSWORD` and `HARNESS_REMOTE_USERNAME` (with `OMP_BRIDGE_*` as a fallback):

```js
const child = spawn(process.execPath, [cliPath, ...bridgeArgs], {
  stdio: "inherit",
  env: { ...process.env, HARNESS_REMOTE_USERNAME: username, HARNESS_REMOTE_PASSWORD: password }
})
```

with the two `push` calls for username and password dropped from `buildBridgeArgs`. Credentials the user passed explicitly can keep flowing through argv untouched, since that is their own choice — or move to the same path for consistency.

Worth a test that asserts the generated password does not appear in the spawned argv, since that is the property that would silently regress.

---

bridge/src/launcher.js
```diff
@@ -0,0 +1,180 @@
...
+  } else {
+    port = await findAvailablePort(requestedPort, host)
+  }
+
+  let username = optionValue(args, "--username")
+  let password = optionValue(args, "--password")
+  if (Boolean(username) !== Boolean(password)) {
+    throw new Error("--username and --password must be supplied together")
+  }
+  const loopback = new Set(["127.0.0.1", "::1", "localhost"]).has(host)
```
giuliastro: Binding to loopback skips credential generation entirely, so the quick start's `--host 127.0.0.1` path starts a bridge with no authentication at all.

Loopback is a weaker boundary than it looks in the two environments this product actually targets. On Android — the documented Termux/proot setup — the loopback interface is shared by every installed application, so any app on the phone can reach the bridge. On a shared or multi-user host, every local account can. And the agent behind that bridge runs with full user privileges and auto-approves its own tool calls, which the README says plainly about `--root` not being a sandbox.

Generating credentials unconditionally costs nothing here: they are printed either way, and the client has to be configured with them either way. The only thing loopback saves is the typing that has to happen regardless.

If the intent is to keep the loopback path frictionless for a local browser, that is worth stating as an explicit decision in the code rather than falling out of the host check, because the reader cannot currently tell whether the asymmetry is deliberate.

---

bridge/src/launcher.js
```diff
@@ -0,0 +1,180 @@
...
+}
+
+export function resolveBackend(args, detected = detectBackends()) {
+  const explicit = optionValue(args, "--backend")
+  if (explicit) return explicit
+  if (detected.length === 1) return detected[0]
+  if (detected.length > 1) {
+    throw new Error(`Multiple supported agents detected (${detected.join(", ")}). Re-run with --backend <${detected.join("|")}>.`)
+  }
+  throw new Error("No supported agent CLI was detected on PATH. Re-run with --backend <omp|pi|claude|codex> to select an ACP backend explicitly.")
```
giuliastro: This error is the first thing an OpenCode-only user sees, and following its advice is wrong for them: it tells them to pick one of four ACP backends they do not have, when the correct answer is that OpenCode does not go through the bridge at all.

`QUICK_START.md` says the right thing in its OpenCode section, but nobody reads the document after the command has already failed. It is also not a rare case — OpenCode is the one backend that works without any of this, and it is what the maintainer's own environment runs today.

Cheapest version: add `opencode` to the detection map, and when it is the only thing found, print the direct-HTTP path instead of throwing:

```text
OpenCode detected. OpenCode does not use the Harness bridge — start it directly:

  OPENCODE_SERVER_USERNAME=harness OPENCODE_SERVER_PASSWORD=… \
    opencode serve --hostname 0.0.0.0 --port 4096

Then select the OpenCode backend in the client.
```

That turns the worst first run in the matrix into a working one, which is the entire point of the issue this PR closes. At minimum, name the OpenCode path in the error string.

---

bridge/src/launcher.js
```diff
@@ -0,0 +1,180 @@
...
+    if (port > 65_535) break
+    if (await canListen(port, host)) return port
+  }
+  throw new Error(`No available port found from ${startPort} through ${Math.min(65_535, startPort + attempts - 1)}.`)
+}
+
+function lanAddress() {
+  for (const addresses of Object.values(networkInterfaces())) {
+    for (const address of addresses ?? []) {
+      if (address.family === "IPv4" && !address.internal) return address.address
```
giuliastro: The first non-internal IPv4 is frequently not the one the phone can reach. A typical development machine enumerates `docker0` at `172.17.0.1`, or a VPN interface, before the Wi-Fi address — and the printed `Connect to:` line is the single output this PR exists to produce, so getting it wrong sends the user to an address that silently never answers.

Two options, both small. Filter the obvious virtual interfaces by name (`docker`, `br-`, `veth`, `virbr`, `tun`, `tap`, `utun`) and prefer what remains. Or, more robustly, ask the routing table which source address would be used to reach the outside world — a UDP socket `connect()` to a public address needs no traffic and returns the right interface:

```js
const probe = net.createConnection // or dgram: socket.connect(53, "8.8.8.8"); socket.address().address
```

Printing every candidate is also acceptable and arguably better on a machine with several real interfaces, as long as the list is short and labelled. What is worth avoiding is confidently printing one address that is wrong, since the failure mode is indistinguishable from the server not working.

---

bridge/src/launcher.js
```diff
@@ -0,0 +1,180 @@
...
+import path from "node:path"
+import { spawn } from "node:child_process"
+import { fileURLToPath } from "node:url"
+
+const BACKEND_EXECUTABLES = {
+  omp: ["omp"],
+  pi: ["pi"],
+  claude: ["claude"],
+  codex: ["codex"]
+}
```
giuliastro, line 16: Detection here means "a file with this name exists in a PATH directory". That correctly satisfies the no-arbitrary-execution requirement from #143, and it is worth being precise about what it does not establish: not that the file is executable, and not that the agent is authenticated or configured.

The consequence is that a `claude` or `codex` present but not logged in is auto-selected when it is the only one found, and the failure surfaces later from inside the bridge handshake, where it reads as a bridge problem rather than a login problem.

Not a blocker for this PR, and #143 already anticipates the fix by asking for available / unauthenticated / unconfigured states. Two things would help meanwhile: check the executable bit rather than mere existence, and phrase the output as "found on PATH" rather than "detected", so the word does not promise a readiness the check has not performed.

---

giuliastro (commented):
This is the right shape for the milestone: 320 additions, no UI changes, no bridge rewrite, existing manual paths untouched, and a launcher that stays honest about being thin. The tests cover the decision logic rather than the plumbing, which is the useful half, and `QUICK_START.md` is unusually straight about what has not been done — the note on why the root package stays private is the kind of thing that normally gets discovered six months later.

Five findings inline, in the order I would act on them. None argues against the approach.

**Two I would fix before merging.** The generated password reaches the bridge on the child's argv, where any local user can read it out of `ps`; the bridge already accepts `HARNESS_REMOTE_PASSWORD` from the environment, so this is a small change rather than a design question. And the loopback path generates no credentials at all, which is a thinner boundary than it looks on a phone, where every installed app shares the loopback interface, and on any multi-user host.

**One that decides whether this issue succeeds at what it set out to do.** An OpenCode-only user — the configuration the maintainer runs today — gets an error telling them to choose an ACP backend they do not have. The quick-start document explains the OpenCode path correctly, but only to someone who reads it after the command has already failed. Detecting `opencode` and printing the direct-HTTP instructions turns the worst first run into a working one.

**Two smaller ones.** The printed LAN address takes the first non-internal IPv4, which on a machine with Docker or a VPN is often not the reachable one — and that single line is the entire output the user acts on. And detection establishes only that a file exists on PATH, not that the agent is authenticated, so "detected" claims slightly more than the check performs.

Two things I checked and found correct rather than assumed: the executable search never runs anything, and `buildBridgeArgs` genuinely leaves user-supplied options alone — the second test case covers exactly the override path that would be easy to break later.

A note on validation rather than on the code: the CI here proves the launcher's logic, not that the end-to-end path works, because no ACP-backed harness is reachable in the environment available. The first real test of this PR is someone running it on a machine with one of the four agents installed, and until that happens the risk sits in the part the tests cannot see — the handshake between a generated configuration and a real agent.

Issue #146, opened 2026-08-12
feat(fleet): manage agent work across multiple local machines


giuliastro:
Deferred until the P0 stabilization gate is complete.

Multi-machine fleet remains a strategic roadmap milestone, but it should not stay open as active work while one-machine daemon/client/task behavior is still being stabilized. Reopen when #143, #180 and #145 are genuinely reliable.

giuliastro:
## Validation gate before implementation

Keep multi-machine fleet management as the strongest currently identified wedge, but separate **scarcity evidence** from **demand evidence**.

Before #146 becomes the main build, validate the demand assumption cheaply: ask existing users/contributors whether they actually run coding agents on more than one machine (desktop + laptop, workstation + server, local + VM, etc.) and whether one control surface would change their workflow.

The milestone should proceed only with that evidence in view. A market gap can exist because nobody built the feature, or because few users want it; we need to distinguish those cases before making #146 the largest investment in the roadmap.

giuliastro:
Implementation update: stacked PR #168 adds the first Fleet client/control-plane foundation on top of #167. It groups per-agent profiles by physical daemon endpoint, discovers multiple machines independently, preserves stable daemon machine identity when reachable, keeps unreachable machines visible without breaking the fleet, aggregates agents/projects/tasks/runs, namespaces task identity by machine, and supports explicit selected-machine task placement through the normalized task APIs. Fleet UI remains intentionally out of scope for this slice.

PR #149, opened 2026-08-12
feat(daemon): introduce machine identity and agent host registry

giuliastro:
## Summary

This is the first implementation slice of #143.

- add a stable machine identity persisted under the Harness state directory
- preserve corrupt identity files for diagnosis and avoid concurrent first-start identity races
- introduce a backend-neutral `MachineRegistry` that can represent heterogeneous agent hosts on one machine
- register the current legacy ACP backend as the first machine host without changing existing session APIs
- track ACP host lifecycle in both directions (`configured` → `available` → `unavailable` → `available`)
- expose the fleet-safe machine contract through `GET /v1/machine` and `GET /global/machine`
- add unit coverage for stable identity, corruption recovery, concurrent identity creation, heterogeneous hosts, lifecycle recovery, the HTTP machine endpoint, and authentication

## Why

The Universal Daemon needs a machine-scoped contract before process ownership and multi-agent routing are layered on top. This slice deliberately establishes:

```text
Machine
  -> AgentHost[]
```

instead of extending the current implicit `server -> backend` model.

The registry already supports heterogeneous transports (for example ACP and direct HTTP), so OpenCode can become a real host in the next slice without changing the machine API shape.

## Host state contract

- `configured`: registered but not yet successfully started
- `available`: transport/agent completed startup successfully
- `unavailable`: a previously started or attempted host exited/failed

Recovery is observable: a later successful ACP restart returns the host to `available`.

## Compatibility

Existing single-backend bridge/session endpoints remain in place. A legacy bridge process simply appears as a machine with one registered agent host.

## Follow-ups in #143

- move from one legacy host per bridge process to multiple concurrently managed hosts in one daemon
- add OpenCode as a managed/direct-HTTP host rather than launcher guidance
- expose richer host health/process lifecycle through the registry
- route agent-scoped session/run operations through the machine daemon

### DIFF
--- a/bridge/src/cli.js
+++ b/bridge/src/cli.js
@@ -3,6 +3,7 @@ import path from "node:path"
 import { AcpClient } from "./acp-client.js"
 import { parseConfig, usage } from "./config.js"
 import { harnessProfile } from "./harness-profiles.js"
+import { loadMachineIdentity, MachineRegistry } from "./machine-registry.js"
 import { createBridgeServer } from "./server.js"
 
 let config
@@ -20,10 +21,22 @@ if (config?.help) {
 
 if (config) {
   const profile = harnessProfile(config.backend)
+  const machineIdentity = await loadMachineIdentity(config.stateDirectory)
+  const machineRegistry = new MachineRegistry(machineIdentity)
+  machineRegistry.registerHost({
+    id: profile.id,
+    label: profile.label,
+    backend: profile.id,
+    transport: "acp",
+    state: "configured",
+    capabilities: profile.capabilities
+  })
+
   const acp = new AcpClient({ command: config.acpCommand, args: config.acpArgs, permissionMode: profile.permissionMode, preferredAuthMethod: profile.authMethod })
   const server = createBridgeServer({
     config,
     acp,
+    machineRegistry,
     serviceOptions: {
       snapshotDirectory: path.join(config.stateDirectory, profile.id),
       historyLoader: profile.historyLoader,
@@ -41,11 +54,13 @@ if (config) {
     process.stderr.write(`[${config.backend}] handled agent request: ${message.method}\n`)
   })
   acp.on("exit", (error) => {
+    machineRegistry.updateHost(profile.id, { state: "unavailable" })
     if (!shuttingDown) process.stderr.write(`[${config.backend}] ${error.message}\n`)
   })
 
   server.listen(config.port, config.host, () => {
     process.stdout.write(`${config.backend.toUpperCase()} bridge listening on http://${config.host}:${config.port}\n`)
+    process.stdout.write(`Machine: ${machineIdentity.name} (${machineIdentity.id})\n`)
   })
 
   const shutdown = () => {
--- a/bridge/src/cli.js
+++ b/bridge/src/cli.js
@@ -3,7 +3,7 @@ import path from "node:path"
 import { AcpClient } from "./acp-client.js"
 import { parseConfig, usage } from "./config.js"
 import { harnessProfile } from "./harness-profiles.js"
-import { loadMachineIdentity, MachineRegistry } from "./machine-registry.js"
+import { loadMachineIdentity, MachineRegistry, trackAgentHostLifecycle } from "./machine-registry.js"
 import { createBridgeServer } from "./server.js"
 
 let config
@@ -32,7 +32,11 @@ if (config) {
     capabilities: profile.capabilities
   })
 
-  const acp = new AcpClient({ command: config.acpCommand, args: config.acpArgs, permissionMode: profile.permissionMode, preferredAuthMethod: profile.authMethod })
+  const acp = trackAgentHostLifecycle(
+    new AcpClient({ command: config.acpCommand, args: config.acpArgs, permissionMode: profile.permissionMode, preferredAuthMethod: profile.authMethod }),
+    machineRegistry,
+    profile.id
+  )
   const server = createBridgeServer({
     config,
     acp,
@@ -54,7 +58,6 @@ if (config) {
     process.stderr.write(`[${config.backend}] handled agent request: ${message.method}\n`)
   })
   acp.on("exit", (error) => {
-    machineRegistry.updateHost(profile.id, { state: "unavailable" })
     if (!shuttingDown) process.stderr.write(`[${config.backend}] ${error.message}\n`)
   })
 
--- a/bridge/src/machine-registry.js
+++ b/bridge/src/machine-registry.js
@@ -30,6 +30,9 @@ async function installIdentity(machineFile, identity) {
   const temporary = `${machineFile}.${process.pid}.${randomUUID()}.tmp`
   await writeFile(temporary, `${JSON.stringify(identity, null, 2)}\n`, { mode: 0o600 })
   try {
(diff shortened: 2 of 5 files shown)


bridge/src/server.js
```diff
@@ -122,15 +105,6 @@ function modelWireName(model) {
...
-/**
- * The app's model API is OpenCode's, which names a model `provider/model`. ACP has no such rule:
- * OMP and PI happen to use that shape, while Claude Code's adapter offers bare ids — `sonnet`,
- * `opus[1m]`. Splitting on "/" and requiring both halves silently dropped every one of them, which
- * is why that backend looked like it exposed no models at all.
- *
- * A bare id is presented under the backend's own name instead, so it reads and behaves like the
- * others — `claude/sonnet`. `AcpService.setModel` puts it back to the id the agent knows.
- */
 function providersResponse(models, fallbackProviderID) {
```
giuliastro, line 134: This PR deletes roughly nine explanatory comments from `server.js` that have nothing to do with machine identity, and the loss is not visible from the title or the summary. Each one recorded a failure that had already been paid for once.

The one that stood above this function is the clearest example. It explained that the app's model API is OpenCode's `provider/model` shape, that ACP has no such rule, that one adapter offers bare ids like `sonnet` and `opus[1m]`, and that splitting on `/` and requiring both halves silently dropped every one of them — "which is why that backend looked like it exposed no models at all". Nothing in the remaining code says why `fallbackProviderID` exists. The next person to tidy this function has no way to know that removing it reintroduces a bug where an entire backend appears to support no models.

The others in the same diff:

- why the Private Network Access header is set, and that it exists because a public PWA reaches a local bridge;
- why the preflight must be answered before authentication, since browsers omit credentials on `OPTIONS`;
- why base64 length is measured instead of decoded;
- why attachments are validated before the prompt reaches the agent;
- why attachment capability comes from the handshake instead of the profile table, and what a static `true` would break;
- why the SSE heartbeat is 10s specifically, and that a mismatch makes clients reconnect in a loop;
- why an image with no caption is a complete prompt;
- why a slash command is sent as prompt text.

Every one answers a "why is this like that?" that the code cannot answer on its own, which is the convention the rest of this repository follows deliberately.

I would restore all of them in this PR. If some are genuinely thought to be noise, that is a reasonable position, but it deserves its own change with that as its stated purpose — not a side effect of a PR about machine identity, where a reviewer looking at the registry has no reason to check.

The reformatting of the multi-line `if` blocks into one-liners in `parseAttachments` is in the same category: harmless on its own, unrelated here, and it enlarges the diff a reviewer must read to find the actual change.

---

bridge/src/cli.js
```diff
@@ -41,11 +54,13 @@ if (config) {
     process.stderr.write(`[${config.backend}] handled agent request: ${message.method}\n`)
   })
   acp.on("exit", (error) => {
+    machineRegistry.updateHost(profile.id, { state: "unavailable" })
```
giuliastro: This is the only place the host state ever changes after registration, and it only moves in one direction. Once the ACP process exits, the host is `unavailable` for the lifetime of the daemon — including after it comes back, since `acp.start()` is called lazily from the request handlers and never tells the registry.

So `/v1/machine` reports a permanently dead agent on a machine where the agent is working again, and the only way to correct it is restarting the bridge. That undercuts the acceptance criterion this slice exists to satisfy: individual host failures are indeed isolated and visible, but recoveries are neither.

The fix belongs wherever the ACP transitions are already observed — set `state: "available"` (or back to `configured`) once `start()` resolves, alongside the `unavailable` transition here. A test that exits and restarts the fake ACP and asserts the snapshot follows it both ways would pin the behaviour; the registry tests currently cover registration and shape, not lifecycle.

Worth deciding the vocabulary in this PR too, since clients will start depending on it: `configured`, `available` and `unavailable` are three states in the code, but only two of them are ever written here.

---

bridge/src/machine-registry.js
```diff
@@ -0,0 +1,83 @@
...
+  const identity = {
+    id: `machine_${(options.randomUUID ?? randomUUID)()}`,
+    name: (options.hostname ?? hostname)(),
+    createdAt: new Date().toISOString()
+  }
+  await mkdir(stateDirectory, { recursive: true })
+  const temporary = `${machineFile}.${process.pid}.tmp`
+  await writeFile(temporary, `${JSON.stringify(identity, null, 2)}\n`, { mode: 0o600 })
+  await rename(temporary, machineFile)
+  return identity
```
giuliastro: A truncated or hand-edited `machine.json` takes the same path as a missing one: the identity is regenerated and the old file is overwritten, with nothing written to stderr. The same happens when the file parses but fails `validIdentity`.

For a value whose entire purpose is stability that is the wrong default. #146 requires machine identity to survive ordinary restarts and to keep machines distinguishable in a fleet; a machine that silently renames itself after a bad shutdown appears in the client as a *new* machine, while the old one lingers as an unreachable ghost nobody can explain. The evidence needed to diagnose it — the previous file — is destroyed by the same operation that causes it.

Suggestion: keep the failure. Rename the unreadable file to `machine.json.corrupt` (or `.corrupt-<timestamp>`) before writing the replacement, and write one line to stderr saying a new identity was generated and where the old one went. Cheap, and it turns a silent identity change into something a user can report.

Second, smaller point on the same block: two daemons started against one state directory both generate an identity and both `rename`, so the file ends up with one and the two processes hold different ones in memory. The temp name includes the pid, so nothing is clobbered mid-write, but the outcome is still two machines claiming one home. Writing with the `wx` flag and re-reading on `EEXIST` closes it, and users will start a second daemon by accident well before the fleet work makes it intentional.

The atomic write itself — temp plus rename, mode `0o600` — is the right shape, and the defensive copies throughout the registry are careful work.

---

giuliastro (commented):
Good first slice of #143. The registry is a small, self-contained module with no dependency on the rest of the bridge, the identity write is atomic and mode-restricted, every accessor returns defensive copies rather than handing out internal state, and the endpoint is exposed under both the `/v1` and `/global` shapes the client already understands. Registering the single current backend as one host of a future many is exactly the right size for a first step — it changes nothing observable while making the shape real.

Two things I checked rather than assumed. `matchesCredentials` runs before the `try` block, so `/v1/machine` is authenticated — worth stating because it exposes hostname and agent inventory, and one of the new tests pins it. And the identity really does survive a restart, which the registry test covers directly.

Three findings inline, one of which I would treat as blocking.

**The deletion of around nine explanatory comments from `server.js` is unrelated to this PR and should be reverted here.** They document past failures — the model-naming one records a regression where an entire backend appeared to expose no models — and a PR titled "machine identity and agent host registry" is the last place a future reader would look for their removal.

**Host state only ever moves toward `unavailable`.** After an ACP crash and recovery the machine snapshot keeps reporting a dead agent until the bridge is restarted, which weakens the visibility criterion this slice is meant to establish.

**A corrupt identity file silently becomes a new machine**, overwriting the evidence. For a value whose only job is stability, that failure mode deserves a stderr line and a preserved copy.

One forward-looking note, not a change request: `/global/machine` is an alias the OpenCode server itself will never serve, since OpenCode does not go through the bridge. Whatever consumes this on the client should treat the endpoint as capability-gated rather than assume it exists everywhere, or the OpenCode path will 404 into a confusing empty state.

---

PR #150, opened 2026-08-12
feat(daemon): manage OpenCode as a supervised host

giuliastro:
## Summary

Second implementation slice of #143.

- add a reusable `ManagedOpenCodeHost` process supervisor
- start `opencode serve` directly instead of printing a second command
- keep OpenCode credentials out of argv and pass them only through `OPENCODE_SERVER_USERNAME` / `OPENCODE_SERVER_PASSWORD`
- report readiness only after an authenticated `/global/health` check succeeds with the generated credentials
- enforce an authoritative startup timeout even if readiness never settles
- supervise process exit and escalate repeated shutdown signals from `SIGTERM` to `SIGKILL`
- add a registry lifecycle adapter with consistent `configured` / `available` / `unavailable` transitions, including failed startup
- add tests for argv secrecy, authenticated readiness, readiness timeout, registry state transitions, process identity, and shutdown escalation
- refresh quick-start docs now that #148 is merged and OpenCode is actually started by Harness

## User-visible change

Before this PR:

```text
harness-remote --backend opencode
  -> prints another `opencode serve ...` command
```

With this PR:

```text
harness-remote --backend opencode
  -> starts OpenCode
  -> verifies authenticated health
  -> prints connection details
  -> supervises the process
```

The Harness Remote client still talks directly to OpenCode HTTP. This slice changes process ownership and lifecycle, not the client protocol.

## Why

The previous quick-start path exposed an important gap: OpenCode was recognized, but Harness did not actually manage it. #143 needs heterogeneous hosts to be real lifecycle-owned resources before a single daemon can route across them.

`ManagedOpenCodeHost` is deliberately reusable by the upcoming multi-host daemon rather than embedding process logic in the launcher.

## Review hardening

The review identified four lifecycle/readiness gaps, now covered by tests:

- no silent startup hang when a readiness operation never settles
- no false-positive readiness from an unrelated listener on the expected port
- failed starts become `unavailable` just like ACP hosts
- a second Ctrl-C escalates shutdown instead of being ignored

## Next slice in #143

- instantiate multiple managed hosts under one machine daemon
- register OpenCode and ACP-backed hosts together
- expose per-host health/lifecycle from one process
- then route agent-scoped operations through that daemon

### DIFF
--- a/bridge/src/launcher.js
+++ b/bridge/src/launcher.js
@@ -6,6 +6,7 @@ import { networkInterfaces } from "node:os"
 import path from "node:path"
 import { spawn } from "node:child_process"
 import { fileURLToPath } from "node:url"
+import { ManagedOpenCodeHost } from "./opencode-host.js"
 
 const BACKEND_EXECUTABLES = {
   omp: ["omp"],
@@ -15,7 +16,6 @@ const BACKEND_EXECUTABLES = {
   opencode: ["opencode"]
 }
 
-const LOOPBACK_HOSTS = new Set(["127.0.0.1", "::1", "localhost"])
 const VIRTUAL_INTERFACE = /^(docker|br-|veth|virbr|tun|tap|utun)/i
 
 function optionValue(args, name) {
@@ -139,11 +139,13 @@ export function lanAddresses(interfaces = networkInterfaces()) {
 }
 
 export function launcherUsage() {
-  return `Usage: harness-remote [options]\n\nQuick start options:\n  --backend <name>       Select omp, pi, claude, codex, or opencode (auto-detected when unambiguous)\n  --host <host>          Bind host (quick-start default: 0.0.0.0)\n  --port <port>          Preferred port (quick-start default: first free port from 4097)\n  --username <username>  Override generated Basic Auth username\n  --password <password>  Override generated Basic Auth password\n  --help                 Show this help\n\nAll other options are forwarded to harness-remote-bridge for ACP-backed agents.`
+  return `Usage: harness-remote [options]\n\nQuick start options:\n  --backend <name>       Select omp, pi, claude, codex, or opencode (auto-detected when unambiguous)\n  --host <host>          Bind host (quick-start default: 0.0.0.0)\n  --port <port>          Preferred port (OpenCode default: 4096; ACP default: 4097)\n  --username <username>  Override generated Basic Auth username\n  --password <password>  Override generated Basic Auth password\n  --help                 Show this help\n\nOpenCode is started and supervised directly. Other options are forwarded to harness-remote-bridge for ACP-backed agents.`
 }
 
-function openCodeGuidance({ host, port, username, password }) {
-  return `OpenCode was found on PATH. OpenCode connects directly to Harness Remote and does not use the ACP bridge.\n\nStart it directly:\n\n  OPENCODE_SERVER_USERNAME=${username} OPENCODE_SERVER_PASSWORD=${password} opencode serve --hostname ${host} --port ${port}\n\nThen select the OpenCode backend in Harness Remote and use the address/credentials above.`
+export async function startManagedOpenCode({ host, port, username, password, command = "opencode", Host = ManagedOpenCodeHost } = {}) {
+  const managed = new Host({ command, host, port, username, password })
+  await managed.start()
+  return managed
 }
 
 async function main() {
@@ -192,7 +194,24 @@ async function main() {
   process.stdout.write(`Password: ${password}\n`)
 
   if (backend === "opencode") {
-    process.stdout.write(`\n${openCodeGuidance({ host, port, username, password })}\n`)
+    process.stdout.write("\nStarting managed OpenCode host...\n")
+    const managed = await startManagedOpenCode({ host, port, username, password })
+    process.stdout.write(`OpenCode is ready on ${host}:${port}. Keep this process running while Harness Remote is connected.\n`)
+
+    let shuttingDown = false
+    const shutdown = () => {
+      if (shuttingDown) return
+      shuttingDown = true
+      managed.stop()
+    }
+    managed.on("unavailable", (error) => {
+      if (!shuttingDown) {
+        process.stderr.write(`${error.message}\n`)
+        process.exitCode = 1
+      }
+    })
+    process.on("SIGINT", shutdown)
+    process.on("SIGTERM", shutdown)
     return
   }
 
--- a/bridge/src/launcher.js
+++ b/bridge/src/launcher.js
@@ -148,6 +148,22 @@ export async function startManagedOpenCode({ host, port, username, password, com
   return managed
 }
 
+export function createManagedShutdown(managed, processObject = process) {
+  let signalCount = 0
+  let exitCode = 0
+  return (signal) => {
+    signalCount += 1
+    if (signalCount === 1) {
+      exitCode = signal === "SIGINT" ? 130 : 143
+      processObject.exitCode = exitCode
+      managed.stop("SIGTERM")
+      return
+    }
+    managed.stop("SIGKILL")
+    processObject.exit(exitCode || 1)
+  }
+}
+
 async function main() {
   const args = process.argv.slice(2)
   if (hasOption(args, "--help")) {
@@ -199,10 +215,10 @@ async function main() {
     process.stdout.write(`OpenCode is ready on ${host}:${port}. Keep this process running while Harness Remote is connected.\n`)
 
     let shuttingDown = false
-    const shutdown = () => {
-      if (shuttingDown) return
+    const shutdown = createManagedShutdown(managed)
+    const handleSignal = (signal) => {
       shuttingDown = true
-      managed.stop()
+      shutdown(signal)
     }
     managed.on("unavailable", (error) => {
       if (!shuttingDown) {
@@ -210,8 +226,8 @@ async function main() {
         process.exitCode = 1
       }
     })
-    process.on("SIGINT", shutdown)
-    process.on("SIGTERM", shutdown)
+    process.on("SIGINT", () => handleSignal("SIGINT"))
+    process.on("SIGTERM", () => handleSignal("SIGTERM"))
     return
   }
 
--- a/bridge/src/opencode-host.js
+++ b/bridge/src/opencode-host.js
@@ -1,0 +1,126 @@
+import { spawn } from "node:child_process"
+import { EventEmitter } from "node:events"
+import net from "node:net"
+
+const DEFAULT_START_TIMEOUT_MS = 15_000
+
+function waitForPort(host, port, timeoutMs = DEFAULT_START_TIMEOUT_MS) {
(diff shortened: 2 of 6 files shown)


bridge/src/opencode-host.js
```diff
@@ -0,0 +1,126 @@
...
+      })
+      socket.once("error", () => {
+        socket.destroy()
+        if (Date.now() >= deadline) {
+          reject(new Error(`OpenCode did not become ready on ${host}:${port} within ${timeoutMs}ms`))
+          return
+        }
+        setTimeout(tryConnect, 100).unref?.()
+      })
+    }
```
giuliastro: The deadline is only consulted inside the `error` handler, so this can wait forever. A TCP connect that neither succeeds nor errors — a firewall that drops rather than rejects, a half-open socket, a host that accepts the SYN and stalls — leaves both `connect` and `error` unfired, and `net.createConnection` has no timeout of its own.

`Promise.race` does not save it either: the only other branch is `exited`, which settles only if the child dies. A `serve` process that starts, binds nothing and stays alive produces a launcher hung on `Starting managed OpenCode host...` with no output and no exit. For a command whose entire promise is that it works in one step, a silent hang is the worst available failure.

Two ways to close it, both small. Give each attempt a bound:

```js
socket.setTimeout(1_000, () => { socket.destroy(); retryOrFail() })
```

or race the whole wait against a single overall timer, which also makes `startTimeoutMs` mean what its name says — right now it is a deadline that only advances when attempts fail fast.

Worth a test with a `waitUntilReady` double that never settles, asserting `start()` rejects within the timeout. The injectable seam is already there, so the test costs almost nothing.

---

bridge/src/opencode-host.js
```diff
@@ -0,0 +1,126 @@
...
+        `OpenCode exited before becoming ready (${code ?? "unknown"}${signal ? `, ${signal}` : ""})`
+      )))
+    })
+
+    try {
+      await Promise.race([
+        this.waitUntilReady(this.readinessHost, this.port, this.startTimeoutMs),
+        exited
+      ])
+      this.emit("available", { pid: this.processID, host: this.host, port: this.port })
```
giuliastro, line 138: Readiness here means "something accepted a TCP connection on that port". It does not establish that the thing listening is the process just spawned, nor that it is answering with the credentials this launcher generated.

That gap has already cost a day in this project. The failure mode was a server running on the expected port with different credentials, and the symptom was an authentication error that looked like a wrong password for hours. This code would report `OpenCode is ready` in exactly that situation.

The pre-flight in `main()` narrows it — `findAvailablePort` and the explicit `--port` branch both verify the port is free before spawning — so the window is a genuine race rather than the common case. But it is a race the product has already lost once, and the fix is one request:

```js
GET http://<readinessHost>:<port>/global/health
Authorization: Basic <generated credentials>
```

Require 200 before emitting `available`. A 401 then says something is listening that is not ours, which is precisely the diagnosis that was missing last time, and it upgrades "ready" from "a socket answered" to "the agent is up and will accept the credentials printed above".

It also makes the readiness signal reusable for the daemon in #143, where hosts need real health rather than port liveness.

---

bridge/src/opencode-host.js
```diff
@@ -0,0 +1,126 @@
...
+    this.child = undefined
+    this.emit("unavailable", error)
+  }
+}
+
+export function trackManagedHostLifecycle(host, registry, hostID) {
+  host.on("available", () => registry.updateHost(hostID, { state: "available", processID: host.processID }))
+  host.on("unavailable", () => registry.updateHost(hostID, { state: "unavailable", processID: undefined }))
+  return host
+}
```
giuliastro, line 180: This adapter never sees a failed start. When `#start()` throws, it calls `this.stop()` first, which clears `this.child` so `#handleExit` returns early, and the pre-success listeners are the ones that reject the `exited` promise rather than the ones that emit `unavailable`. So the only signal is the rejected promise, and a registry attached through this function keeps whatever state it had — `configured` — for a host that just failed to start.

That is the opposite of the sibling adapter merged in #149, where `trackAgentHostLifecycle` wraps `start()` and sets `unavailable` when it rejects. Two host types, two different answers to the same question, and the daemon in #143 is about to read both through one API.

Harmless right now, because this is only exercised by tests — `startManagedOpenCode` does not attach a registry, so nothing in the running product uses it yet. That is also the reason to fix it now rather than later: it is currently free, and the moment the multi-host daemon wires it up, a host that fails to start reports as healthy and the failure isolation criterion in #143 quietly stops holding.

Either emit `unavailable` before rethrowing in `#start()`, or wrap `start()` here the way the ACP adapter does. The second keeps both adapters shaped alike, which matters more than either shape individually.

---

bridge/src/launcher.js
```diff
@@ -192,7 +194,24 @@ async function main() {
...
+      managed.stop()
+    }
+    managed.on("unavailable", (error) => {
+      if (!shuttingDown) {
+        process.stderr.write(`${error.message}\n`)
+        process.exitCode = 1
+      }
+    })
+    process.on("SIGINT", shutdown)
+    process.on("SIGTERM", shutdown)
```
giuliastro: Registering a `SIGINT` listener replaces Node's default behaviour of terminating on Ctrl-C, and `shutdown` only sends a single `SIGTERM` to the child. The `shuttingDown` guard then makes every subsequent signal a no-op.

So if OpenCode ignores or is slow to handle `SIGTERM` — a child mid-write, or one that traps the signal — the user presses Ctrl-C, sees nothing happen, presses it again, and still nothing happens. The usual escape from a stuck foreground process has been taken away by the handler that was added to make shutdown clean.

Suggested shape: on the first signal, `stop()`; on a second, escalate to `SIGKILL` and exit. Something like keeping a counter instead of a boolean, and having `stop()` accept the signal to send. Worth also deciding explicitly whether the parent exits after a clean stop, rather than relying on the event loop draining once the child handle is released — that works today, but it is incidental rather than stated, and it will stop being true the moment the daemon holds a listening socket.

---

giuliastro (commented):
Good slice, and the right thing to build second: turning OpenCode from something the launcher describes into something it owns is what makes "heterogeneous hosts" more than a word in the daemon design.

Several things here are done carefully. Credentials go to the child only through the environment, with a test that asserts they never reach argv — that habit is now consistent across both host types. The `Promise.race` against an `exited` promise means a child that dies during startup fails immediately instead of waiting out the readiness timeout, which is the failure most implementations get wrong. `spawnProcess` and `waitUntilReady` are injectable, so the whole supervisor is testable without real processes or real ports. And mapping a `0.0.0.0` bind to a `127.0.0.1` readiness probe is a small detail that would have caused a confusing failure had it been missed.

**First, the CI failure is not yours.** The Debug APK job died in `npm ci` inside `web/`, on the deprecation output of the electron-builder dependency tree, after about fifty seconds. This PR changes no file under `web/` and does not touch a lockfile, and the sibling job installed the same dependency set successfully twenty seconds earlier from the same cache key. That points at a transient registry failure rather than anything in the diff. Re-run the job; if it fails a second time in the same place it is worth a proper look, but I would not read it as a signal about this change.

Four findings inline, in the order I would act on them.

**The readiness wait can hang indefinitely** — the timeout is only consulted when a connection attempt errors, so a connect that stalls leaves the launcher silent forever. On a one-command promise that is the worst shape of failure.

**Readiness proves only that a socket answered**, not that it is the process just spawned or that it accepts the generated credentials. This project has already lost a day to precisely that situation; one authenticated health request closes it and produces a much better diagnostic.

**A failed start never reports `unavailable`**, which contradicts the adapter merged in #149 and will silently break the failure-isolation criterion in #143 the moment the daemon wires this up. Free to fix now, awkward later.

**Ctrl-C twice does nothing**, because the handler replaces Node's default and guards against re-entry without escalating.

None of these argue against the design. The first two I would fix before merging; the third before the daemon starts consuming this adapter.

One note for the next slice rather than this one: `startManagedOpenCode` builds its own host and returns it, while the ACP path spawns a bridge child. When the daemon starts owning both, these two shapes need to meet somewhere — worth deciding whether the daemon supervises processes directly or keeps delegating to the launcher, before a third host type makes the answer expensive.

---

giuliastro (commented):
All four findings are addressed, and two of them are better than what I proposed.

**Readiness is now a real health check.** An authenticated `GET /global/health`, with a 401 raised immediately as `rejected the generated credentials` rather than retried until the deadline. That is exactly the diagnosis that was missing when this failure mode cost a day, and it now arrives in one sentence. `httpHost` bracketing IPv6 literals is a detail I did not ask for and would have missed.

**The hang is closed twice over** — a per-attempt `AbortController` and an overall `startTimeout` racing the whole start, so neither a stalled connect nor a process that lives without binding can wedge the launcher. `startTimeoutMs` now means what its name says.

**`trackManagedHostLifecycle` wraps `start()`**, matching `trackAgentHostLifecycle` from #149. Both host types now answer the failure question the same way, which is what the daemon needs more than either shape individually.

**`createManagedShutdown` escalates**: first signal terminates, second kills and exits, with 130 and 143 as the exit codes. Extracting it made it testable, which the new tests use.

Also worth noting: the `start()` guard moved from `!child.killed` to `exitCode == null && signalCode == null`, which is the correct question — `killed` only records that a signal was sent, not that the process died.

Two small things, neither blocking.

`waitForOpenCodeHealth` distinguishes the credential rejection by matching on the message text (`error?.message?.includes("rejected the generated credentials")`). That works, and it breaks silently the day someone rewords the string — a flag on the error, or a small subclass, survives editing. It is contained enough that I would fix it whenever this file is next open rather than now.

After a failed start, `stop()` no longer clears `this.child`, so `processID` keeps reporting the pid of a dead process until the next `start()`. Harmless today; worth knowing once the daemon reports pids as part of host state.

---

**On CI: my previous call was wrong, and the correction matters.**

I said the Debug APK failure looked transient and suggested a re-run. It has now failed three times in the same place, so that was incorrect. The real picture:

The failing step is **`npx @capacitor/assets generate`**, not `npm ci`. The job reaches it, installs `@capacitor/assets@3.0.5`, and exits 1 about a second later.

What it is not: `@capacitor/assets` has not published since March 2024, so this is not a new release. It is not this PR's code either — the diff touches no file under `web/` or `android/`. And it is not main: `android-apk.yml` ran the identical step against `1db3d61`, this PR's base, four minutes before the first failure here, and succeeded.

What I cannot tell from the log is why the tool exits, because it prints nothing to the captured output before the failure — the root cause needs the full step log from the Actions UI, or the step re-run with more verbosity.

Regardless of the cause, one change is worth making on its own merits: **icon generation should not be able to block a pull request's debug APK.** It regenerates launcher icons, which are irrelevant to a build whose only job is letting someone install a branch and try it, and it is the single unpinned network install in that job. Making it non-fatal in `pr-checks.yml`, or dropping it there and keeping it in the release workflow where icons actually matter, removes a class of failures that has nothing to do with the code under review. I introduced that step by copying it from the release workflow without asking whether a test build needs it; it does not.

---

PR #151, opened 2026-08-12
feat(daemon): run ACP and OpenCode under one machine daemon

giuliastro:
## Summary

Third implementation slice of #143.

- add a `MachineDaemon` runtime that owns one shared `MachineRegistry`
- register ACP and managed HTTP hosts under the same stable machine identity
- supervise OpenCode eagerly while keeping the primary ACP host lazy/compatible with the existing bridge behavior
- isolate host failures so one unavailable agent does not erase or stop the other
- pass the shared multi-host registry into the existing HTTP server, so `/v1/machine` and `/global/machine` expose both hosts
- add a real `harness-remote-daemon` entrypoint with one primary ACP host plus managed OpenCode
- expose the daemon command from both the bridge package and repository root
- start multiple eager managed hosts concurrently rather than serially
- start the daemon HTTP API before eager hosts finish readiness, so host state remains observable during startup failures/timeouts
- bind managed OpenCode to loopback by default, with explicit `--opencode-host` opt-in for wider exposure
- preflight the OpenCode port before spawning and fail with an actionable collision message
- add tests for concurrent heterogeneous hosts, startup failure isolation, bind safety, port preflight, shared server registry, shutdown, and daemon option parsing
- document the experimental multi-host path and its current routing limitation

## User-visible shape

```text
Harness daemon :4097
  ├── primary ACP host (Codex / Claude / OMP / PI)
  └── managed OpenCode 127.0.0.1:4096
```

A single machine query now sees both host records and their independent lifecycle states.

From a checkout:

```bash
npm run daemon -- --backend codex --host 127.0.0.1
```

Managed OpenCode stays on loopback even when the daemon binds to `0.0.0.0`. During this migration phase, wider OpenCode exposure is explicit:

```bash
harness-remote-daemon --backend codex --host 0.0.0.0 --opencode-host 0.0.0.0
```

## Review hardening

Claude identified four daemon-boundary issues, all addressed in this head:

- no implicit second LAN listener for OpenCode
- machine API listens before eager-host readiness completes
- eager managed hosts start concurrently
- occupied OpenCode ports are detected before process spawn

## Scope boundary

Session/run endpoints still use the selected primary ACP backend in this slice. OpenCode remains a direct HTTP protocol transport. The next #143 slice will introduce agent-scoped routing over the machine daemon instead of pretending that registry convergence alone is full transport convergence.

## #143 acceptance progress

This slice is intended to satisfy the first daemon acceptance milestone in executable form: one Harness daemon process can represent/manage at least two supported agent hosts concurrently, with isolated lifecycle state and one machine registry/API.

### DIFF
--- a/bridge/src/daemon-cli.js
+++ b/bridge/src/daemon-cli.js
@@ -1,0 +1,161 @@
+#!/usr/bin/env node
+import path from "node:path"
+import { AcpClient } from "./acp-client.js"
+import { parseConfig, usage as bridgeUsage } from "./config.js"
+import { harnessProfile } from "./harness-profiles.js"
+import { loadMachineIdentity } from "./machine-registry.js"
+import { MachineDaemon, createMachineDaemonServer } from "./machine-daemon.js"
+import { ManagedOpenCodeHost } from "./opencode-host.js"
+
+function requireValue(args, index, option) {
+  const value = args[index + 1]
+  if (!value || value.startsWith("--")) throw new Error(`${option} requires a value`)
+  return value
+}
+
+function parsePort(value, option) {
+  const port = Number(value)
+  if (!Number.isInteger(port) || port < 1 || port > 65_535) throw new Error(`${option} must be an integer between 1 and 65535`)
+  return port
+}
+
+export function parseDaemonOptions(args, environment = process.env) {
+  const bridgeArgs = []
+  const options = {
+    openCode: true,
+    openCodeCommand: environment.HARNESS_REMOTE_OPENCODE_COMMAND ?? "opencode",
+    openCodePort: parsePort(environment.HARNESS_REMOTE_OPENCODE_PORT ?? "4096", "--opencode-port")
+  }
+
+  for (let index = 0; index < args.length; index += 1) {
+    const option = args[index]
+    if (option === "--no-opencode") {
+      options.openCode = false
+      continue
+    }
+    if (option === "--opencode-command") {
+      options.openCodeCommand = requireValue(args, index, option)
+      index += 1
+      continue
+    }
+    if (option === "--opencode-port") {
+      options.openCodePort = parsePort(requireValue(args, index, option), option)
+      index += 1
+      continue
+    }
+    bridgeArgs.push(option)
+    if (["--backend", "--host", "--port", "--username", "--password", "--acp-command", "--acp-arg", "--root", "--cors", "--state-dir"].includes(option)) {
+      bridgeArgs.push(requireValue(args, index, option))
+      index += 1
+    }
+  }
+
+  return { config: parseConfig(bridgeArgs, environment), ...options }
+}
+
+export function daemonUsage() {
+  return `${bridgeUsage()}\n\nMulti-host daemon options:\n  --opencode-command <path>  OpenCode executable (default: opencode)\n  --opencode-port <port>     Managed OpenCode port (default: 4096)\n  --no-opencode              Start only the primary ACP host`
+}
+
+async function main() {
+  let parsed
+  try {
+    parsed = parseDaemonOptions(process.argv.slice(2))
+  } catch (error) {
+    process.stderr.write(`${error.message}\n\n${daemonUsage()}\n`)
+    process.exitCode = 1
+    return
+  }
+
+  const { config, openCode, openCodeCommand, openCodePort } = parsed
+  if (config.help) {
+    process.stdout.write(`${daemonUsage()}\n`)
+    return
+  }
+
+  if (openCode && openCodePort === config.port) {
+    throw new Error(`OpenCode port ${openCodePort} conflicts with the Harness daemon port`)
+  }
+
+  const identity = await loadMachineIdentity(config.stateDirectory)
+  const daemon = new MachineDaemon(identity)
+  const profile = harnessProfile(config.backend)
+  const acp = new AcpClient({
+    command: config.acpCommand,
+    args: config.acpArgs,
+    permissionMode: profile.permissionMode,
+    preferredAuthMethod: profile.authMethod
+  })
+
+  daemon.registerAcpHost({
+    id: profile.id,
+    label: profile.label,
+    backend: profile.id,
+    capabilities: profile.capabilities,
+    agent: acp
+  })
+
+  if (openCode) {
+    const managedOpenCode = new ManagedOpenCodeHost({
+      command: openCodeCommand,
+      host: config.host,
+      port: openCodePort,
+      username: config.username,
+      password: config.password
+    })
+    daemon.registerManagedHttpHost({
+      id: "opencode",
+      label: "OpenCode",
+      backend: "opencode",
+      capabilities: { sessions: true },
+      host: managedOpenCode
+    })
+  }
+
+  const server = createMachineDaemonServer({
+    daemon,
+    config,
+    primaryAcp: acp,
+    serviceOptions: {
+      snapshotDirectory: path.join(config.stateDirectory, profile.id),
+      historyLoader: profile.historyLoader,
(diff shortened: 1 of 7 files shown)


bridge/src/daemon-cli.js
```diff
@@ -0,0 +1,161 @@
...
+
+  if (openCode) {
+    const managedOpenCode = new ManagedOpenCodeHost({
+      command: openCodeCommand,
+      host: config.host,
+      port: openCodePort,
+      username: config.username,
+      password: config.password
+    })
+    daemon.registerManagedHttpHost({
```
giuliastro, line 119: Managed OpenCode inherits `config.host`, so starting the daemon with `--host 0.0.0.0` — which is what the quick-start path does by default — opens a **second** service on the LAN, on port 4096, that the user never asked for and that nothing tells them about. The startup banner prints `✓ OpenCode [http] available` with no address, so the only way to discover the extra listener is to look for it.

What makes it worth changing rather than documenting is the scope boundary this PR states plainly: session and run endpoints still go to the primary ACP host, and OpenCode remains a direct HTTP transport. So in this slice the daemon exposes a network service that nothing reaches *through the daemon*. It is exposure without a consumer, which is the least defensible kind.

Credentials do flow through, so it is not unauthenticated — the bridge's own rule forces a username and password beyond loopback and OpenCode receives them. The issue is reach, not auth.

Suggestion: default the managed host to `127.0.0.1` regardless of the daemon's bind address, and add `--opencode-host` for the case where someone deliberately wants it reachable. When routing lands in the next slice, the client will talk to the daemon and the managed host will not need to be externally reachable at all — so loopback is also the shape that survives.

If it stays inherited, the banner should at least print the address each host is listening on, so a user can see what was opened on their behalf.

---

bridge/src/daemon-cli.js
```diff
@@ -0,0 +1,161 @@
...
+  acp.on("stderr", (line) => process.stderr.write(`[${profile.id}] ${line}`))
+  acp.on("exit", (error) => process.stderr.write(`[${profile.id}] ${error.message}\n`))
+
+  const managedResults = await daemon.startManagedHosts()
+  for (const result of managedResults) {
+    if (result.status === "available") process.stdout.write(`[${result.id}] available\n`)
+    else process.stderr.write(`[${result.id}] unavailable: ${result.error?.message ?? "startup failed"}\n`)
+  }
+
+  server.listen(config.port, config.host, () => {
```
giuliastro: The server starts listening only after every eager host has finished starting, which inverts an availability relationship: `/v1/machine` is the endpoint that reports which hosts are unavailable, and it is unreachable for exactly as long as a host is failing to become available.

Concretely, an OpenCode that never becomes healthy holds the whole daemon dark for the full 15-second readiness timeout, and a client polling the machine endpoint during that window gets connection refused rather than `state: "configured"` followed by `unavailable`. The old bridge did not have this property — it listened immediately and let the ACP host start lazily.

The failure isolation this slice implements is real and tested, but a client can only observe it after the slowest host has resolved.

Suggestion: listen first, then start the eager hosts, letting the registry transitions do the reporting they were built for. The `managedResults` loop can keep printing its lines afterwards; nothing about it needs to precede `listen`. That also composes with the point on `startManagedHosts` — with several hosts, serial startup before listening means the daemon's own API waits on the sum of every host's timeout.

---

bridge/src/machine-daemon.js
```diff
@@ -0,0 +1,80 @@
...
+      try {
+        await entry.host.start()
+        results.push({ id: entry.id, status: "available" })
+      } catch (error) {
+        results.push({ id: entry.id, status: "unavailable", error })
+      }
+    }
+    return results
+  }
+
```
giuliastro, line 48: Eager hosts start one after another, so daemon startup costs the sum of their readiness timeouts rather than the slowest one. With today's single managed host that is invisible; the whole point of #143 is that it will not stay single, and five hosts at the 15-second default is a worst case of a minute and a quarter.

The result shape already anticipates the fix — every entry carries its own `status` and `error`, which is exactly what `Promise.allSettled` gives back:

```js
const eager = [...this.hosts.values()].filter((entry) => entry.eager)
const settled = await Promise.allSettled(eager.map((entry) => entry.host.start()))
return eager.map((entry, index) => settled[index].status === "fulfilled"
  ? { id: entry.id, status: "available" }
  : { id: entry.id, status: "unavailable", error: settled[index].reason })
```

Same isolation guarantee, and the existing tests should pass unchanged since they assert on the returned records rather than on ordering. Worth doing while there is one host and the change is trivial to verify, rather than when a slow host is holding four others hostage.

---

bridge/src/daemon-cli.js
```diff
@@ -0,0 +1,161 @@
...
+    return
+  }
+
+  const { config, openCode, openCodeCommand, openCodePort } = parsed
+  if (config.help) {
+    process.stdout.write(`${daemonUsage()}\n`)
+    return
+  }
+
+  if (openCode && openCodePort === config.port) {
```
giuliastro, line 88: This catches the two ports being equal, but not the far likelier collision: something else already listening on 4096. Anyone who followed the current quick-start guide has an `opencode serve` running on exactly that port, and this project's own maintainer had one running for most of today.

What happens then is worse than a clear error. The spawned OpenCode fails to bind and exits, while the health check — which now runs against `host:4096` — reaches the *other* server. If that one carries different credentials it answers 401 and the daemon reports `rejected the generated credentials`, which sends the reader to look at credentials that were never the problem. If it happens to carry the same ones, the daemon reports the host available and supervises a process that is already dead.

The launcher solved this in #148 and the helper is right there: `canListen` before spawning, with a message that names the likely cause.

```js
if (openCode && !(await canListen(openCodePort, config.host))) {
  throw new Error(`Port ${openCodePort} is already in use — is OpenCode already running? Use --opencode-port to choose another.`)
}
```

Two behaviours for the same situation across two entrypoints is the kind of divergence that gets discovered by a user rather than by a test.

---

giuliastro (commented):
This is the slice where #143 stops being a design and starts being a program, and the shape holds up. `MachineDaemon` owns one registry and knows nothing about how either host is implemented; the two adapters from #149 and #150 plug in unchanged; `createMachineDaemonServer` reuses the existing HTTP server rather than forking it. Registering the ACP host as lazy and the managed host as eager is the right asymmetry, and it is expressed as data on the entry rather than as a branch.

The tests are the strongest part. Concurrent heterogeneous hosts, one host failing without erasing the other, failed eager startup visible in the snapshot, shared registry reaching the server, and shutdown covering both host kinds — that is the acceptance criterion from #143 written as executable assertions rather than claimed in prose. The scope boundary in the description is also worth the credit: saying outright that registry convergence is not transport convergence, and that routing comes next, is what keeps this from reading as more than it is.

I checked one thing I expected to be a bug and it is not: `close?.()` on the ACP entry does reach `AcpClient.close()`, so shutdown is not silently skipping the ACP child. The optional chaining hides nothing here.

Four findings inline, none of them architectural.

Two concern the fact that this daemon now starts a **second network service**, and its defaults come from the wrong place. Managed OpenCode inherits the daemon's bind host, so a `0.0.0.0` daemon quietly opens 4096 to the LAN for a host that nothing in this slice routes to. And the port conflict check only compares the two configured ports instead of asking whether 4096 is actually free — which it very often will not be, since the documented setup until yesterday told people to run `opencode serve` on it themselves.

The other two are about startup shape. Eager hosts start serially, so the daemon will eventually pay the sum of every host's readiness timeout instead of the slowest. And the HTTP server listens only after that finishes, which means the machine endpoint — the one that exists to report which hosts are unavailable — is unreachable during precisely the window when a host is failing to start.

None of these blocks the slice from being correct today with one managed host. All four get more expensive the moment there are three.

PR #152, opened 2026-08-13
feat(daemon): route agent-scoped requests through one machine connection

giuliastro:
## Summary

Fourth implementation slice of #143.

- add `/v1/agents/:agentId/...` routing on the machine daemon connection
- route the selected primary ACP agent back through the existing normalized bridge API
- proxy managed HTTP agents such as OpenCode through the daemon instead of requiring clients to know the internal loopback port
- replace client-facing Basic Auth with the managed host's internal credentials before proxying
- keep CORS enforcement at the daemon boundary and do not forward browser Origin / CORS negotiation headers upstream
- preserve streaming responses, including OpenCode SSE, by piping the upstream HTTP response rather than buffering it
- return explicit 404 / 503 responses for unknown or unavailable agents
- keep all legacy unprefixed routes working during migration
- add end-to-end local HTTP tests for ACP path rewriting, OpenCode proxying, auth replacement, and availability isolation

## API shape

```text
GET /v1/machine

/v1/agents/codex/session
/v1/agents/codex/global/event
/v1/agents/opencode/session
/v1/agents/opencode/global/event
```

The daemon remains the only machine-level address the client needs. Managed OpenCode can stay on `127.0.0.1:4096`; external clients reach it through the daemon route.

## Security boundary

The daemon authenticates external agent-scoped HTTP routes using the daemon credentials. When proxying to a managed OpenCode process it strips the external `Authorization` header and injects the managed host credentials instead. It also owns CORS policy rather than delegating browser-origin trust to the internal process.

## Scope boundary

This PR establishes the transport/routing contract. It does not yet migrate the Android/web UI to choose an agent from `/v1/machine`; that client change is the next slice after this API is reviewed and stable.

### DIFF
--- a/bridge/src/agent-router.js
+++ b/bridge/src/agent-router.js
@@ -1,0 +1,156 @@
+import http from "node:http"
+import { timingSafeEqual } from "node:crypto"
+
+const AGENT_ROUTE = /^\/v1\/agents\/([^/]+)(\/.*)?$/
+const HOP_BY_HOP = new Set([
+  "connection",
+  "keep-alive",
+  "proxy-authenticate",
+  "proxy-authorization",
+  "te",
+  "trailers",
+  "transfer-encoding",
+  "upgrade"
+])
+
+function allowedOrigin(request, config) {
+  const origin = request.headers.origin
+  if (!origin || !config.corsOrigins?.length) return undefined
+  return config.corsOrigins.includes(origin) ? origin : undefined
+}
+
+function applyCorsHeaders(request, response, config) {
+  if (!config.corsOrigins?.length) return
+  response.setHeader("Vary", "Origin")
+  const origin = allowedOrigin(request, config)
+  if (!origin) return
+  response.setHeader("Access-Control-Allow-Origin", origin)
+  response.setHeader("Access-Control-Allow-Credentials", "true")
+  response.setHeader("Access-Control-Allow-Headers", "authorization, content-type")
+  response.setHeader("Access-Control-Allow-Methods", "GET, POST, PATCH, DELETE, OPTIONS")
+  if (request.headers["access-control-request-private-network"] === "true") {
+    response.setHeader("Access-Control-Allow-Private-Network", "true")
+  }
+}
+
+function matchesCredentials(request, config) {
+  if (!config.username) return true
+  const header = request.headers.authorization
+  if (!header?.startsWith("Basic ")) return false
+  const expected = Buffer.from(`${config.username}:${config.password}`)
+  const received = Buffer.from(header.slice("Basic ".length), "base64")
+  return received.length === expected.length && timingSafeEqual(received, expected)
+}
+
+function writeJSON(response, status, body) {
+  response.writeHead(status, { "Content-Type": "application/json; charset=utf-8", "Cache-Control": "no-store" })
+  response.end(JSON.stringify(body))
+}
+
+function proxyHeaders(headers, authorization) {
+  const result = {}
+  for (const [name, value] of Object.entries(headers)) {
+    if (value === undefined || HOP_BY_HOP.has(name.toLowerCase()) || name.toLowerCase() === "host" || name.toLowerCase() === "authorization") continue
+    result[name] = value
+  }
+  if (authorization) result.Authorization = authorization
+  return result
+}
+
+function forwardResponseHeaders(upstream, response) {
+  for (const [name, value] of Object.entries(upstream.headers)) {
+    if (value === undefined || HOP_BY_HOP.has(name.toLowerCase())) continue
+    response.setHeader(name, value)
+  }
+}
+
+function internalAuthorization(host) {
+  if (!host.username && !host.password) return undefined
+  return `Basic ${Buffer.from(`${host.username ?? ""}:${host.password ?? ""}`).toString("base64")}`
+}
+
+export function agentScopedRequest(request) {
+  const url = new URL(request.url ?? "/", `http://${request.headers.host ?? "localhost"}`)
+  const match = AGENT_ROUTE.exec(url.pathname)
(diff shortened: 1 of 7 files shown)


bridge/src/agent-router.js
```diff
@@ -0,0 +1,165 @@
...
+      path: `${route.path}${route.search}`,
+      headers: proxyHeaders(request.headers, internalAuthorization(host))
+    }, (upstreamResponse) => {
+      forwardResponseHeaders(upstreamResponse, response)
+      response.writeHead(upstreamResponse.statusCode ?? 502)
+      upstreamResponse.pipe(response)
+      upstreamResponse.once("end", resolve)
+    })
+    upstream.once("error", reject)
+    request.pipe(upstream)
```
giuliastro: Nothing tears the upstream request down when the client goes away, and the primary consumer of this proxy is a phone.

`/global/event` is an SSE stream that never ends on purpose. When the client disappears — screen off, network switch, app killed, tunnel dropped — `response` closes but `upstream` keeps its connection to OpenCode open and keeps receiving events with nowhere to write them. There is no `end`, so the promise never settles either. Every dropped mobile connection leaves one upstream connection behind, permanently, and on a phone dropped connections are the normal case rather than the exception.

```js
request.once("close", () => upstream.destroy())
```

is most of the fix. Worth also destroying `upstreamResponse` on the same signal, since `pipe` will not do it for you once the destination is gone.

Two related gaps in the same block:

`upstream.once("error", reject)` covers the outgoing request, but `upstreamResponse` has no `error` listener and `pipe()` does not forward errors. An upstream that resets mid-stream therefore emits `error` on an emitter nobody is listening to, which in Node is an uncaught exception — the daemon dies. That is the opposite of the host isolation the previous slices established: a managed agent crashing at the wrong moment takes down the process that is supposed to report it as unavailable.

And `requestImpl` is called with no timeout, so a managed host that accepts the connection and then never answers holds the client request open indefinitely. A socket timeout on non-streaming routes would bound it; streaming routes need the disconnect handling above rather than a timeout.

---

bridge/src/agent-router.js
```diff
@@ -0,0 +1,165 @@
...
+  const header = request.headers.authorization
+  if (!header?.startsWith("Basic ")) return false
+  const expected = Buffer.from(`${config.username}:${config.password}`)
+  const received = Buffer.from(header.slice("Basic ".length), "base64")
+  return received.length === expected.length && timingSafeEqual(received, expected)
+}
+
+function writeJSON(response, status, body) {
+  response.writeHead(status, { "Content-Type": "application/json; charset=utf-8", "Cache-Control": "no-store" })
+  response.end(JSON.stringify(body))
```
giuliastro: `matchesCredentials`, `applyCorsHeaders`, `allowedOrigin` and `writeJSON` are copies of the versions in `server.js` — I diffed the first one and it is byte-identical.

That means the daemon now enforces authentication and CORS in two files, from two copies of the same code, on two paths a request can take. Duplication is usually a style question; here it is not. The next fix to either — a tightened origin check, a change to what `Access-Control-Allow-Headers` permits, handling for a malformed `Authorization` header — lands in one file and silently leaves the other behind, and the resulting hole is invisible in review because both files still look correct on their own.

The Private Network Access header is a good illustration of the cost: it is set in both places, and the comment explaining why it exists lives in `server.js` only. A reader of this file sees an unexplained header; a maintainer editing this file has no reason to know the other copy exists.

Extracting them into something like `bridge/src/http-policy.js` and importing from both is a small change while there are exactly two callers. It also gives the tests one place to target rather than asserting the same properties twice.

---

giuliastro (commented):
This is the slice that makes the daemon worth having: one address, one credential, agents behind it. The security boundary is drawn in the right place and drawn deliberately — external `Authorization` stripped and replaced with the managed host's own, `Origin` and the CORS negotiation headers not forwarded upstream, `access-control-*` stripped off the way back so the daemon owns browser trust rather than delegating it to a process that never agreed to be a public server. Hop-by-hop headers are handled properly, which is the detail most hand-written proxies miss.

Piping the upstream response instead of buffering keeps SSE working, and returning 404 / 409 / 503 for unknown, non-routable and unavailable agents means a client can tell those three apart without guessing.

Three things I checked rather than assumed, all of which came out correct:

**No path traversal.** `agentScopedRequest` builds a `URL` first, so `/v1/agents/opencode/../../secret` is normalized to `/secret` before the route regex runs and simply does not match the agent route. The forwarded `route.path` cannot escape upward.

**No auth bypass on the delegated paths.** Both branches that hand off to `bridgeServer` — legacy unprefixed routes and the primary agent prefix — land in a handler that runs `matchesCredentials` before anything else, so the two paths that skip this file's own check are still authenticated.

**The agent id never reaches the upstream URL.** It is decoded and used only as a registry key; the upstream path comes from the regex tail.

Two findings inline.

The one I would fix before merging is that **a client disconnect does not tear down the upstream request**. On `/global/event` that leaks one upstream connection per dropped mobile session, forever, and the phone is the main client. In the same block, `upstreamResponse` has no `error` listener, so an upstream that resets mid-stream can crash the daemon — which is precisely the isolation the earlier slices were built to guarantee.

The second is that **`matchesCredentials`, `applyCorsHeaders`, `allowedOrigin` and `writeJSON` are byte-identical copies of the versions in `server.js`**. Authentication and CORS now exist twice, and the next change to either will land in one copy.

One note for the next slice rather than this one. The description says the client migration comes after this API stabilizes, which is the right order — but the client will need to know that `/v1/agents/<id>` exists at all, and today only `/v1/machine` reveals the ids. Worth deciding now whether the machine snapshot should advertise each host's route explicitly, so the client composes URLs from data rather than from a convention it has to hard-code.

---

PR #153, opened 2026-08-13
feat(client): route one machine connection to selected agents

giuliastro:
## Summary

Fifth implementation slice of #143.

- extend saved server profiles with an optional `agentId`
- keep legacy profiles byte-for-byte compatible when no agent is selected
- add machine-scoped discovery through `GET /v1/machine`
- route browser/Android requests below `/v1/agents/:agentId` by making the selected agent part of the request base
- preserve the same agent scope for global/project SSE URLs
- carry `agentId` through Electron profile IPC and route desktop requests below the same daemon path
- keep `/v1/machine` machine-scoped on desktop rather than accidentally nesting it below the selected agent
- persist and sanitize the selected agent in saved profiles
- add regression coverage for routing, URL encoding, SSE path preservation, legacy compatibility, and profile persistence

## Client contract

Legacy profile:

```text
http://workstation:4097/session
```

Daemon-backed profile selecting OpenCode:

```text
Machine: http://workstation:4097
Agent:   opencode
Request: http://workstation:4097/v1/agents/opencode/session
Events:  http://workstation:4097/v1/agents/opencode/global/event
```

Machine discovery always remains:

```text
GET http://workstation:4097/v1/machine
```

## Scope boundary

This PR establishes and validates the client-side transport/profile contract. It intentionally does not yet add the agent picker to the connection wizard; the next small UI slice can consume `discoverMachine()` and persist the chosen `agentId` without changing transport again.


web/electron/request-transport.ts
```diff
@@ -71,16 +71,21 @@ function validPath(path: unknown): path is string {
...
   } catch {
     return null
   }
+  const machineScoped = path === "/v1/machine" || path === "/global/machine"
+  const prefix = machineScoped || approved.pathname === "/" ? "" : approved.pathname.replace(/\/$/, "")
+  const target = new URL(approved)
+  target.pathname = `${prefix}${path}`
+  target.search = ""
+  target.hash = ""
   return target.origin === approved.origin ? target : null
```
giuliastro, line 89: This breaks every desktop request that carries a query string, which is most of the ones that matter.

Assigning a string containing `?` to `url.pathname` percent-encodes it rather than splitting it, and `target.search = ""` then discards what would have been the query. Reproduced:

```js
const approved = new URL("http://ws:4097/v1/agents/opencode")
const path = "/session?directory=/work/repo"
const target = new URL(approved)
target.pathname = `${approved.pathname}${path}`
target.search = ""
// http://ws:4097/v1/agents/opencode/session%3Fdirectory=/work/repo
```

`api.ts` builds these paths through `withDirectory`, so `/session?directory=…`, `/session/status?directory=…`, `/file?path=…&directory=…`, `/path`, `/project/current` and `/file/status` all arrive here with a query attached, and `desktopRequest` passes the path through unchanged. On desktop they now request a path whose last segment literally contains `%3F`, and the directory scoping is silently gone.

The change itself was necessary — the previous `new URL(path, baseUrl(profile))` resolves an absolute path against the origin and drops the `/v1/agents/<id>` prefix, which is exactly what this slice needed to fix. The fix just needs to keep the two halves apart:

```ts
const parsed = new URL(path, approved.origin)
const target = new URL(approved)
target.pathname = `${prefix}${parsed.pathname}`
target.search = parsed.search
target.hash = ""
```

CI would not catch this: the desktop surface has only the macOS application-menu smoke check, and no test exercises `targetURL` with a query. One directly against this function — a path with `?directory=` in and the query intact out — is worth adding alongside the fix, since this is now the third distinct behaviour that function has to get right.

---

web/src/serverConfig.ts
```diff
@@ -13,11 +13,22 @@ export function baseUrl(config: ServerConfig): string {
...
+  return agentID ? `${machine}/v1/agents/${encodeURIComponent(agentID)}` : machine
+}
+
+/** Useful when a caller already has a path and does not build through baseUrl. */
+export function agentScopedPath(config: ServerConfig, path: string): string {
+  const agentID = config.agentId?.trim()
+  if (!agentID) return path
+  const normalized = path.startsWith("/") ? path : `/${path}`
+  return `/v1/agents/${encodeURIComponent(agentID)}${normalized}`
+}
```
giuliastro, line 31: The new doc block sits where the old one did, so the diff reads as a replacement — but the text it replaced documented a different function, and that documentation is now gone from the file.

What was removed:

> A host typed one character at a time passes through states such as `http:` and `http://` that produce an unparseable base URL. Callers must check this before building any URL, because a throw on the render path blanks the whole app and a persisted invalid host reproduces that crash on every launch.

That belonged to `isValidServerConfig`, which now carries no explanation at all. It records an incident that is still live in the test file below — `server-config.test.mjs` walks `http:`, `http://`, `https:`, `https://` and asserts each is rejected, with its own comment describing the blank screen that persisted until app data was cleared. The code and the tests now disagree about how much this matters: the tests treat it as a scar, the source treats it as an ordinary validator.

Restoring it above `isValidServerConfig` costs nothing and keeps the two in step. Same category as the comment removals in #149 — not a style preference, but the difference between a future reader knowing why the guard exists and having to rediscover it.

Separately, on `agentScopedPath` just below: it is exported and covered by tests, but nothing in the app calls it. The only place that needed exactly this logic — `targetURL` in the Electron transport — re-implemented it inline, and that re-implementation is where the query-string bug in my other comment lives. That is the argument for wiring it up rather than deleting it: one implementation, one place to get right, one place to test.

---

web/src/machineClient.ts
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  let response: Response
+  try {
+    response = await fetch(target, { headers: headers(config) })
+  } catch {
+    throw new Error(`Cannot reach ${config.host}:${config.port}.`)
+  }
+  if (response.status === 404) return null
+  if (!response.ok) throw new Error(`HTTP ${response.status}`)
```
giuliastro: Discovery treats 404 as "this is a legacy server" and everything else as a failure, but the daemon has a second way of saying it has no machine to report. From #149, `server.js` answers `/v1/machine` with **503** and `{ error: "Machine registry is not configured" }` when no registry was passed in. A bridge built from that code path — and anything embedding `createBridgeServer` without a registry — returns 503, and `discoverMachine` throws `HTTP 503` instead of falling back to the legacy profile it was designed to protect.

Adding 503 to the null branch keeps the contract honest: both statuses mean "no machine here", and the caller's job is the same in either case.

The desktop branch has a sharper version of the same problem. It decides between legacy and error by pattern-matching the error text:

```ts
if (error instanceof Error && /404|not found/i.test(error.message)) return null
```

Any desktop transport error whose message happens to contain those words is silently reclassified as "legacy server", and any rewording of the message on the Electron side silently turns legacy servers into hard failures. Neither direction announces itself. Since `desktopRequest` already returns a structured `{ ok: false, error: { code, message } }`, the status or code is available to check directly — matching on prose is the one part of this file that cannot be made reliable by testing it.

---

giuliastro (commented):
**I would not merge this as it stands** — it breaks desktop requests that carry a query string, which is most of the ones that do useful work. Submitting as a comment only because GitHub will not let a review request changes on its author's own pull request; treat the first finding as blocking.

Everything else here is sound, and the split that causes the bug is the right split.

Separating `machineBaseUrl` from an agent-scoped `baseUrl` is the correct shape, and pointing `isValidServerConfig` at the machine URL rather than the agent URL is exactly right — an agent id can never be the reason a configuration is unusable, and validating the composed URL would have made a stale agent selection look like a broken host. Legacy profiles staying byte-for-byte identical, `/v1/machine` deliberately kept machine-scoped on desktop instead of nesting below the selected agent, and `agentId` trimmed and dropped when empty during profile parsing are all careful. `discoverMachine` returning null instead of throwing against a legacy server is the right default for a client that has to keep working across three generations of server at once.

**The blocker.** `targetURL` now assigns a path containing `?` to `url.pathname`, which percent-encodes it, and then clears `search`. `/session?directory=/work/repo` becomes `…/session%3Fdirectory=/work/repo` with no query at all — I reproduced it, and the repro is in the inline comment. Every desktop call built through `withDirectory` loses its directory scoping: session listing, statuses, the file browser, `project/current`. The previous code preserved the query and dropped the agent prefix instead, which is the thing this slice set out to fix, so the change was necessary and only the implementation is wrong. CI cannot see it — the desktop surface has one macOS menu smoke check, and nothing exercises this function.

Two smaller things inline. The comment recording why `isValidServerConfig` exists — the half-typed host that blanked the app on every launch — was removed and replaced with documentation for a different function, while the tests encoding that same regression still sit a few lines below. And `discoverMachine` treats only 404 as "no daemon", although `server.js` answers 503 when it has no registry, with the desktop branch deciding the same question by regex-matching an error message.

One structural note that ties two of these together: `agentScopedPath` is exported and tested but has no caller, while the one place that needed precisely that logic re-implemented it inline — and that re-implementation is where the blocking bug lives. Wiring the helper into the Electron transport collapses both problems into a single tested path.

---

PR #123, opened 2026-08-10
Add initial Prime Agent ACP support

giuliastro:
## What changed

Adds an initial Prime Agent backend to Harness Remote using Prime Agent's native ACP mode (`prime-agent --mode acp`). The bridge uses a small Prime-specific ACP client shim because Prime Agent intentionally does not expose `session/list` in ACP.

The web and desktop apps now recognize Prime Agent as a backend, create/store Prime server profiles, expose it in setup, and declare the capabilities supported by this first integration.

## Why

Prime Agent already speaks ACP natively, so Harness Remote can support prompt streaming, tool-call updates and cancellation without a terminal parser or an external ACP adapter.

## Current MVP limitation

Prime Agent ACP currently supports one session per connection and omits `session/list`/`session/load`. This PR therefore targets sessions created by the currently running bridge process. Persisted Prime Agent session browsing/resume and multiple concurrently addressable sessions are intentionally deferred to a follow-up integration.

## Validation

- Added bridge tests for the Prime Agent profile and the Prime-specific empty session-list behavior.
- Reviewed the backend registries used by web storage, desktop profile validation, setup and capability dispatch.
- GitHub currently exposes no CI/status checks for this branch, so no hosted build result is available from the repository.

PR #156, opened 2026-08-13
feat(client): select discovered agents from one machine connection

giuliastro:
## Summary

Sixth implementation slice of #143.

- keep the existing connection wizard fully compatible with legacy per-harness servers
- probe `GET /v1/machine` first once host and credentials are available, so users do not need to guess the daemon's primary backend
- when a machine daemon is discovered, show its agent hosts in the wizard
- prefer the discovered agent matching the harness the user initially chose, otherwise select the first available host
- prevent unavailable hosts from being selected
- persist the chosen `agentId` in the saved server profile
- derive the saved backend from the selected machine host so the existing client capability layer remains correct
- fall back to the existing backend-specific health test when `/v1/machine` is absent
- reuse the agent-scoped transport introduced by #153 without another App-level routing refactor
- add regression coverage for machine discovery and agent persistence

## User-visible flow

Legacy server:

```text
Choose harness -> host/port -> credentials -> test -> save
```

Machine daemon:

```text
Choose any starting harness -> machine host/port -> credentials -> test
                                                  -> discover /v1/machine
                                                  -> choose Codex / Claude / OpenCode / ...
                                                  -> save one machine-scoped profile
```

The starting harness is only a setup hint for daemon connections; discovery identifies the actual available hosts and the saved backend follows the selected host.

The profile then routes through `/v1/agents/:agentId/...` using the transport contract already merged in #153.

## Compatibility

If `/v1/machine` is absent or the server is a legacy bridge/OpenCode endpoint, discovery returns `null` and the wizard behaves as before.

### DIFF
--- a/web/src/components/panels.tsx
+++ b/web/src/components/panels.tsx
@@ -288,8 +288,12 @@ export function ConnectServerWizard({
     try {
       const result = await onTest(discoveryConfig)
       setTestResult(result)
-      if (!result.ok || !onDiscover) return
-      const discovered = await onDiscover(discoveryConfig)
+      if (!result.ok) return
+      const discover = onDiscover ?? (async (candidate: ServerConfig) => {
+        const { discoverMachine } = await import("../machineClient")
+        return discoverMachine(candidate)
+      })
+      const discovered = await discover(discoveryConfig)
       if (!discovered) return
       const available = discovered.agents.filter((agent) => agent.state === "available" || agent.state === "configured")
       const preferred = available.find((agent) => agent.backend === backend) ?? available[0]
--- a/web/src/components/panels.tsx
+++ b/web/src/components/panels.tsx
@@ -19,7 +19,7 @@ import {
   backendSetupCommand
 } from "../backendSetup"
 import type { Translator } from "../i18n"
-import type { BackendKind, FileEntry, ServerConfig } from "../types"
+import type { BackendKind, FileEntry, MachineSnapshot, ServerConfig } from "../types"
 
 /** Breadcrumb pieces for an absolute path, POSIX or Windows, each carrying the path to browse to.
  *  Typing or pasting a deep path and then wanting to step back up one level is the common case, and
@@ -69,13 +69,6 @@ function CopyButton({ text, copyLabel, copiedLabel }: { text: string; copyLabel:
   )
 }
 
-/**
- * Choosing where a new session runs. The old dialog dropped the user into whatever folder the
- * server happened to report and offered a flat list plus a "parent folder" row — no way to see
- * where you were, no way to type a path you already knew, and no shortcut to a project you had
- * opened ten minutes earlier. All three are here now, and the folder that will actually be used is
- * stated in full before the button that uses it.
- */
 export function NewSessionDialog({
   t,
   path,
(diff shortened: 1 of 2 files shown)


web/src/components/panels.tsx
```diff
@@ -420,31 +449,56 @@ export function ConnectServerWizard({
...
+                        key={agent.id}
+                        value={agent.id}
+                        disabled={agent.state !== "available" && agent.state !== "configured"}
+                      >
+                        {agent.label} · {agent.state}
+                      </option>
+                    ))}
+                  </select>
+                  <small className="field-hint">One machine connection; requests are routed to the selected agent.</small>
+                </label>
```
giuliastro, line 505: These strings are hardcoded English in a file that routes every other user-visible string through the translator — 35 `t('…')` calls surround them.

Five in this PR:

- `Agent on {machine.machine.name}` (line 464)
- `One machine connection; requests are routed to the selected agent.` (line 484)
- `Connected to ${discovered.machine.name}` and `${discovered.machine.name} has no available agents` (line 308)
- `· N agents discovered` (line 501)

The app ships English, Italian and two Chinese locales. On any of the three non-English ones, the agent picker — the entire user-facing point of this slice — appears half-translated: the labels around it in Italian, the new control in English.

CI will not catch it. `test:i18n` verifies that keys resolve in every locale; it has no way to notice a string that never became a key. That makes this the class of regression that only surfaces when a non-English user opens the screen.

The keys also need adding to all four locale blocks in `i18n.ts`, and the union type there will enforce that — so the compiler catches a missed locale once the key exists, which is the reason to do it now rather than later.

The agent state shown in the option (`{agent.label} · {agent.state}`) needs the same treatment: `available`, `configured` and `unavailable` are protocol values being rendered as UI text. Worth mapping them through the translator rather than displaying the wire vocabulary.

---

web/src/components/panels.tsx
```diff
@@ -420,31 +449,56 @@ export function ConnectServerWizard({
...
             </>
           )}
         </div>
 
-        {/* Outside the scrolling body on purpose. Testing the connection is the first thing anyone
-            does on this step, and with a keyboard open the body is only a couple of lines tall — a
-            test button living at the end of it scrolled out of sight and came to rest against the
-            save button, which is the one press you do not want to hit by accident. */}
         {step === "credentials" && (
           <div className="wizard-test">
```
giuliastro, line 516: The comment that used to sit here is gone, and it is the one I would least want to lose:

> Outside the scrolling body on purpose. Testing the connection is the first thing anyone does on this step, and with a keyboard open the body is only a couple of lines tall — a test button living at the end of it scrolled out of sight and came to rest against the save button, which is the one press you do not want to hit by accident.

That explains a layout decision which looks arbitrary from the markup alone. `wizard-test` sitting outside the scrolling `div` is exactly the kind of thing someone tidies up while making the wizard taller for the new agent picker — which is what this PR does — and the result is an accidental-save hazard on a phone with the keyboard open.

Three others went with it in this diff: the `NewSessionDialog` doc block, the `ConnectServerWizard` doc block explaining why the wizard replaced the five-field form, and the note that the step indicators double as navigation because a failed test is usually the address rather than the password.

This is the third pull request in a row where explanatory comments unrelated to the change are removed — #149 took nine out of `server.js`, #153 removed the one recording why `isValidServerConfig` exists. Each was restored on request, which is good, but the pattern is now worth naming rather than catching one at a time: these comments are the only place the reasons survive, they are attached to decisions that look wrong until you know the reason, and a reviewer looking at an agent picker has no cause to check whether a UI note three hundred lines away still exists.

Also, the file now ends without a trailing newline (`\ No newline at end of file`). Small, but it makes the next diff of this file noisier than it needs to be.

---

giuliastro (commented):
The flow this establishes is right, and the compatibility work behind it is the careful part: discovery runs first, a null result falls through to the existing backend health test, and the starting harness demotes itself to a hint once a machine answers. Someone connecting to a legacy bridge sees no change at all, which is what makes this safe to ship mid-transition.

Two decisions worth calling out as good. Deriving the saved `backend` from the selected host rather than from the harness the user picked in step one keeps the capability layer correct without a second source of truth — the wizard's first question becomes a guess that discovery is allowed to overrule. And clearing `machine` and `agentId` whenever the credentials change is the right instinct: a discovered machine is only valid for the credentials that discovered it, and stale agent lists are how a picker ends up offering something the server will refuse.

Two findings inline.

**The new UI strings are hardcoded English** in a file where 35 other strings go through the translator, and the app ships four locales. The agent picker — the whole point of this slice — renders in English on an Italian or Chinese UI. `test:i18n` cannot catch it, since it checks that keys resolve rather than that strings became keys.

**Four explanatory comments were removed**, including the one recording why the test button sits outside the scrolling body: with a keyboard open it used to come to rest against the save button. That is a note about a hazard, attached to a layout this PR makes taller, and it is the third consecutive pull request to remove unrelated comments of this kind.

One smaller thing, not worth its own thread. When a machine reports no selectable agents, `agentId` stays empty while the `<select>` still renders every agent as a disabled option — so the browser displays the first unavailable agent as though it were chosen, and only the message below explains that nothing is selectable. Saving is correctly blocked. Rendering a placeholder option in that case would make the state match what the user sees.

A note on #157, which you also asked about: it does not exist. The repository has exactly two open pull requests, this one and #123, so either the number is off by one or it has not been pushed yet.

---

giuliastro:
Correction confirmed: #157 exists as the TaskDeck rebrand/migration issue, not as a pull request. No technical follow-up is required from this comment.

Issue #157, opened 2026-08-13
Rebrand Harness Remote to TaskDeck without breaking existing installs


giuliastro:
## Goal

Adopt **TaskDeck** as the product name while preserving existing user installations, saved profiles, credentials, and upgrade continuity.

## Principle

The rebrand must be **cosmetic/product-facing first, technical identifiers second (or never)**. Users should not need to reinstall the app or re-enter server settings just because the visible product name changes.

## Safe first phase

- change visible product name from Harness Remote to TaskDeck
- update README, screenshots/copy, release titles, desktop productName, Android display label, PWA name/short_name, help text and docs
- update repository name only when redirects/automation have been checked
- preserve existing Android `applicationId`
- preserve Electron app identity/user-data location unless an explicit migration is implemented
- preserve localStorage keys (`opencode.remote.*`) until a tested migration exists
- preserve persisted profile schema and `~/.harness-remote` daemon state paths initially
- keep compatibility aliases for CLI/package names where practical

## Later technical migration (optional)

If we eventually want TaskDeck-native package IDs, storage keys, state directories or executable names, implement explicit read-old/write-new migration with rollback coverage before changing them.

## Acceptance criteria

- upgrading an existing Android install retains profiles and credentials
- upgrading desktop builds retains profiles/settings
- PWA/browser users retain local configuration
- daemon machine identity remains stable across the rename
- old CLI/repository entry points have a documented compatibility path or redirect
- release/update workflows still work after any repository rename
- no existing user is forced to reinstall solely for the rename

## Positioning

**TaskDeck — The local-first control plane for coding work.**

The name intentionally focuses on the work rather than the underlying agents. This issue is branding/migration work and should not block the current Harness 3 implementation sequence.

giuliastro:
Moved from the active issue tracker into the archived Harness 3 roadmap/backlog. The safe rebrand constraints are preserved in `archive/harness-3-2026-08-15:docs/HARNESS_3_ARCHIVE_2026-08-15.md`. Revisit only from a stable baseline and as an independently validated migration.

PR #158, opened 2026-08-13
feat(launcher): auto-start the machine daemon on multi-agent hosts

giuliastro:
## Summary

Final low-friction startup slice of #143.

- keep the existing single-backend quick-start behavior when exactly one supported CLI is detected
- when multiple supported CLIs are present and at least one ACP backend is available, start the machine daemon automatically instead of failing and asking the user to choose
- `--backend` selects the daemon primary on a multi-agent machine, consistently with what the same flag means to the daemon itself
- `--single` is the explicit opt-out back to the legacy single-backend path, and is stripped from child argv as a launcher-only flag
- assign managed OpenCode a free loopback port automatically, excluding the daemon's own, so an OpenCode already running on 4096 does not abort the zero-config path
- validate an explicit `--opencode-port` for availability instead of discovering the conflict after startup
- reuse generated credentials and the existing LAN/port selection path without putting secrets on child argv
- add regression coverage for launch-plan selection, flag precedence, daemon argv construction and port assignment
- update quick-start documentation to describe the automatic daemon path and its current primary-ACP + OpenCode scope accurately

## User-visible change

Before:

```text
harness-remote
  -> detects codex + claude + opencode
  -> exits: choose --backend
```

Now:

```text
harness-remote
  -> detects codex + claude + opencode
  -> starts the machine daemon automatically
  -> selects an ACP primary and prints it
  -> manages OpenCode on a free loopback port
  -> prints one machine address + generated credentials
```

Both overrides remain available:

```bash
harness-remote --backend claude      # daemon, with claude as primary
harness-remote --single --backend codex   # legacy single-backend path
```

## Scope boundary

This does not pretend every detected ACP CLI is instantiated concurrently yet. The automatic daemon path owns one detected ACP primary plus managed OpenCode when present. Other detected ACP CLIs are discovery information for now; the machine API and client transport are already agent-scoped for future expansion.

The `ACP_BACKENDS` order encodes which primary a multi-agent machine gets by default, and now says so in the source — reordering it silently changes that default on every such machine.

Part of #143.

### DIFF
--- a/bridge/src/launcher.js
+++ b/bridge/src/launcher.js
@@ -16,6 +16,7 @@ const BACKEND_EXECUTABLES = {
   opencode: ["opencode"]
 }
 
+const ACP_BACKENDS = ["codex", "claude", "omp", "pi"]
 const VIRTUAL_INTERFACE = /^(docker|br-|veth|virbr|tun|tap|utun)/i
 
 function optionValue(args, name) {
@@ -73,6 +74,25 @@ export function resolveBackend(args, detected = detectBackends()) {
   throw new Error("No supported agent CLI was found on PATH. Install/select omp, pi, claude, codex, or opencode, then re-run with --backend if needed.")
 }
 
+/** Choose the low-friction startup shape without executing any discovered CLI. */
+export function resolveLaunchPlan(args, detected = detectBackends()) {
+  const explicit = optionValue(args, "--backend")
+  if (explicit) return { mode: "single", backend: explicit, detected }
+  if (detected.length === 0) {
+    throw new Error("No supported agent CLI was found on PATH. Install/select omp, pi, claude, codex, or opencode, then re-run with --backend if needed.")
+  }
+  if (detected.length === 1) return { mode: "single", backend: detected[0], detected }
+
+  const primary = ACP_BACKENDS.find((backend) => detected.includes(backend))
+  if (!primary) return { mode: "single", backend: detected[0], detected }
+  return {
+    mode: "daemon",
+    backend: primary,
+    detected,
+    openCode: detected.includes("opencode")
+  }
+}
+
 export function generateCredentials() {
   return {
     username: "harness",
@@ -98,6 +118,12 @@ export function buildBridgeArgs(args, { backend, host, port }) {
   if (!hasOption(result, "--backend")) result.push("--backend", backend)
   if (!hasOption(result, "--host")) result.push("--host", host)
   if (!hasOption(result, "--port")) result.push("--port", String(port))
+  return result
+}
+
+export function buildDaemonArgs(args, { backend, host, port, openCode }) {
+  const result = buildBridgeArgs(args, { backend, host, port })
+  if (!openCode && !hasOption(result, "--no-opencode")) result.push("--no-opencode")
   return result
 }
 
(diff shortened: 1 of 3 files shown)


bridge/src/launcher.js
```diff
@@ -209,6 +257,14 @@ async function main() {
   process.stdout.write(`Username: ${username}\n`)
   process.stdout.write(`Password: ${password}\n`)
 
+  if (plan.mode === "daemon") {
+    const daemonArgs = buildDaemonArgs(args, { backend, host, port, openCode: plan.openCode })
+    process.stdout.write("\nStarting machine daemon...\n")
+    const daemonPath = fileURLToPath(new URL("./daemon-cli.js", import.meta.url))
+    spawnNodeEntrypoint(daemonPath, daemonArgs, username, password)
+    return
+  }
```
giuliastro, line 311: The launcher picks a free port for the daemon and says nothing about the port the daemon is about to need for OpenCode — and on `main` that second port is now fatal rather than degraded.

`daemon-cli.js` calls `ensureOpenCodePortAvailable` before the server listens, and it throws:

> OpenCode port 4096 is already in use on 127.0.0.1. Is OpenCode already running? Use --opencode-port to choose another.

That check is right, and it exists because I asked for it on #151. The problem is what it now means in this path. A user with `opencode serve` already running on 4096 — which is precisely the configuration the documentation told everyone to create until two days ago — types `harness-remote`, and the entire zero-config launch aborts over a port that has nothing to do with the ACP primary they were mostly trying to reach.

That is the friction this PR opens by describing as removed: *"starts the machine daemon automatically instead of failing and asking the user to choose"*. It stops failing over which backend to pick and starts failing over which port OpenCode may use.

The launcher already knows how to solve this for itself. Two options, either acceptable:

- probe from 4096 with `findAvailablePort` and pass `--opencode-port <free>` in `buildDaemonArgs`, so the automatic path picks a working port the same way it picks its own;
- or keep 4096 and let the daemon degrade instead of abort — start without OpenCode, print why, and leave the ACP primary working. The host isolation from #151 already supports reporting one host as unavailable.

The first is better for the promise this PR makes. The second is better if you want the printed OpenCode port to stay predictable. Aborting is the only option that breaks the one-command claim.

---

bridge/src/launcher.js
```diff
@@ -73,6 +74,25 @@ export function resolveBackend(args, detected = detectBackends()) {
...
+
+  const primary = ACP_BACKENDS.find((backend) => detected.includes(backend))
+  if (!primary) return { mode: "single", backend: detected[0], detected }
+  return {
+    mode: "daemon",
+    backend: primary,
+    detected,
+    openCode: detected.includes("opencode")
+  }
+}
```
giuliastro, line 112: `--backend` now carries two incompatible meanings depending on which entrypoint reads it.

Here, passing it forces `mode: "single"` — so on a machine with codex, claude and opencode installed, `--backend codex` is the way to say *do not use the daemon*. In `daemon-cli.js` the same flag names the daemon's ACP primary. Same spelling, opposite intent, and the launcher is the one users will type.

The gap it leaves is the combination someone will want first: run the daemon, but choose which ACP agent is primary. Today the automatic path decides that silently from `ACP_BACKENDS` order and there is no way to override it without falling back to single-agent mode and losing the daemon entirely.

A separate switch would keep both meanings distinct — `--single` to opt out of the daemon, with `--backend` continuing to mean "which agent", consistent across both commands. `--backend` alone could then keep selecting the primary and still imply the daemon when several agents are present.

On the selection itself: `ACP_BACKENDS.find(…)` makes codex beat claude beat omp beat pi. It is deterministic, as the summary says, and the choice is printed, which is what matters most. But the order is unexplained in the source — worth a line saying what it encodes, because the first person to reorder that array for tidiness changes which agent every multi-agent machine picks by default, and nothing will fail.

---

giuliastro (commented):
This closes the loop the launcher opened in #148: a machine with several agents installed used to be the one case the one-command path refused to handle, and now it is the case the daemon exists for. Splitting `resolveLaunchPlan` out as a pure function of the detected list is the right shape — the decision is testable without spawning anything, and the new tests exercise it directly rather than through the CLI.

Two things carried forward correctly from earlier slices: credentials still travel to the child through `bridgeEnvironment` rather than argv, with a test that pins it, and `spawnNodeEntrypoint` re-raises the child's signal instead of swallowing it. Printing both the detected list and the chosen primary is also worth more than it looks — the automatic choice is visible rather than silent, which is the difference between a helpful default and a mystery.

Two findings inline.

**The zero-config path can now abort over the OpenCode port.** `ensureOpenCodePortAvailable` on `main` throws before the daemon listens, so a user with `opencode serve` already on 4096 — the setup the docs prescribed until two days ago — gets a hard failure from `harness-remote` with no arguments. The launcher already solves this for its own port by finding a free one; it should either do the same for the port it implicitly causes the daemon to claim, or let the daemon start without OpenCode and say so. Aborting is the one outcome that contradicts the PR's own framing.

**`--backend` now means two different things** across the two entrypoints: "do not use the daemon" here, "choose the primary" in `daemon-cli.js`. That also leaves no way to run the daemon with a chosen primary, which is the first override someone with codex and claude installed will reach for.

One observation rather than a finding, about the scope boundary. The summary is honest that other detected ACP CLIs are discovery information rather than running hosts — but the client side landed in #156, and its picker lists what `/v1/machine` reports. So a user with codex, claude and opencode installed will see two agents in the wizard and reasonably wonder where claude went. Worth deciding whether the machine snapshot should represent detected-but-not-instantiated agents as a distinct state, so the client can show "installed, not started" instead of omitting them. That is the next slice's problem, but the vocabulary decision belongs to the registry and is cheaper to make before a client depends on the current shape.

Separately: my note on #156 said #157 did not exist. That was wrong — I queried the pull request endpoint and #157 is an issue. If it is worth correcting the record on that thread, say so and I will; I cannot edit posted comments, so it would be a short follow-up comment.

Issue #145, opened 2026-08-12
P0: make New Task reliable end to end


giuliastro:
## Goal

Make New Task boring, fast and trustworthy on Android before enabling it by default.

This single issue now owns the whole task-launch stabilization surface: model catalog/configuration, immediate task visibility, launch lifecycle, finish/restart robustness and recoverable errors.

## Confirmed failures from real-device testing

- model refresh can hang for a long time and then time out;
- reopening may instantly show cached models without saying whether they are current;
- Start navigates away before the task appears;
- the task/session may appear many seconds later;
- opening it can show model state as unavailable/yellow;
- task/session run configuration is not yet trustworthy end to end.

### New Task opens immediately
- never block the whole dialog on model discovery;
- model field has explicit `loading / fresh / stale / unavailable` state;
- live refresh has a short bounded timeout;
- stale cache is clearly marked and never authoritative;
- zero existing user sessions is normal and supported.

### Run configuration is durable
Persist the execution configuration separately from placement:

```text
placement: machine + project + workspace + agent
configuration: model + future variant/attachments/options
```

For model selection:
- catalog → selection → persisted task config → launch revalidation → session/current-model readback;
- no silent fallback;
- if the model disappears, launch fails explicitly and the task remains recoverable;
- ACP catalog implementation may use one hidden reusable technical session per agent if required, never one per task/dialog;
- HTTP agents use native catalog APIs where available.

### Start creates identity before slow work
- POST/create persists a stable task immediately;
- client shows that exact task immediately as `starting`;
- worktree/agent/session startup happens asynchronously;
- same task transitions to `running | failed | completed | interrupted` without manual refresh;
- startup failure stays visible with a reason; no navigation to an empty list.

### Finish/restart robustness
This issue also owns the concrete reliability bugs previously tracked in #174/#175:
- finish is retry-safe if cleanup succeeded but clearing persisted workspace failed;
- already-removed managed worktree is idempotent only on the finish path, while inspect/result still report missing worktree honestly;
- reconciliation persistence failures are surfaced/logged rather than swallowed;
- restart interruption is distinguishable from an agent-reported failure;
- error-code → HTTP mapping is shared rather than duplicated inconsistently.

### Compatibility
- existing session flow keeps working;
- existing legacy direct-server behavior is not changed by task stabilization;
- normal builds keep `TASK_LAUNCH_ENABLED=false` until this issue passes manual acceptance.

## Folded-in issues

[...]

PR #159, opened 2026-08-13
feat(tasks): add machine project discovery and persistent task drafts

giuliastro:


bridge/src/task-store.js
```diff
@@ -0,0 +1,62 @@
...
+    this.loaded = true
+    try {
+      const parsed = JSON.parse(await readFile(this.file, "utf8"))
+      this.tasks = Array.isArray(parsed?.tasks) ? parsed.tasks.filter((task) => task?.machineId === this.machineID) : []
+    } catch (error) {
+      if (error?.code !== "ENOENT") throw error
+      this.tasks = []
+    }
+  }
+
```
giuliastro, line 41: `this.loaded = true` runs before the read, so a failed load is remembered as a successful empty one — and the next write destroys the file.

Trace a `tasks.json` that exists but cannot be read, say after a permissions change or a partial disk error:

1. `create()` → `load()` → `readFile` throws `EACCES` → rethrown, `create()` rejects. Correct so far.
2. The user tries again. `load()` returns immediately because `loaded` is already true, and `this.tasks` is still the constructor's `[]`.
3. `create()` pushes one task onto that empty array and calls `persist()`, which renames a temp file over `tasks.json`.

Every previously saved task is gone, and nothing reported an error on the second attempt. `list()` has the milder version of the same bug: it answers "no tasks" for a file it could not read.

Setting the flag only after a successful parse fixes the read path. The write path deserves its own guard as well — a store that has never loaded successfully should refuse to persist rather than treat an empty in-memory list as the truth:

```js
async load() {
  if (this.loaded) return
  try {
    const parsed = JSON.parse(await readFile(this.file, "utf8"))
    this.tasks = Array.isArray(parsed?.tasks) ? parsed.tasks.filter(…) : []
  } catch (error) {
    if (error?.code !== "ENOENT") throw error
    this.tasks = []
  }
  this.loaded = true
}
```

A malformed `tasks.json` takes the same path, since `JSON.parse` throws a `SyntaxError` with no `code` — which is also worth deciding deliberately. `machine-registry.js` already chose the answer for its own file: preserve the unreadable copy under a `.corrupt-…` name and say so on stderr, rather than silently starting over. Task history is worth at least the same treatment.

---

bridge/src/task-store.js
```diff
@@ -0,0 +1,62 @@
...
+    this.loaded = false
+    this.tasks = []
+  }
+
+  async load() {
+    if (this.loaded) return
+    this.loaded = true
+    try {
+      const parsed = JSON.parse(await readFile(this.file, "utf8"))
+      this.tasks = Array.isArray(parsed?.tasks) ? parsed.tasks.filter((task) => task?.machineId === this.machineID) : []
```
giuliastro: Filtering on load while persisting the whole array means any task belonging to a different machine id is not merely hidden — it is deleted by the next write.

That combination has a live trigger already in the codebase. `loadMachineIdentity` deliberately generates a **new** machine id when `machine.json` is missing, unreadable or invalid, preserving the old file as `.corrupt-…`. So the recovery path introduced in #149 to protect machine identity now quietly destroys task history: new id, every existing task filtered out on load, all of them erased on the first `create()`.

The same shape appears whenever the state directory outlives the identity — a restored backup, a synced home directory, `--state-dir` pointed at shared storage.

Two ways to make it safe, and they are not exclusive. Keep foreign tasks in memory and write them back untouched, filtering only for what `list()` returns — that makes the store additive rather than authoritative. Or key the file by machine id, so a different identity simply reads a different file and cannot overwrite another's.

The second is probably cleaner given that the state directory is already per-machine by intent, and it removes the need to filter at all.

---

bridge/src/project-catalog.js
```diff
@@ -0,0 +1,60 @@
...
+export async function discoverProjects({ machineID, roots }) {
+  const projects = new Map()
+  for (const configuredRoot of roots) {
+    const root = await realpath(configuredRoot)
+    const add = async (candidate, configured = false) => {
+      const resolved = await realpath(candidate)
+      if (projects.has(resolved)) return
+      const git = await isGitProject(resolved)
+      if (!configured && !git) return
+      projects.set(resolved, {
```
giuliastro: Two things about the boundary this function draws.

**One unreadable root takes the whole catalog down.** `realpath(configuredRoot)` on line 27 is the only unguarded call here — the `readdir` below it continues past a failure, and each child `add` is wrapped. But a configured root is the most likely of the three to be wrong: a typo, an unmounted volume, a directory removed since the daemon was configured. When it throws, `discoverProjects` rejects and `/v1/projects` returns nothing at all, including the roots that are perfectly fine. Wrapping it the same way the others are wrapped keeps a broken root from being everyone's problem.

**Symlinked children leave the configured root, which the comment above says they do not.** `add()` resolves each candidate through `realpath`, so `~/dev/external → /etc` is added as a project with `path: "/etc"`. The comment promises discovery does not cross "the configured boundary"; the code crosses it whenever a child is a symlink.

That matters more than a documentation slip because of what this catalog becomes. `/v1/tasks` resolves the workspace path from a catalog entry rather than from the request body — a good decision, and the reason a client cannot name an arbitrary directory. But it makes the catalog the thing that defines where work may run, so anything the catalog can reach is reachable by a task. `--root` is documented as bounding the bridge's own surface even though it is explicitly not a sandbox for the agent, and this widens that surface without saying so.

Resolving the child first and then checking that it is still inside the resolved root is a couple of lines, and keeps the comment true:

```js
const resolved = await realpath(candidate)
if (!configured && !resolved.startsWith(root + path.sep)) return
```

---

giuliastro (commented):
Good groundwork for #145, and one decision in it is better than it looks: `POST /v1/tasks` takes a `projectId` and resolves the path from the server-side catalog rather than trusting a path in the body. That means a client cannot name an arbitrary directory to run work in, and the catalog becomes the single place where "where may work happen" is decided. Worth keeping that property in mind as the launcher lands on top — it is much easier to preserve than to reintroduce.

The rest is careful too: machine routes go through the same authentication helper as the agent routes, the body reader caps at a megabyte instead of buffering whatever arrives, the task store writes atomically at `0600` like the machine registry, and `list()` hands back clones rather than internal state. Discovery being shallow and Git-gated is the right first cut, and it is documented as a choice rather than left implicit.

Three findings inline, the first two on the same file.

**A failed load is remembered as a successful empty one, and the next write deletes the file.** `loaded` is set before the read, so a `tasks.json` that exists but cannot be read or parsed ends with the store persisting a single new task over the top of everything that was there. Moving the flag past the parse fixes the read; refusing to persist from a never-loaded store fixes the write.

**Filtering by machine id on load while persisting the whole array turns "not mine" into "deleted".** That already has a trigger in this codebase: `loadMachineIdentity` generates a new machine id whenever `machine.json` is corrupt, by design — so the identity-recovery path from #149 now silently erases task history the first time a task is created afterwards.

**In the catalog, one bad configured root fails everything**, because that single `realpath` is the only unguarded call in a function that otherwise tolerates failures carefully. And symlinked children resolve outside the configured root, which contradicts the comment directly above them — a documentation slip that matters because this catalog now decides where tasks may run.

None of these change the design. The first two are the ones I would fix before merging, since both end in data loss rather than an error message.

---

PR #160, opened 2026-08-13
feat(tasks): prepare isolated Git worktrees for draft tasks

giuliastro:
## Summary

Second implementation slice of #145.

- add a WorktreeManager that creates deterministic isolated Git worktrees for draft tasks
- keep the user's primary checkout untouched by creating the workspace from HEAD
- expose `POST /v1/tasks/:taskId/worktree` on the machine daemon
- persist the prepared workspace on the Task only after Git succeeds
- roll back a just-created worktree if Task persistence fails before any agent is launched
- reject non-Git projects, non-draft tasks, unknown tasks, and path collisions explicitly
- keep cleanup after launch out of scope so user work is never deleted implicitly
- add regression coverage for deterministic naming, idempotence, rollback, route ordering, and unknown-task handling

## Scope boundary

This PR prepares the isolated workspace but does not launch the selected agent yet. Agent launch and task/run linkage are the next #145 slice.

Part of #145.

### DIFF
--- a/bridge/src/agent-router.js
+++ b/bridge/src/agent-router.js
@@ -2,6 +2,7 @@ import http from "node:http"
 import { allowedOrigin, applyCorsHeaders, matchesCredentials, writeJSON } from "./http-policy.js"
 
 const AGENT_ROUTE = /^\/v1\/agents\/([^/]+)(\/.*)?$/
+const TASK_WORKTREE_ROUTE = /^\/v1\/tasks\/([^/]+)\/worktree$/
 const MACHINE_ROUTES = new Set(["/v1/projects", "/v1/tasks"])
 const HOP_BY_HOP = new Set([
   "connection",
@@ -154,12 +155,14 @@ export function createAgentRoutingServer({
   bridgeServer,
   taskStore,
   projectCatalog,
+  worktreeManager,
   createServer = http.createServer,
   proxyRequest = proxyManagedHttpRequest
 }) {
   return createServer(async (request, response) => {
     const requestURL = new URL(request.url ?? "/", `http://${request.headers.host ?? "localhost"}`)
-    if (MACHINE_ROUTES.has(requestURL.pathname)) {
+    const worktreeMatch = TASK_WORKTREE_ROUTE.exec(requestURL.pathname)
+    if (MACHINE_ROUTES.has(requestURL.pathname) || worktreeMatch) {
       if (!authenticateMachineRequest(request, response, config)) return
       try {
         if (request.method === "GET" && requestURL.pathname === "/v1/projects") {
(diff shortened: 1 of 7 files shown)


bridge/src/agent-router.js
```diff
@@ -192,7 +195,25 @@ export function createAgentRoutingServer({
...
+          }
+          const workspace = await worktreeManager.prepare(task)
+          try {
+            const updated = await taskStore.setWorkspace(taskID, workspace)
+            writeJSON(response, 200, updated)
+          } catch (error) {
+            await worktreeManager.rollback(workspace)
+            throw error
+          }
+          return
```
giuliastro, line 213: The rollback only runs on a throw, but `setWorkspace` has a failure mode that returns instead of throwing:

```js
const index = this.tasks.findIndex((task) => task.id === taskID)
if (index < 0) return undefined
```

So if the task disappears between the `get` above and this call — a concurrent delete, or a reload that drops it — the route takes the success path with `updated === undefined`. Three things then go wrong at once: the freshly created worktree is never rolled back, the client gets a `200` whose body is `JSON.stringify(undefined)`, which is an empty response rather than JSON, and nothing anywhere records that a workspace was created for a task that no longer exists.

That is precisely the case rollback was added for, arriving through the one door it does not watch.

`setWorkspace` already throws for a non-draft task, so throwing for a missing one is the consistent fix and needs no change here. Alternatively treat a falsy result as failure in this block. Either way the invariant worth stating is that this route must never return success without a persisted task, because the worktree on disk is the thing that outlives the mistake.

Worth a test with a store whose `setWorkspace` resolves `undefined`, asserting both the rollback call and a non-2xx response — the existing rollback test covers the throwing path only.

---

bridge/src/worktree-manager.js
```diff
@@ -0,0 +1,63 @@
...
+    try {
+      await this.runGit(["-C", workspace.source, "worktree", "remove", "--force", workspace.path])
+    } catch {
+      return
+    }
+    try {
+      await this.runGit(["-C", workspace.source, "branch", "-D", workspace.branch])
+    } catch {
+      // The worktree is already detached from the project; a leftover branch is safe to keep.
+    }
```
giuliastro: The rollback deliberately tolerates a failed branch deletion — "a leftover branch is safe to keep" — but the branch name is deterministic from the task id, so that leftover makes the task permanently unable to get a workspace.

The sequence: `prepare` creates `task/<key>`, persistence fails, `rollback` removes the worktree, `branch -D` fails for any reason and is swallowed. The task is still a draft whose `workspace.mode` is not `worktree`, so the client retries — and `worktree add -b task/<key>` now dies with *a branch named 'task/<key>' already exists*. The retry cannot succeed, and the error surfaces as a raw git failure that says nothing about the rollback that caused it.

The comment is right that a stray branch is harmless in the repository. It just is not harmless to this task.

`-B` instead of `-b` resolves it directly: the branch is task-scoped and the task is a draft with no work in it, so resetting it to `HEAD` is exactly what a retry wants. Checking for the branch first and reusing it would also work, at the cost of a second git call.

Two smaller things in this file, neither blocking.

`execFile` has no `timeout`, so `worktree add` on a large repository — or one waiting on an index lock held by something else — holds the HTTP request open indefinitely. A bound with a clear message beats a phone client that appears to hang.

And the worktree is created under the daemon's state directory, deliberately and rightly, which puts every task workspace **outside** any configured `--root`. That is fine while nothing runs there, but the next slice launches an agent in that directory, and `--root` is documented as bounding which working directories a session may use. Worth deciding explicitly before the launch slice whether worktree paths are implicitly permitted, rather than discovering that the policy was bypassed by construction.

---

giuliastro (commented):
Reviewing this as a draft, so treat everything below as notes rather than gates.

The shape is right and several of the risky decisions went the safe way. Git is invoked through `execFile` with an argument array, so nothing here can be shell-injected by a project path or a task id. The worktree lands under the daemon's own state directory rather than beside the user's checkout, which keeps `git status` in their repository clean. `rev-parse --show-toplevel` establishes the project really is Git before anything is created, the existing-path check turns a collision into a message instead of a surprise, and `prepare` is idempotent for a task that already has a worktree.

The ordering is the best part: Git first, persistence second, rollback if persistence fails. That is the right way round — a worktree with no task is recoverable, a task pointing at a workspace that was never created is not. And keeping cleanup out of scope so nothing deletes user work implicitly is the correct call for this slice, even though it defers a real question.

Two findings inline.

**Rollback has a door it does not watch.** `setWorkspace` returns `undefined` rather than throwing when the task has disappeared, so that path takes the success branch: no rollback, a `200` with an empty body, and a worktree on disk belonging to nothing.

**The rollback's own tolerance bricks the retry.** Leaving the branch behind is harmless to the repository and fatal to the task, because the branch name is derived from the task id — the next `prepare` for that task fails on an existing branch, permanently.

Two notes rather than findings, both about what happens after this slice. There is no timeout on the git invocation, which on a phone-facing HTTP route means an index lock elsewhere can hang a request with no explanation. And every worktree lives outside any configured `--root`, by construction — fine while nothing executes there, worth an explicit decision before the launch slice puts an agent inside one.

One thing I would think about before the cleanup slice, since it affects storage rather than correctness: each prepared worktree is a full checkout, they accumulate, and on the setup this project actually targets the state directory can be a phone. Nothing yet reports how much has been created or gives a user a way to see it. Not deleting silently is right; leaving it invisible is the part worth solving at the same time.

---

PR #161, opened 2026-08-13
feat(tasks): launch selected agents from prepared task workspaces

giuliastro:
## Summary

Third implementation slice of #145.

- add a task run lifecycle store with `starting`, `running`, and `failed` states
- launch the task's selected agent inside its prepared workspace
- require isolated worktrees before launching Git tasks
- create and persist a run identity before starting agent work
- link the created agent session back to the Task before sending the prompt
- support both ACP-backed agents and managed OpenCode HTTP sessions
- expose `POST /v1/tasks/:taskId/launch` through a machine-level wrapper that delegates all existing agent/project/task routes unchanged
- add regression coverage for lifecycle ordering, worktree enforcement, failure persistence, launch routing, and delegation

## Scope boundary

This PR starts the selected agent and records its session/run identity. It does not yet implement completion detection, finish-work actions, or UI for creating/launching tasks; those remain later #145 slices.

Part of #145.

### DIFF
--- a/bridge/src/task-launch-server.js
+++ b/bridge/src/task-launch-server.js
@@ -1,0 +1,3 @@
+export function taskLaunchPath(pathname) {
+  return /^\/v1\/tasks\/([^/]+)\/launch$/.exec(pathname)
+}
--- a/bridge/src/task-launch-server.js
+++ b/bridge/src/task-launch-server.js
@@ -1,3 +1,52 @@
-export function taskLaunchPath(pathname) {
-  return /^\/v1\/tasks\/([^/]+)\/launch$/.exec(pathname)
+import http from "node:http"
+import { allowedOrigin, applyCorsHeaders, matchesCredentials, writeJSON } from "./http-policy.js"
+
+const TASK_LAUNCH_ROUTE = /^\/v1\/tasks\/([^/]+)\/launch$/
+
+function authenticate(request, response, config) {
+  applyCorsHeaders(request, response, config)
+  if (request.method === "OPTIONS") {
+    response.writeHead(allowedOrigin(request, config) ? 204 : 403)
+    response.end()
+    return false
+  }
+  if (!matchesCredentials(request, config)) {
+    response.writeHead(401, { "WWW-Authenticate": 'Basic realm="Harness Remote Daemon"' })
+    response.end()
+    return false
+  }
+  return true
+}
+
+function launchStatus(error) {
+  const message = error instanceof Error ? error.message : String(error)
+  if (message.startsWith("Unknown task:")) return 404
+  if (message.includes("unavailable")) return 503
+  if (message.includes("Only draft tasks") || message.includes("worktree") || message.includes("workspace")) return 409
+  return 500
+}
+
+export function createTaskLaunchServer({ innerServer, config, taskRunController, createServer = http.createServer }) {
+  return createServer(async (request, response) => {
+    const requestURL = new URL(request.url ?? "/", `http://${request.headers.host ?? "localhost"}`)
+    const match = TASK_LAUNCH_ROUTE.exec(requestURL.pathname)
+    if (!match) {
+      innerServer.emit("request", request, response)
+      return
+    }
+
+    if (!authenticate(request, response, config)) return
+    if (request.method !== "POST") {
+      response.writeHead(405, { Allow: "POST, OPTIONS" })
+      response.end()
+      return
+    }
+
+    const taskID = decodeURIComponent(match[1])
+    try {
+      writeJSON(response, 200, await taskRunController.launch(taskID))
+    } catch (error) {
+      writeJSON(response, launchStatus(error), { error: error instanceof Error ? error.message : String(error) })
+    }
+  })
 }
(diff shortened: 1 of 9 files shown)


bridge/src/task-launcher.js
```diff
@@ -0,0 +1,86 @@
...
+
+  async startPrompt(task, run) {
+    const entry = this.daemon.hostEntry(task.agentId)
+    if (!entry) throw new Error(`Unknown agent: ${task.agentId}`)
+
+    if (entry.kind === "acp") {
+      void entry.host.request("session/prompt", {
+        sessionId: run.sessionId,
+        prompt: [{ type: "text", text: task.prompt }]
+      }, 300_000).catch(() => {})
```
giuliastro: The `failed` state this PR introduces is unreachable for ACP agents once the launch returns.

Not awaiting the prompt is correct — it runs for as long as the agent works, and the launch route must not hold the connection open for it. But `.catch(() => {})` throws the outcome away, so the four ACP-backed harnesses have no path from a failed prompt to the run store. The controller's `catch` only covers what happens before `setRunState("running")`; after that, an agent that rejects the prompt, dies mid-turn, or hits the 300-second timeout leaves the task sitting at `running` and looking healthy.

The HTTP branch below is stricter — it awaits the response and throws on a non-ok status — so the two transports disagree about what `running` guarantees. On OpenCode it means the prompt was accepted; on ACP it means the call was dispatched.

The controller already knows how to record a failure, so the fix is to route the rejection there instead of dropping it:

```js
entry.host.request("session/prompt", { … }, 300_000)
  .catch((error) => void onPromptFailed?.(error))
```

with the controller passing a callback that calls `setRunState(taskID, { status: "failed", run, error })`. That keeps the launch non-blocking while making the failure land where a client can see it.

This matters more than it would in isolation because there is no completion detection yet, by design. Until that slice lands, `running` is a terminal state in practice — so a task that failed at the prompt is indistinguishable from one still working, forever, and the only place the error existed was this discarded promise.

---

bridge/src/task-launch-server.js
```diff
@@ -0,0 +1,52 @@
...
+  return true
+}
+
+function launchStatus(error) {
+  const message = error instanceof Error ? error.message : String(error)
+  if (message.startsWith("Unknown task:")) return 404
+  if (message.includes("unavailable")) return 503
+  if (message.includes("Only draft tasks") || message.includes("worktree") || message.includes("workspace")) return 409
+  return 500
+}
```
giuliastro: The HTTP status contract is derived by matching prose, which makes every one of these error strings load-bearing without saying so.

Two concrete consequences. `message.includes("worktree")` matches far more than the guard it was written for: `WorktreeManager` raises "Worktree path already exists" and git failures mentioning a worktree, and all of them now report `409 Conflict` — a client is told to fix its request when the daemon has an internal problem. And `message.includes("unavailable")` matches that word arriving from anywhere in the stack, including a git or filesystem error that happens to contain it.

In the other direction, rewording any of those messages silently changes the status codes this API returns. Nothing fails, no test notices, and the client's error handling quietly changes meaning.

This is the third time this pattern has come up, and the previous two were both resolved by carrying structure instead of text — `machineClient` moved off `/404|not found/i` to a real status, and `ManagedOpenCodeHost` stopped identifying the credential rejection by its sentence. The same answer applies here: give the controller's errors a `code` (`unknown_task`, `agent_unavailable`, `invalid_state`) and map codes to statuses.

That also makes `launchStatus` testable on its own terms rather than through the exact wording of errors raised three modules away.

---

giuliastro (commented):
Reviewing as a draft. The lifecycle ordering here is the part that had to be right and is: a run identity is persisted before any agent is contacted, the session id is linked back before the prompt is sent, and `running` is only written after both. That ordering means a crash at any point leaves a task whose recorded state is behind reality rather than ahead of it, which is the recoverable direction.

Four things I checked rather than assumed, all of which came out correct:

**No credentials reach disk.** `createSession` returns `base` and `authorization` for the managed HTTP path, and the controller builds `linkedRun` by picking only `sessionId` and `transport` — so the Basic header stays in memory and never enters `tasks.json` or the `/v1/tasks` response. That is easy to get wrong by spreading the session object, and it was not.

**The launch route is authenticated**, through the shared `http-policy.js` — which also means the duplication I flagged on #152 has since been resolved rather than spread further.

**`prompt_async` is the right endpoint.** It is what `api.ts` uses against OpenCode directly, so the launcher speaks the same protocol as the client rather than inventing one.

**`session/new` is bounded**, since `AcpClient.request` applies a default timeout when none is passed — a hung ACP agent cannot hold the launch route open indefinitely.

Two findings inline.

**The `failed` state is unreachable for ACP agents after launch.** Not awaiting the prompt is right, but discarding its rejection means a prompt that fails, dies or times out leaves the task at `running`. With completion detection deliberately out of scope, `running` is currently terminal — so a failed task and a working one are indistinguishable, permanently. Routing that rejection into `setRunState("failed")` keeps the launch non-blocking and makes the new state mean something.

**`launchStatus` derives HTTP codes by matching error prose**, which both over-matches — any message containing "worktree" becomes a 409, including internal `WorktreeManager` failures — and makes every error string a silent part of the API contract. This is the third instance of that pattern; the previous two were fixed by carrying a code instead of a sentence.

One design note for the next slice rather than this one. `running` currently means "the prompt was dispatched", and the two transports do not even agree on that. #131 already defines a normalized vocabulary for exactly this — `working`, `waiting`, `retrying`, plus terminal states supplied explicitly. Mapping task runs onto `AgentRun` now, while there is one producer and no client depending on the shape, is considerably cheaper than reconciling two status vocabularies after the Inbox starts reading both.

---

PR #162, opened 2026-08-13
feat(tasks): add safe worktree lifecycle and run reconciliation

giuliastro:
## Summary

Backend-only continuation of #145.

- Add safe worktree inspection and explicit cleanup APIs.
- Refuse active/dirty cleanup, reject paths outside daemon state, preserve unmerged committed branches, and restore the task workspace to the project after cleanup.
- Keep the existing worktree preparation POST unchanged.
- Remove the unused experimental lifecycle server wrapper.
- Persist task completion and restart reconciliation in the machine-scoped task store.
- Surface ACP prompt completion, use OpenCode status where available, and fail unconfirmed persisted active runs to avoid ghost `running` tasks.
- Guard terminal updates by run id so stale events cannot overwrite a newer run.
- Add backend regressions for worktree lifecycle, route compatibility, and completion reconciliation.

No UI/UX changes and no rebrand/install/storage compatibility changes.

Advances #145.


bridge/src/task-run-controller.js
```diff
@@ -1,58 +1,77 @@
...
-      if (current.status !== "starting" && current.status !== "running") return
-      await this.taskStore.setRunState(taskID, { status: "failed", run: current.run, error })
-    } catch {}
+  async #terminal(taskID, run, status, error = null) {
+    try { await this.taskStore.setRunState(taskID, { status, run, error, expectedRunId: run?.id }) } catch {}
+  }
+
+  async reconcileAll() {
+    for (const task of await this.taskStore.list()) {
+      if (!["starting", "running"].includes(task.status)) continue
```
giuliastro: Reconciliation starts in the constructor and its promise is stored without a `.catch()`, while the `list()` that begins it sits in the `for` header where the surrounding try/catch cannot reach it. So a rejection here becomes an unhandled rejection, and since Node 15 the default for those is to terminate the process.

`TaskStore.load()` still rethrows anything that is neither `ENOENT` nor a `SyntaxError` — `EACCES` on `tasks.json` after a permissions change, `EPERM`, an I/O error — and the `SyntaxError` branch itself calls `rename`, which can fail for the same reasons. Any of those and the daemon exits during startup.

The consequence is out of proportion to the cause: an unreadable task file kills the ACP host and the supervised OpenCode along with it. That is the failure isolation #143 spent four slices establishing, undone from a feature none of those hosts depend on. It is also the least debuggable moment to die — before the listener is up, so nothing answers `/v1/machine` to explain why.

Every route already funnels through `await this.reconciliation`, so the fix keeps that shape and just makes the failure a value rather than a crash:

```js
this.reconciliation = (typeof taskStore?.list === "function" ? this.reconcileAll() : Promise.resolve())
  .catch((error) => { this.reconciliationError = error })
```

with the task routes surfacing `reconciliationError` as a 503 and everything else carrying on. The daemon then starts, the agents work, and the one broken subsystem reports itself — which is the pattern the rest of this codebase already follows for a failed host.

Worth a test that constructs the controller with a store whose `list()` rejects and asserts the daemon still answers, since the current tests only exercise stores that load cleanly.

---

bridge/src/task-launcher.js
```diff
@@ -89,4 +98,31 @@ export class TaskLauncher {
...
+    // ACP gives us an authoritative completion signal while the prompt request is alive,
+    // but there is no backend-neutral post-restart session status primitive to query here.
+    if (entry.kind === "acp") return "unknown"
+    if (entry.kind !== "http") return "unknown"
+
+    try {
+      await entry.host.start?.()
+      const host = entry.host.readinessHost ?? entry.host.host ?? "127.0.0.1"
+      const base = run.base ?? `http://${httpHost(host)}:${entry.host.port}`
+      const authorization = run.authorization ?? basicAuthorization(entry.host.username, entry.host.password)
```
giuliastro: These two fallbacks read fields that are deliberately never persisted, and by existing they invite someone to start persisting them.

`TaskRunController.launch` builds `linkedRun` by picking only `sessionId` and `transport` out of the session — that omission is what keeps the managed host's Basic credential out of `tasks.json` and out of the `GET /v1/tasks` response. Here, `run.base` and `run.authorization` are read from the persisted run, so both branches are unreachable today.

The risk is what happens next. Someone tracing why `inspectRun` recomputes the base URL after a restart finds code that clearly expects `run.base` to be there, concludes the persistence is incomplete, and adds the two fields — at which point a credential lands on disk and in an API response, and nothing fails. The code reads as a missing feature rather than a deliberate boundary.

Deleting the `??` branches makes the intent explicit and loses nothing, since the right-hand sides are what actually runs. If they stay, they need a line saying these must never be persisted and why.

Same reasoning for the `run.authorization` reference in `startPrompt`, which is fed the in-memory session rather than the stored run — correct, and equally non-obvious to a later reader.

---

giuliastro (commented):
This is the most careful slice of the series, and it closes findings from three earlier reviews at once. `prepare` now uses `-B`, so a leftover branch can no longer wedge a retry, and the rollback comment says exactly that. Errors carry codes through `taskLaunchError` and `worktreeError` instead of being classified by prose. `TaskStore.load()` sets its flag after the parse, preserves a malformed file under `.corrupt-…` with a warning, and no longer filters by machine id — which together remove both data-loss paths from #159.

The cleanup design is the part I would have got wrong. Refusing to remove a dirty worktree using `--untracked-files=all` catches the case that actually destroys work — a file the agent created and never staged. `worktree remove` without `--force` and `branch -d` rather than `-D` mean committed-but-unmerged work survives by construction rather than by intention. Blocking cleanup while the task is `starting` or `running` closes the window where an agent is still writing. And the containment check is real: `path.relative` against the resolved worktrees root, rejecting `..` and absolute results, plus refusing a symlinked target.

The run-id guard works as advertised — `expectedRunId` mismatches return the current task untouched instead of overwriting, so a late completion from an old run cannot clobber a newer one. And returning `"unknown"` for ACP in `inspectRun`, with a comment explaining there is no backend-neutral post-restart status primitive, is the honest answer rather than an invented one.

Two findings inline.

**Reconciliation can take the daemon down at startup.** The constructor stores `reconcileAll()` without a `.catch()`, and the `list()` that starts it is outside the loop's try. `load()` still rethrows `EACCES` and friends, so an unreadable `tasks.json` becomes an unhandled rejection — which Node terminates on by default. An unreadable task file should not stop the ACP host and the supervised OpenCode from running, and it certainly should not fail before the listener is up, where nothing can explain why.

**`inspectRun` reads `run.base` and `run.authorization`**, which are deliberately never persisted. The branches are dead today, and they read as an incomplete implementation rather than as a boundary — the obvious "fix" for a future reader is to start persisting them, which puts a credential in `tasks.json` and in the `/v1/tasks` response.

One smaller note, not worth a thread: the containment check resolves lexically, so a symlink at `worktrees` itself — rather than at the target — would still pass. The state directory is the user's own, so this is defence in depth rather than a hole, but `realpath` on both sides closes it for the cost of one call.

---

Issue #163, opened 2026-08-13
feat(tasks): add finish-work result and safe finalization primitives


giuliastro:
## Why

Task completion currently tells us that an agent run stopped, but Harness still lacks a normalized way to inspect the Git result and explicitly finish the workspace lifecycle.

This is deliberately distinct from agent `completed`: completion is a run state; finishing is a user-controlled lifecycle action.

## Scope

- expose a backend-neutral task result summary for managed Git worktrees;
- report dirty/change count, task branch/source heads, commits ahead/behind, and whether task commits are already represented by the source checkout;
- add an explicit finish action for inactive tasks;
- refuse finish for active tasks or dirty worktrees;
- remove a clean managed worktree without force;
- preserve a branch that still contains unmerged committed work;
- restore the task workspace binding to the project after successful finish;
- no UI changes;
- no PR creation or GitHub-specific coupling in this slice.

## API direction

- `GET /v1/tasks/:id/result`
- `POST /v1/tasks/:id/finish`

## Acceptance criteria

- [ ] Result inspection never mutates the repository.
- [ ] Active tasks cannot be finished.
- [ ] Dirty worktrees cannot be finished or removed.
- [ ] Clean worktrees can be released explicitly.
- [ ] Unmerged committed branches survive finish.
- [ ] Non-worktree tasks have a clear no-op finish result.
- [ ] Existing launch/session/worktree APIs remain unchanged.

Follow-up foundation for roadmap P3 review / PR lifecycle.

giuliastro:
Implemented by #164 and merged into `main` as `a36778fd3b5c379b3d1763df14ba8bc7c8d35dfc`.

The finish-work backend foundation now covers result inspection, active/dirty safety guards, explicit clean worktree release, preservation of unmerged committed branches, non-worktree no-op behavior, restart reconciliation before finish/result, and missing-branch handling.

The broader diff/tests/review/PR workflow remains roadmap work, but this issue's scoped primitives are complete.

PR #164, opened 2026-08-13
feat(tasks): add finish-work result and safe finalization primitives

giuliastro:
## Summary

Backend-only follow-up for #163 and the task lifecycle started in #145.

- Add `GET /v1/tasks/:id/result` for normalized Git result inspection.
- Report dirty/change count, task/source heads, commits ahead/behind, and whether task commits are already represented by the source checkout.
- Add `POST /v1/tasks/:id/finish` as an explicit lifecycle action separate from agent-run completion.
- Refuse finish while a task is active.
- Reuse safe worktree cleanup so dirty work is never deleted and unmerged committed branches are preserved.
- Restore the task workspace binding to the project after successful finish.
- Make non-worktree finish a clear no-op.
- Add regressions for divergence reporting, successful finish, and active-task refusal.

No UI/UX changes, no PR creation, and no GitHub-specific runtime coupling.

Advances #163 and #145.

### DIFF
--- a/bridge/src/task-finish-server.js
+++ b/bridge/src/task-finish-server.js
@@ -1,32 +1,18 @@
 import http from "node:http"
-import { allowedOrigin, applyCorsHeaders, matchesCredentials, writeJSON } from "./http-policy.js"
+import { authenticateDaemonRequest, writeJSON } from "./http-policy.js"
 import { inspectTaskWork } from "./task-finish.js"
 
 const resultRoute = /^\/v1\/tasks\/([^/]+)\/result$/
 const finishRoute = /^\/v1\/tasks\/([^/]+)\/finish$/
 
-function authenticate(request, response, config) {
-  applyCorsHeaders(request, response, config)
-  if (request.method === "OPTIONS") {
-    response.writeHead(allowedOrigin(request, config) ? 204 : 403)
-    response.end()
-    return false
-  }
-  if (!matchesCredentials(request, config)) {
-    response.writeHead(401, { "WWW-Authenticate": 'Basic realm="Harness Remote Daemon"' })
-    response.end()
-    return false
-  }
-  return true
-}
-
 function status(error) {
   if (error?.code === "unknown_task") return 404
+  if (error?.code === "agent_unavailable") return 503
   if (["task_active", "worktree_dirty", "invalid_worktree", "worktree_outside_state", "worktree_missing"].includes(error?.code)) return 409
   return 500
 }
 
-export function createTaskFinishServer({ innerServer, config, taskStore, worktreeManager, createServer = http.createServer }) {
+export function createTaskFinishServer({ innerServer, config, taskStore, worktreeManager, taskRunController, createServer = http.createServer }) {
   return createServer(async (request, response) => {
     const pathname = new URL(request.url ?? "/", `http://${request.headers.host ?? "localhost"}`).pathname
     const resultMatch = resultRoute.exec(pathname)
@@ -37,9 +23,17 @@ export function createTaskFinishServer({ innerServer, config, taskStore, worktre
       innerServer.emit("request", request, response)
       return
     }
-    if (!authenticate(request, response, config)) return
+    if (!authenticateDaemonRequest(request, response, config)) return
 
     try {
+      if (taskRunController) {
+        await taskRunController.reconciliation
+        if (taskRunController.reconciliationError) {
+          const error = new Error("Task state is unavailable")
+          error.code = "agent_unavailable"
+          throw error
+        }
+      }
       const taskID = decodeURIComponent((resultMatch ?? finishMatch)[1])
       const task = await taskStore.get(taskID)
       if (!task) {
--- a/bridge/src/task-finish-server.js
+++ b/bridge/src/task-finish-server.js
@@ -1,0 +1,72 @@
+import http from "node:http"
+import { allowedOrigin, applyCorsHeaders, matchesCredentials, writeJSON } from "./http-policy.js"
+import { inspectTaskWork } from "./task-finish.js"
+
+const resultRoute = /^\/v1\/tasks\/([^/]+)\/result$/
+const finishRoute = /^\/v1\/tasks\/([^/]+)\/finish$/
+
+function authenticate(request, response, config) {
+  applyCorsHeaders(request, response, config)
+  if (request.method === "OPTIONS") {
+    response.writeHead(allowedOrigin(request, config) ? 204 : 403)
+    response.end()
+    return false
+  }
+  if (!matchesCredentials(request, config)) {
+    response.writeHead(401, { "WWW-Authenticate": 'Basic realm="Harness Remote Daemon"' })
+    response.end()
+    return false
+  }
+  return true
+}
+
+function status(error) {
+  if (error?.code === "unknown_task") return 404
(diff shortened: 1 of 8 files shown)


bridge/src/task-finish-server.js
```diff
@@ -0,0 +1,72 @@
...
+      }
+      if (task.workspace?.mode !== "worktree") {
+        const result = { managed: false, dirty: false, changeCount: 0 }
+        writeJSON(response, 200, finish ? { task, result, cleanup: { removed: false, branchDeleted: false } } : result)
+        return
+      }
+      if (finish && ["starting", "running"].includes(task.status)) {
+        const error = new Error("An active task cannot be finished")
+        error.code = "task_active"
+        throw error
```
giuliastro, line 52: This is the first task route that reads task state without going through `TaskRunController`, which means it is also the first one that does not wait for reconciliation — and the state it reads is exactly the one reconciliation exists to correct.

`createTaskFinishServer` receives `taskStore` and `worktreeManager` but no controller, and it wraps the launch server, so a request here never reaches the layer where `await this.reconciliation` happens. Every other task route — launch, worktree, cleanup — starts with that await.

The failure it produces is the one a user meets first after a restart. A task persisted as `running` whose agent died with the daemon is still `running` in `tasks.json` until reconciliation rewrites it. `POST /finish` reads that directly, sees an active status, and answers `409 task_active` for a task that has been dead since the process restarted. It keeps answering 409 until something else — a launch, a cleanup, a worktree call — happens to trigger reconciliation first.

And finishing the work left running is precisely what someone does after restarting the daemon, so the broken order is also the likely one.

Passing the controller in and awaiting `runs.reconciliation` before `taskStore.get` is enough; the composition in `createMachineDaemonServer` already has it in scope. Alternatively expose reconciliation as its own awaitable from the daemon so any layer can gate on it without depending on the controller, which would be sturdier as more wrappers appear.

Worth a test that seeds a store with a `running` task, constructs the finish server after a simulated restart, and asserts `finish` succeeds rather than returning 409 — the current tests cover the refusal of a genuinely active task, which is the same status but a different situation.

---

bridge/src/task-finish.js
```diff
@@ -0,0 +1,25 @@
+function output(result) {
+  return String(result?.stdout ?? "").trim()
+}
+
+export async function inspectTaskWork(workspace, worktreeManager) {
+  const status = await worktreeManager.inspect(workspace)
+  const sourceHead = output(await worktreeManager.runGit(["-C", workspace.source, "rev-parse", "HEAD"]))
+  const branchHead = output(await worktreeManager.runGit(["-C", workspace.source, "rev-parse", workspace.branch]))
```
giuliastro: `rev-parse workspace.branch` throws a raw git error when the branch no longer exists, and that turns into a `500` through the fallback arm of `status()`.

A deleted branch is not a broken state, it is the normal state after the work has landed: the user merges the PR, GitHub deletes the head branch, the branch is pruned locally, then the phone opens the task and asks for `/result`. What comes back is `500` with `fatal: ambiguous argument 'harness/task-…'` as the message — indistinguishable from a genuine daemon fault, and there is no way forward from the client.

The rest of this file is careful about honesty (`mergedIntoSource: commitsAhead === 0` computes exactly what it claims). This case deserves the same treatment: catch the failure of that single `rev-parse` and report the branch as gone — for example `branchHead: null`, `commitsAhead: 0`, `commitsBehind: 0`, plus something explicit like `branchMissing: true` — rather than collapsing it into a server error. `mergedIntoSource` would then still read `true`, which is the truthful answer in the overwhelmingly common case that reached this state.

Worth deciding deliberately either way; if you prefer to keep it as a hard error, a distinct `error.code` mapped to `409` would at least let the client say something useful.

---

bridge/src/task-finish-server.js
```diff
@@ -0,0 +1,72 @@
+import http from "node:http"
+import { allowedOrigin, applyCorsHeaders, matchesCredentials, writeJSON } from "./http-policy.js"
+import { inspectTaskWork } from "./task-finish.js"
+
+const resultRoute = /^\/v1\/tasks\/([^/]+)\/result$/
+const finishRoute = /^\/v1\/tasks\/([^/]+)\/finish$/
+
+function authenticate(request, response, config) {
```
giuliastro: Third copy of `authenticate`. `task-launch-server.js` has the same fifteen lines, and so does the layer under it — same CORS call, same `OPTIONS` handling, same realm string, same `401`.

The pieces are already shared (`allowedOrigin`, `applyCorsHeaders`, `matchesCredentials` all come from `http-policy.js`); it is only the assembly that keeps being retyped. Since each wrapper is a security boundary, a divergence between copies is exactly the kind of thing nobody notices — one layer gaining a check the others don't have looks like an ordinary local edit in the diff.

Suggest lifting it into `http-policy.js` as something like `authenticateRequest(request, response, config)` returning a boolean, and having all three wrappers call it. Small change, and it makes the next wrapper — there will be one — a one-liner instead of a fourth chance to get it subtly wrong.

---

giuliastro (commented):
## Review of #164 — finish-work result and safe finalization

Head `385d71c`, both CI jobs green. Read the two new modules in full, plus `machine-daemon.js` to see how the layer is wired in, and `task-run-controller.js` for the reconciliation question below.

**Verdict: good change, one thing worth fixing before merge, two worth deciding on.** Nothing here is dangerous — the design deliberately refuses to destroy work, which is the right instinct for a primitive whose whole job is finalization.

### What is right

- **`status(error)` maps by `error.code`, not by matching prose.** This is the #161 lesson applied without being asked, and it is the difference between an error the client can act on and a string that breaks the moment someone rewords a message.
- **The non-worktree path is a clean no-op with a consistent shape.** `{ managed: false, dirty: false, changeCount: 0 }` for `/result`, and the same object plus a `cleanup` that reports it removed nothing for `/finish`. A caller never has to branch on whether the task happened to be worktree-backed.
- **Reusing `worktreeManager.cleanup` rather than reimplementing removal** means the dirty-work and unmerged-branch guards built earlier apply here for free. Uncommitted changes are never deleted; an unmerged branch survives finalization. That is the correct default for something reached from a phone, where the user cannot inspect before confirming.
- **`mergedIntoSource: commitsAhead === 0` computes precisely what its name claims** — no heuristics about branch names or reflogs, just the count from `rev-list --left-right`.
- **Refusing to finish a `starting`/`running` task** is the right guard, and `409` is the right status for it.

### The one to fix

The reconciliation-bypass comment on lines 44–58: these are the first task routes that read `taskStore` directly instead of going through `TaskRunController`, so they never `await this.reconciliation`. After a daemon restart, a task persisted as `running` whose agent process is gone stays `running` until some *other* route happens to trigger reconciliation — and `/finish` refuses it with `409 task_active` in the meantime.

The reason this matters more than it looks: finalizing left-over work is exactly what a person does immediately after a restart. The first thing they try is the one thing that is broken, and the error text says the task is active when it demonstrably is not.

### The two to decide on

- `rev-parse workspace.branch` turning a deleted branch into a `500` (comment on `task-finish.js:8`). A branch deleted after a merged PR is a normal state, not a fault.
- The third copy of `authenticate` (comment on `task-finish-server.js:8`). Not a bug today; a security boundary that has been retyped three times is a divergence waiting to happen.

### Direction

This is the first slice where the finish-work loop stops being a roadmap bullet and becomes something a client can call. The shape it settles on — `/result` to look before acting, `/finish` to act, both refusing rather than guessing — is the right contract to build the phone UI against. Worth landing once the reconciliation gap is closed.

---

Issue #133, opened 2026-08-11
roadmap: stabilize the core before expanding Harness Remote


giuliastro:
## Current rule

Stop expanding the surface area until the basic product works reliably on a real Android client.

We have been iterating on this for days, so the active plan is now intentionally narrower: **make task creation and model selection work first.** Avoid invasive UX/UI rewrites until that core flow is reliable.

# Active P0

## 1. #145 — New Task + model selection reliability
This is now the top priority.

New Task must:
- open immediately;
- discover models with a short, bounded timeout;
- work even when there are zero existing sessions;
- refresh models without hanging the whole flow;
- clearly distinguish fresh vs stale cached model data;
- never silently fall back when the selected model is unavailable;
- create a visible `starting` task immediately after Start;
- transition to running/failed without requiring a manual refresh;
- persist and display the selected/current model correctly in the resulting task/session;
- remain correct through reopen/restart/finish.

Normal builds keep task launch gated until this flow passes real-device acceptance.

## 2. #143 — daemon foundations needed by New Task
Only daemon work required to make #145 reliable is active now.

Credential policy:
- explicit username/password remain exactly what the user chose across restarts;
- omitted credentials produce a fresh random password for that process only;
- generated passwords are printed and never persisted.

Daemon requirements for task/model flow:
- stable machine identity;
- `/v1/machine` and agent routing are reliable enough for task/model requests;
- no silent default to an uninstalled backend;
- one broken/slow agent does not block the rest of the daemon;
- model/task requests fail quickly and explicitly instead of hanging.

Full universal machine-first UX work is **not** a prerequisite for shipping the task flow.

## 3. #181 — session correctness
Fix the PI/ACP `ciao` + `ciai` message-boundary corruption separately and minimally. Do not mix this fix with task/model or daemon architecture work.

# Lower priority / deferred

## #180 — machine-first client UX
Defer this until the task/model flow is stable.

Do **not** currently redesign the app around one machine with agents as children. That is a larger UX/UI migration and is not required to make New Task useful. Existing server/profile UX remains in place for now, with compatibility preserved.

Fleet, Attention/Inbox, review/PR lifecycle and other orchestration work also remain deferred.

# Compatibility rule

Every stabilization PR must preserve the existing product while the task flow is repaired:

- legacy direct server profiles continue to work;
- profiles with no `agentId` preserve their exact URL behavior;
- existing credentials/settings are never destroyed;
- no forced profile migration;
[...]

PR #166, opened 2026-08-13
docs: align README and roadmap with current task lifecycle

giuliastro:
## Summary

Documentation-only status alignment after #164.

- Move shipped Universal Daemon and backend task/worktree/finish foundations into `What works today`.
- Keep the distinction between backend/API capability and the still-missing task-first client UX explicit.
- Refresh the canonical roadmap with completed #147/#143, current #145 status, completed #163, and the next review/PR + fleet milestones.
- Preserve current positioning and local-first strategy.

Related issue hygiene:
- #163 closed as completed.
- #145 updated with the remaining closure gap.
- #133 updated with current roadmap status.

No runtime, UI, compatibility, storage or branding changes.

### DIFF
--- a/docs/HARNESS_3_ROADMAP.md
+++ b/docs/HARNESS_3_ROADMAP.md
@@ -134,7 +134,7 @@ The daemon now provides stable machine identity, multiple agent-host representat
 As of August 13, 2026:
 
 - ✅ **#147 — One-command startup** is complete.
-- ✅ **#143 — Universal Daemon** is complete.
+- ✅ **#143 — Universal Daemon** is complete as an implementation milestone. Its architecture and mechanics are well covered by tests, but a real ACP-backed harness still needs to be run end to end before heterogeneous daemon compatibility is described as validated.
 - 🟡 **#145 — Create work** has most backend foundations complete: project discovery, normalized tasks, isolated worktrees, agent launch, persisted task/run linkage, restart reconciliation, safe cleanup, result inspection and explicit finish semantics. The major remaining closure gap is the task-first client UX.
 - ✅ **#163 — Finish-work result and safe finalization primitives** is complete through #164.
 - ⏳ Full review/tests/PR lifecycle remains ahead.
@@ -158,6 +158,8 @@ One-command startup and the Universal Daemon established the adoption/runtime ba
 - isolated host health/failure;
 - backward-compatible single-backend paths;
 - fleet-safe machine boundaries.
+
+The architecture/mechanics are implemented, but real heterogeneous multi-host validation still requires at least one reachable ACP-backed harness environment. Test doubles are evidence for the architecture, not proof of real harness compatibility.
 
 #### Current — finish #145 as a product workflow
 
@@ -188,7 +190,7 @@ run → diff → tests/checks → review → PR → CI visibility → finish
 
 Next slices should add these incrementally without coupling the core task model to one forge too early.
 
-#### P2 — Multi-machine Fleet (#146)
+#### Later — Multi-machine Fleet (#146)
 
 Before implementation becomes the largest roadmap investment, validate demand cheaply with existing users/contributors:
 
@@ -209,7 +211,7 @@ Workspace  New worktree
 
 Automatic machine selection comes later.
 
-#### P3 — Coordinate
+#### Later — Coordinate
 
 Only after task/fleet fundamentals are reliable:
 
@@ -300,13 +302,17 @@ Do not prioritize:
 
 The roadmap should remain falsifiable.
 
+Before the multi-agent daemon is described as validated:
+
+- run at least one real ACP-backed harness end to end against it. Test doubles validate architecture and mechanics; they are not evidence of real harness compatibility.
+
 Before #146 becomes the largest build:
 
 - validate real multi-machine demand.
 
 Before backend-specific deferred permission behavior ships:
 
-- validate it against real ACP-backed harness environments rather than only test doubles.
+- validate deferred approval behavior against real ACP-backed harness environments rather than only test doubles.
 
 Before hosted relay or automatic routing:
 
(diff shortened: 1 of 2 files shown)


docs/HARNESS_3_ROADMAP.md
```diff
@@ -311,44 +296,43 @@ Do not prioritize:
...
 The roadmap should remain falsifiable.
 
-Before or during P0B:
+Before #146 becomes the largest build:
 
-- obtain at least one reachable ACP-backed harness environment for real integration validation.
+- validate real multi-machine demand.
 
-Before P2 becomes the largest build:
+Before backend-specific deferred permission behavior ships:
```
giuliastro, line 313: **This is the one change I would not make.** Two things are being dropped here, and both are the honest kind.

The old P0B section carried this sentence:

> Real heterogeneous multi-host validation requires at least one reachable ACP-backed harness environment. Test doubles may validate architecture/mechanics, but they must not be presented as proof of real harness compatibility.

and the old gate read:

> Before or during P0B: obtain at least one reachable ACP-backed harness environment for real integration validation.

The rewrite deletes the first outright and narrows the second to *"before backend-specific deferred permission behavior ships"*. At the same time #143 becomes ✅ complete and the README states the daemon can launch supported agents inside prepared workspaces.

I checked whether the gate was met before it was narrowed. It was not. Every daemon test injects a launcher literal:

```js
new TaskRunController({ taskStore: store, taskLauncher: { async inspectRun() { return "completed" } } })
new TaskRunController({ ..., taskLauncher: { async createSession() { throw new Error("agent unavailable") } } })
```

`task-run-controller.test.js`, `task-completion.test.js` and `task-routes.test.js` all do this. Nothing in `bridge/test/` starts a real Claude Code, Codex, OMP or PI process. The architecture is well covered; real ACP compatibility is not covered at all.

So the sentence was not stale — it describes the current state of the evidence exactly. Deleting it at the same commit that marks the milestone complete removes the only written record that the claim rests on doubles, and does so precisely when it would start to bite.

The mechanical change I would make: move the gate rather than narrow it.

```
Before the multi-agent daemon is described as validated:

- run at least one real ACP-backed harness end to end against it.
  Test doubles validate architecture and mechanics; they are not
  evidence of harness compatibility.
```

`#143 ✅` can stay — the code shipped. What should not disappear is the fact that it shipped without a real-harness integration run. That is a genuinely open item, and it is cheap to keep visible and expensive to rediscover from a bug report.

---

docs/HARNESS_3_ROADMAP.md
```diff
@@ -127,81 +127,66 @@ Fleet control
...
 
 ```text
-run → diff → tests/checks → review → PR
+run → diff → tests/checks → review → PR → CI visibility → finish
 ```
 
-Keep this separate from #145 if necessary to ship sooner, but do not describe task launch alone as the complete table-stakes experience.
+Next slices should add these incrementally without coupling the core task model to one forge too early.
 
 #### P2 — Multi-machine Fleet (#146)
```
giuliastro: Orphan label. `P0A`, `P0B`, `P1` and `P1B` are all gone from this document after the rewrite — replaced by prose headings — but `P2` survives here, and it is now the only one left. A reader meeting `P2` has no P0 or P1 to place it against.

Either rename it in the same style as its new siblings (`#### Later — Multi-machine Fleet (#146)`), or keep the whole ladder. I would rename: the ✅/🟡 status in §6 and the priority ladder in §12 already carry the ordering, so the P-numbers no longer do work.

---

docs/HARNESS_3_ROADMAP.md
```diff
@@ -127,81 +127,66 @@ Fleet control
...
 ```
 
-Each daemon should expose stable machine identity, agent health/capabilities, projects, runs/tasks and an integration point for attention state.
+The daemon now provides stable machine identity, multiple agent-host representation, project/task foundations and fleet-safe ownership boundaries. Machine-scoped identifiers should continue to be designed so a second or third machine can be added without redefining the model.
 
-Machine-scoped identifiers must be designed so a second or third machine can be added without redefining the model.
+## 6. Current implementation status
 
-## 6. Execution sequencing
+As of August 13, 2026:
```
giuliastro, line 134: Minor, but worth deciding now rather than after it has drifted: per-issue status is stated twice in this document — here with ✅/🟡/⏳ prose, and again in §12 as the priority ladder with the same markers. Both have to be edited together on every milestone, and the one that is forgotten will be the one someone reads.

The dated header helps (a reader knows how old the claim is), so I would keep it and instead make §12 purely an *ordering* diagram with no status markers, leaving §6 as the single place that says what is done. Ordering changes rarely; status changes every few days.

---

giuliastro (commented):
## Review of #166 — README and roadmap status alignment

Head `8453973`, documentation only, 70 additions and 80 deletions across two files. I read the full diff and then checked each capability claim against `main` at `a36778f` rather than taking the summary at its word.

**Verdict: the factual claims are accurate and the framing is more honest than the document it replaces. One deletion I would reverse, two small consistency points.**

### The claims check out

Every capability in the new README paragraph exists on `main` today:

| Claim | Where |
|---|---|
| discover known projects | `project-catalog.js` (#159) |
| persist normalized tasks | `task-store.js` (#159) |
| prepare isolated Git worktrees | `worktree-manager.js` (#160) |
| launch supported agents inside those workspaces | `task-launcher.js`, `task-launch-server.js` (`5255f26`) |
| reconcile run state after daemon restarts | `task-run-controller.js` (#162) |
| inspect Git results | `task-finish.js` (#164) |
| release worktrees without deleting dirty or unmerged work | `worktree-manager.js` cleanup guards, exercised by the suite |

And `#163 complete through #164` holds: all three points I raised on #164 landed before it merged — `taskRunController.reconciliation` is now awaited in the finish server with a `503 agent_unavailable` when it failed, `authenticateDaemonRequest` was lifted into `http-policy.js` so the third copy is gone, and `inspectTaskWork` catches the missing branch and reports `branchMissing` instead of returning a raw git `500`.

### The framing is the real improvement

> These task/worktree/finish primitives are currently **backend/API foundations**. The complete task-first client experience — selecting a project, creating work and reviewing the result directly from the app — is still being built.

That paragraph is the most valuable line in the diff. Seven capabilities that no client can exercise yet, listed under a heading that says *What works today*, would read as overclaiming without it. With it, the section is accurate and a visitor knows exactly where the edge is. Same for `🟡 #145` — "most backend foundations complete, the major remaining closure gap is the task-first client UX" is a truthful description of the state.

Reordering *Where it is going* so the shipped items leave and the client UX arrives first is also right: the old list still promised the daemon and worktree isolation as future work months after both shipped.

### The one to reverse

The deleted test-doubles sentence and the narrowed ACP validation gate (comment on §11). The gate has not been met — I checked the suite, every daemon test injects a launcher literal and no real harness is ever started — so the sentence is a current fact, not a stale one, and it is being removed in the same commit that marks the milestone complete.

### The two small ones

- `P2` is now the only surviving P-label after the others were replaced by prose headings (comment on §7).
- Status is stated in two places that must be edited together (comment on §6).

### One observation beyond the diff

This document now describes a backend that is complete through `run → result → finish` and a client that exposes none of it. That gap *is* the project's current risk, and the PR states it plainly in three separate places, which is the right call. Worth being equally plain in the next planning step: everything shipped since #143 is invisible to a user of the app, so the next slice that produces a visible change is worth more than the next slice that deepens the backend.

---

PR #167, opened 2026-08-14
feat(web): expose task-first launch workflow

giuliastro:
Superseded by #172, which consolidates the task-first client work, Android/OpenCode integration fixes, UI cleanup, Fleet foundation and task review summary into one reviewable PR.

Closing this PR to avoid parallel review/merge paths. Nothing from this PR should be merged independently.

### DIFF
--- a/web/src/components/session-list.tsx
+++ b/web/src/components/session-list.tsx
@@ -1,4 +1,4 @@
-import { useState, type PointerEvent as ReactPointerEvent, type RefObject, type ReactNode } from "react"
+import { useEffect, useMemo, useState, type PointerEvent as ReactPointerEvent, type RefObject, type ReactNode } from "react"
 import {
   CloseIcon,
   FolderIcon,
@@ -6,6 +6,7 @@ import {
   LoadingIcon,
   OfflineIcon,
   PencilIcon,
+  PlayIcon,
   PlusIcon,
   RefreshIcon,
   SaveIcon,
@@ -14,6 +15,9 @@ import {
   TrashIcon
 } from "../Icons"
 import type { Translator } from "../i18n"
+import { discoverMachine } from "../machineClient"
+import { loadActiveServerProfile, loadServerProfiles } from "../serverProfiles"
+import { taskCopy } from "../taskCopy"
 import type { HarnessCapabilities, SessionView } from "../types"
 import { TaskLaunchDialog } from "./task-launch-dialog"
 
@@ -45,6 +49,22 @@ export function formatRelativeTime(epoch: number, locale: string): string {
     if (Math.abs(deltaSeconds) >= secondsInUnit) return formatter.format(Math.round(deltaSeconds / secondsInUnit), unit)
   }
   return formatter.format(deltaSeconds, "second")
+}
+
+function useTaskLaunchAvailability(offline: boolean) {
+  const profile = useMemo(() => loadActiveServerProfile(loadServerProfiles()), [])
+  const [state, setState] = useState<"checking" | "available" | "legacy">("checking")
+  useEffect(() => {
+    if (offline) return
+    let cancelled = false
+    void discoverMachine(profile.config).then((machine) => {
+      if (!cancelled) setState(machine ? "available" : "legacy")
+    }).catch(() => {
+      if (!cancelled) setState("legacy")
+    })
+    return () => { cancelled = true }
+  }, [offline, profile])
+  return state
 }
 
 export type SessionRenameState = {
@@ -204,6 +224,9 @@ export function SessionSidebar({
   sessionCardProps: SessionCardProps
 }) {
   const [showTaskLaunch, setShowTaskLaunch] = useState(false)
+  const taskAvailability = useTaskLaunchAvailability(offline)
+  const taskEnabled = !offline && taskAvailability === "available"
+  const taskTitle = offline ? t('sessions.offlineHint') : taskEnabled ? taskCopy(sessionCardProps.language, "newTask") : taskCopy(sessionCardProps.language, "requiresDaemon")
   return (
     <>
       <aside className="desktop-sidebar fade-in" style={{ width, flex: `0 0 ${width}px` }}>
@@ -213,10 +236,10 @@ export function SessionSidebar({
           <button onClick={onRefresh} className="btn-secondary" disabled={refreshing} aria-label={t('sessions.refresh')} title={t('sessions.refresh')}>
             {refreshing ? <LoadingIcon size={16} /> : <RefreshIcon size={16} />}
           </button>
-          <button onClick={() => setShowTaskLaunch(true)} className="btn-primary" disabled={offline} aria-label="New task" title={offline ? t('sessions.offlineHint') : "New task"}>
-            <PlusIcon size={16} /><span className="sidebar-footer-label">Task</span>
+          <button onClick={() => setShowTaskLaunch(true)} className={taskEnabled ? "btn-primary" : "btn-secondary"} disabled={!taskEnabled} aria-label={taskCopy(sessionCardProps.language, "newTask")} title={taskTitle}>
+            <PlayIcon size={16} />
           </button>
-          <button onClick={onNewSession} className="btn-secondary" disabled={creating || offline} aria-label={t('sessions.new')} title={offline ? t('sessions.offlineHint') : t('sessions.new')}>
+          <button onClick={onNewSession} className={taskEnabled ? "btn-secondary" : "btn-primary"} disabled={creating || offline} aria-label={t('sessions.new')} title={offline ? t('sessions.offlineHint') : t('sessions.new')}>
             {creating ? <LoadingIcon size={16} /> : <PlusIcon size={16} />}
           </button>
         </div>
@@ -234,7 +257,7 @@ export function SessionSidebar({
           <button type="button" className="btn-secondary" onClick={onShowSettings} title={t('nav.settings')}><SettingsIcon size={18} /><span className="sidebar-footer-label">{t('nav.settings')}</span></button>
         </div>
       </aside>
-      {showTaskLaunch && <TaskLaunchDialog t={t} onClose={() => setShowTaskLaunch(false)} onLaunched={onRefresh} />}
+      {showTaskLaunch && <TaskLaunchDialog t={t} language={sessionCardProps.language} onClose={() => setShowTaskLaunch(false)} onLaunched={onRefresh} />}
     </>
   )
 }
@@ -283,6 +306,9 @@ export function SessionsPanel({
   sessionCardProps: SessionCardProps
 }) {
   const [showTaskLaunch, setShowTaskLaunch] = useState(false)
+  const taskAvailability = useTaskLaunchAvailability(offline)
+  const taskEnabled = !offline && taskAvailability === "available"
+  const taskTitle = offline ? t('sessions.offlineHint') : taskEnabled ? taskCopy(sessionCardProps.language, "newTask") : taskCopy(sessionCardProps.language, "requiresDaemon")
   return (
     <>
       <section className="panel sessions fade-in">
@@ -297,8 +323,8 @@ export function SessionsPanel({
           </div>
           <div className="inline-actions sessions-header-actions">
             <button onClick={onRefresh} className="btn-secondary" disabled={refreshing}>{refreshing ? <LoadingIcon size={18} /> : <RefreshIcon size={18} />}{t('sessions.refresh')}</button>
-            <button onClick={() => setShowTaskLaunch(true)} className="btn-primary" disabled={offline} title={offline ? t('sessions.offlineHint') : "Start a new task"}><PlusIcon size={18} />New Task</button>
-            <button onClick={onNewSession} className="btn-secondary" disabled={creating || offline} title={offline ? t('sessions.offlineHint') : undefined}>{creating ? <LoadingIcon size={18} /> : <PlusIcon size={18} />}{creating ? t('sessions.creating') : t('sessions.new')}</button>
+            <button onClick={() => setShowTaskLaunch(true)} className={taskEnabled ? "btn-primary" : "btn-secondary"} disabled={!taskEnabled} title={taskTitle}><PlayIcon size={18} />{taskCopy(sessionCardProps.language, "newTask")}</button>
+            <button onClick={onNewSession} className={taskEnabled ? "btn-secondary" : "btn-primary"} disabled={creating || offline} title={offline ? t('sessions.offlineHint') : undefined}>{creating ? <LoadingIcon size={18} /> : <PlusIcon size={18} />}{creating ? t('sessions.creating') : t('sessions.new')}</button>
           </div>
         </div>
         <div className="toolbar"><input placeholder={t('sessions.searchPlaceholder')} value={query} onChange={(event) => onQueryChange(event.target.value)} className="search" /></div>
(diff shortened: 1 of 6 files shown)


web/src/components/session-list.tsx
```diff
@@ -202,32 +203,39 @@ export function SessionSidebar({
...
+    <>
+      <aside className="desktop-sidebar fade-in" style={{ width, flex: `0 0 ${width}px` }}>
+        <div className="resize-handle resize-handle--end" onPointerDown={onResize} role="separator" aria-orientation="vertical" aria-label="Resize panels" />
+        <div className="sidebar-toolbar">
+          <div className="search-field"><SearchIcon size={14} /><input ref={searchInputRef} placeholder={t('sessions.searchPlaceholder')} value={query} onChange={(event) => onQueryChange(event.target.value)} className="search" /></div>
+          <button onClick={onRefresh} className="btn-secondary" disabled={refreshing} aria-label={t('sessions.refresh')} title={t('sessions.refresh')}>
+            {refreshing ? <LoadingIcon size={16} /> : <RefreshIcon size={16} />}
+          </button>
+          <button onClick={() => setShowTaskLaunch(true)} className="btn-primary" disabled={offline} aria-label="New task" title={offline ? t('sessions.offlineHint') : "New task"}>
+            <PlusIcon size={16} /><span className="sidebar-footer-label">Task</span>
```
giuliastro: **This label can never hide, and the sidebar is resizable down to 220px.**

`.sidebar-footer-label` has exactly one rule in the whole stylesheet:

```css
.sidebar-footer {
  container-type: inline-size;   /* line 1024 — the ONLY container in styles.css */
  …
}

@container (max-width: 240px) {  /* line 1039 — the ONLY @container query */
  .sidebar-footer-label { display: none; }
}
```

A container query resolves against the nearest ancestor that establishes containment. `.sidebar-footer` does; `.sidebar-toolbar` does not, and nothing above it does either — I grepped, there is one `container-type` and one `@container` in the file. So for a `.sidebar-footer-label` placed inside the *toolbar*, the query has no container to match against and never applies. The word "Task" stays rendered at every width.

What that produces concretely. `SIDEBAR_WIDTH_MIN` is 220 (`App.tsx:85`), below the 240px threshold, so dragging the sidebar to its minimum gives you a footer that has correctly collapsed to icons and a toolbar directly above it that has not — the two rows disagree inside the same 220px column. And the toolbar's flex arithmetic makes it worse: `.search-field` is `flex: 1 1 auto; min-width: 0`, so it is the only thing that can give. Refresh, Task-with-label and New-session hold their intrinsic widths and the search input is squeezed toward nothing.

Two things to fix, and they are separable:

1. **The label.** Either give `.sidebar-toolbar` its own `container-type: inline-size` and a toolbar-specific rule, or introduce `.sidebar-toolbar-label` with its own threshold. Reusing a class whose only rule lives in another subtree's container query looks like it works and silently does not.
2. **The two plus icons.** Once the label is gone at narrow widths, the toolbar reads `+ +` — two identical `PlusIcon size={16}` buttons side by side, distinguishable only by `btn-primary` versus `btn-secondary`. Give the task button a different glyph, or make the new-session action a menu item under one primary button. Right now the only affordance separating "start a task" from "start a session" is a fill colour.

Worth also deciding deliberately about the primary-action swap this line makes: New Session was `btn-primary` and is now `btn-secondary`, on both surfaces. That is defensible for a task-first product, but it silently moves every existing user's muscle memory to a button that, for a non-daemon profile, cannot succeed (see my comment on the dialog). I would swap the emphasis in the same release that makes the task path work for everyone, not before.

---

web/src/components/task-launch-dialog.tsx
```diff
@@ -0,0 +1,142 @@
...
+      setStarting(false)
+    }
+  }
+
+  return (
+    <div className="modal-backdrop" role="presentation" onClick={onClose}>
+      <section className="modal-card wizard fade-in" role="dialog" aria-modal="true" aria-labelledby="new-task-title" onClick={(event) => event.stopPropagation()}>
+        <div className="wizard-header">
+          <div className="wizard-header-text">
+            <h2 id="new-task-title">New Task</h2>
```
giuliastro: **Every string in this dialog is hardcoded English except one.**

`t` is imported and threaded in as a prop, then used exactly twice — both times for `t('session.cancel')`. Everything else the user reads is a literal: "New Task", "Start isolated agent work on …", "Project", "Task", "Agent", "Machine", "Loading machine projects and agents…", "Describe the work the agent should complete…", "Use a new isolated Git worktree", "Start Task", "Starting…", both explanatory paragraphs, and the two thrown error messages on lines 41 and 48.

The app ships four locales — `en`, `it`, `zh-TW`, `zh-CN` (`i18n.ts:1`) — and the wizard this one is modelled on translates every single string; `panels.tsx` has no English literal in it at all. So on an Italian or Chinese device the new primary surface of the product opens as a fully English panel with one Italian button in the corner. That is the most visible possible place for the seam to show.

`i18n.test.mjs` did not catch it because it asserts specific known keys rather than scanning components for literals, so CI being green says nothing here.

This is mechanical rather than difficult: add a `task.*` group to `TranslationKey` and the four locale tables, and route the strings through `t`. Worth doing before merge rather than as a follow-up — a translation pass over a shipped screen is much more tedious than writing the keys while the copy is still being decided.

Two related copy points while the strings are being moved:

- The same action is spelled three ways: `aria-label="New task"`, button text `New Task`, tooltip `Start a new task` (`session-list.tsx:216` and `:300`). Pick one and let the key enforce it.
- `<h2>New Task</h2>` with `<p className="subtle">Start isolated agent work on {machine name}</p>` reads well, but the subtitle asserts isolation before the user has confirmed the worktree checkbox further down — and for a non-Git project the dialog itself says the opposite ("the task will run in the project directory"). Make the subtitle unconditional ("Start agent work on …") and let the checkbox own the isolation claim.

---

web/src/components/task-launch-dialog.tsx
```diff
@@ -0,0 +1,142 @@
...
+
+              <div className="folder-picker-current">
+                <span className="eyebrow">Machine</span>
+                <strong><ServerIcon size={15} /> {machine?.machine.name ?? "Harness daemon"}</strong>
+              </div>
+
+              <label className="field" style={{ flexDirection: "row", alignItems: "center", gap: "0.65rem" }}>
+                <input type="checkbox" checked={isolated} onChange={(event) => setIsolated(event.target.checked)} disabled={selectedProject?.kind !== "git"} />
+                <span><FolderIcon size={14} /> Use a new isolated Git worktree</span>
+              </label>
```
giuliastro, line 130: Three things in this block work against the "reuse the design system" intent that the rest of the file honours well.

**1. The inline style override (line 121).**

```tsx
<label className="field" style={{ flexDirection: "row", alignItems: "center", gap: "0.65rem" }}>
```

`.field` is a stacked label/control pair; this reaches in and turns it into a row with a magic `0.65rem` that matches no spacing token — the stylesheet works in `var(--space-*)`. A checkbox row is a recurring shape, not a one-off, so it deserves a real class (`.field-inline`, or a `.checkbox-row` alongside `.inline-actions`). Inline overrides also lose the theme: they cannot participate in a media query, a container query or a future density setting, so this is the one row in the dialog that will not respond when the rest does.

**2. `folder-picker-current` for the Machine row (line 116).**

That class belongs to the folder-picker's "you are here" chrome. Borrowing it for a machine identity row means the two are now coupled by accident: whoever adjusts the folder picker's presentation silently restyles this. The visual result is fine today — it is the ownership that is wrong. A small `.context-row` used by both, or a dedicated class, keeps them independent.

**3. The Agent select is not a choice (line 31, rendered at 110).**

```ts
const profileAgent = availableAgents.find(…)
const agents = profileAgent ? [profileAgent] : []
```

`agents` is always length 0 or 1, so the `<select>` renders as a dropdown the user can open and find nothing to pick — or, when empty, as a disabled empty control with no explanation. A select with one option is a control that advertises an affordance it does not have.

Since the machine is already presented as a static row two lines below, the agent should match it: same treatment, showing the agent name and its icon, with the "cross-agent placement arrives with Fleet" line as the explanation. Then the dialog has two facts (Machine, Agent) and two decisions (Project, Task) plus one option (worktree) — which is a much clearer read than four controls of which one is inert.

While rearranging: the current order is Project → Task → Agent → Machine → worktree. Context should precede decisions and options should follow them — Machine → Project → Agent, then the Task textarea, then the worktree checkbox. As it stands the machine you are about to run work on is disclosed *after* you have already written the prompt.

---

web/src/components/task-launch-dialog.tsx
```diff
@@ -0,0 +1,142 @@
...
+        const activeAgent = selectable.find((agent) => config.agentId ? agent.id === config.agentId : agent.backend === config.backend)
+        if (!activeAgent) {
+          const label = config.agentId ?? config.backend
+          throw new Error(`The active agent ${label} is unavailable on this machine.`)
+        }
+        setMachine(discovered)
+        setProjects(knownProjects)
+        setProjectId(knownProjects[0]?.id ?? "")
+        setAgentId(activeAgent.id)
+      } catch (cause) {
```
giuliastro, line 54: **For a direct-OpenCode profile this dialog can only ever fail, and it is now the primary button.**

`discoverMachine` falls back to legacy mode on 404 and on a registry-less 503 — `machineClient.ts` does this deliberately, and `server-config.test.mjs` asserts it. A user pointed at a plain `opencode serve` therefore gets `null` here, which becomes a thrown error, which renders as `✗ Task launch requires a Harness machine daemon.` inside an otherwise empty modal.

That is the *whole* interaction for anyone who has not adopted the daemon: press the most prominent button in the sidebar, wait for a spinner, read an error, close. And it is not a small population — it is every existing installation from before #143, which is the compatibility path this repository has otherwise been careful to keep working.

The dialog is not the right place to discover this. The profile already knows: `discoverMachine` runs at connection time for the session surface. Gate the button on that result — hide it, or disable it with `title` explaining that task launch needs the machine daemon and pointing at the one-command startup. An action that is visibly unavailable with a reason is a good experience; an action that looks available and fails after a network round trip is not.

Two smaller gaps in the same effect:

- **Zero projects renders as an empty disabled dropdown** (line 52: `setProjectId(knownProjects[0]?.id ?? "")`). If the daemon has no configured roots, the user sees a `<select>` with nothing in it, a disabled Start button, and no text anywhere saying why. This is a state people will actually hit — configuring roots is not obvious, and the failure is silent. It needs the same `empty-state` treatment the folder picker already has, naming what is missing.
- **`loading` and `error` are not mutually exclusive with the form.** On the error path the body renders the full set of controls plus the error, so a user can type a prompt into a dialog that has already established it cannot launch anything. Returning early to an error-only body would be both simpler and less misleading.

---

web/src/taskClient.ts
```diff
@@ -0,0 +1,123 @@
...
+    response = await fetch(target, {
+      method,
+      headers,
+      body: options.body === undefined ? undefined : JSON.stringify(options.body)
+    })
+  } catch {
+    throw new Error(`Cannot reach ${config.host}:${config.port}.`)
+  }
+  if (!response.ok) {
+    let detail = `HTTP ${response.status}`
```
giuliastro, line 100: A 401 from this client surfaces as the bare string `HTTP 401`.

`api.ts` already learned this lesson and encodes it:

```ts
function unauthorizedDetail(config: ServerConfig): string {
  return hasCredentials(config)
    ? "HTTP 401: the server rejected these credentials."
    : "HTTP 401: this server requires a username and password, and none were sent."
}
```

The distinction exists because those two states look identical from the app and are completely different to act on — one means re-check the password, the other means a field is empty and nothing was ever sent. `server-config.test.mjs` even asserts the message is present, precisely so it cannot regress.

It is module-private today, so this needs `export` on that function and a call here in both branches (native at line 85, fetch at line 100). Otherwise the task path reintroduces exactly the ambiguity the session path was fixed for, on a screen where the user has just typed a prompt and has the least patience for a vague failure.

Same shape applies to the transport error: `Cannot reach ${config.host}:${config.port}.` duplicates a string `api.ts` also produces. Worth sharing rather than re-typing — three copies of a user-facing error message drift.

---

giuliastro (commented):
## Review of #167 — task-first launch workflow

Head `a4b43fa`, 385 additions across 5 files. I read the new files in full, the existing wizard in `panels.tsx` for comparison, and the parts of `styles.css` and `App.tsx` that govern how the new controls behave at real widths.

**Verdict: the architecture is right and the transport fix is genuinely good. The interface is not ready — there is one provable layout defect, the dialog is untranslated, and the primary button is a dead end for pre-daemon installations.** None of it is deep; all of it is visible to the first person who opens the app.

### What is well done

- **The Electron scoping fix is the right shape.** `isMachineScopedPath` replaces a two-value equality check with an explicit list, `/v1/tasks/` included by prefix, and agent scoping is preserved for everything else. `machine-task-transport.test.mjs` pins both directions — a machine path staying unscoped *and* an ordinary session path still getting the agent prefix. That second assertion is the one that matters, and it is there.
- **Legacy profiles without `agentId` are handled at both places that need it** — the effect and the render agree on the same fallback rule, so a migrated profile resolves its agent by backend rather than failing.
- **Reusing the session UI instead of building a task-detail screen** is the correct call. A second detail experience would have doubled the surface for no user-visible gain, and the PR says so explicitly.
- **`taskClient` is cleanly separated by transport** — desktop, Capacitor native and fetch, each with its own error path, and `responseDetail` unwraps a JSON error body rather than showing a status code when the daemon has said something useful.
- Worktree isolation defaulting on for Git projects, with a stated fallback for non-Git ones, is the right default.

### What needs to change before this ships

Ordered by how likely a user is to hit it:

1. **`sidebar-footer-label` in the toolbar can never hide** (comment on `session-list.tsx:216`). Its only rule lives in the stylesheet's single `@container` query, anchored to `.sidebar-footer`, which is the stylesheet's single container. In the toolbar the query has nothing to resolve against. At `SIDEBAR_WIDTH_MIN = 220` — below the 240px threshold — the footer collapses to icons and the toolbar above it does not, in the same 220px column, while the search field is squeezed to make room. This is the one item I would call a defect rather than a judgement.
2. **The dialog is in English on an app that ships four locales** (comment on `task-launch-dialog.tsx:85`). One `t()` call, fourteen literals. `panels.tsx` has zero literals. On an Italian device this opens as an English panel with a single Italian button.
3. **A direct-OpenCode profile can only ever see an error** (comment on `task-launch-dialog.tsx:39`). `discoverMachine` returns `null` for pre-daemon servers by design; here that becomes a thrown error behind the most prominent button in the sidebar. Gate the button on the discovery the app already performs.
4. **Design-system fidelity in the dialog body** (comment on `task-launch-dialog.tsx:110`): an inline `flexDirection`/`0.65rem` override of `.field`, `folder-picker-current` borrowed for the Machine row, and an Agent `<select>` that always holds exactly one option. Plus an ordering that reveals the target machine after the prompt has been written.
5. **`HTTP 401` loses the distinction `api.ts` already draws** (comment on `taskClient.ts:85`).

### On the interface standard

Since it matters here: the parts of this that follow the existing system — modal/wizard structure, `btn-primary`/`btn-secondary`, `empty-state compact`, `fade-in`, icon sizing — are correct and consistent, and the dialog reads as part of the app rather than bolted on. The problems are all at the edges where it *stops* following it: one inline style, one borrowed class, one class used outside the subtree its rules live in, and a control shaped like a choice that is not one. Those edges are exactly where an interface stops feeling considered, and all four are cheap to close now.

I would hold the merge for items 1–3 and take 4–5 in the same pass, since they touch the same two files.

PR #168, opened 2026-08-14
feat(fleet): add resilient multi-machine client foundation

giuliastro:
Superseded by #172, which consolidates the Fleet client foundation with the task-first workflow and the integration fixes discovered during Android testing.

Closing this stacked PR to leave a single review path. Nothing from this PR should be merged independently.

### DIFF
--- a/web/src/fleetClient.ts
+++ b/web/src/fleetClient.ts
@@ -7,8 +7,11 @@ export async function loadFleet(profiles: SavedServerProfile[]): Promise<FleetMa
   return discoverFleet(profiles, async (config) => {
     const machine = await discoverMachine(config)
     if (!machine) throw new Error("Harness machine daemon is unavailable")
-    const projects = await taskClient.listProjects(config)
-    return { machine, projects }
+    const [projects, tasks] = await Promise.all([
+      taskClient.listProjects(config),
+      taskClient.listTasks(config)
+    ])
+    return { machine, projects, tasks }
   })
 }
 
--- a/web/src/fleetClient.ts
+++ b/web/src/fleetClient.ts
@@ -1,0 +1,37 @@
+import { discoverMachine } from "./machineClient"
+import { discoverFleet, type FleetMachine } from "./fleetModel"
+import type { SavedServerProfile } from "./serverProfiles"
+import { taskClient, type MachineTask } from "./taskClient"
+
+export async function loadFleet(profiles: SavedServerProfile[]): Promise<FleetMachine[]> {
+  return discoverFleet(profiles, async (config) => {
+    const machine = await discoverMachine(config)
+    if (!machine) throw new Error("Harness machine daemon is unavailable")
+    const projects = await taskClient.listProjects(config)
+    return { machine, projects }
+  })
+}
+
+export async function launchFleetTask(machine: FleetMachine, input: {
+  projectId: string
+  agentId: string
+  prompt: string
+  isolated?: boolean
+}): Promise<MachineTask> {
+  if (machine.state !== "online" || !machine.machine) throw new Error("The selected machine is unreachable")
+  if (!machine.agents.some((agent) => agent.id === input.agentId && (agent.state === "available" || agent.state === "configured"))) {
+    throw new Error(`Agent ${input.agentId} is unavailable on ${machine.machine.name}`)
+  }
+  const project = machine.projects.find((candidate) => candidate.id === input.projectId)
+  if (!project) throw new Error(`Unknown project on ${machine.machine.name}`)
+
+  let task = await taskClient.createTask(machine.config, {
+    projectId: input.projectId,
+    agentId: input.agentId,
+    prompt: input.prompt
+  })
+  if (input.isolated !== false && project.kind === "git") {
+    task = await taskClient.prepareWorktree(machine.config, task.id)
+  }
+  return taskClient.launch(machine.config, task.id)
+}
--- a/web/src/fleetModel.ts
+++ b/web/src/fleetModel.ts
@@ -1,0 +1,86 @@
+import type { SavedServerProfile } from "./serverProfiles"
+import type { MachineSnapshot, ServerConfig } from "./types"
+import type { MachineProject } from "./taskClient"
+
+export type FleetMachine = {
+  key: string
+  profileIds: string[]
+  profileNames: string[]
+  config: ServerConfig
+  state: "online" | "unreachable"
+  machine: MachineSnapshot["machine"] | null
+  agents: MachineSnapshot["agents"]
+  projects: MachineProject[]
+  error?: string
+}
+
+export type FleetObservation = {
+  machine: MachineSnapshot
+  projects: MachineProject[]
+}
+
+export type FleetDiscover = (config: ServerConfig) => Promise<FleetObservation>
+
+/**
+ * Profiles for different agents on the same daemon share one machine endpoint. Fleet discovery must
+ * contact that endpoint once, not turn five saved agent profiles into five fake machines.
+ */
+export function machineEndpointKey(config: ServerConfig): string {
+  const raw = config.host.trim().replace(/\/+$/, "")
+  const normalized = /^https?:\/\//i.test(raw) ? raw.toLowerCase() : `http://${raw.toLowerCase()}`
+  return `${normalized}:${config.port}`
+}
+
+export function groupProfilesByMachineEndpoint(profiles: SavedServerProfile[]): Map<string, SavedServerProfile[]> {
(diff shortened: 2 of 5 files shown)


web/src/fleetModel.ts
```diff
@@ -0,0 +1,115 @@
...
+        config: profile.config,
+        state: "online",
+        machine: observation.machine.machine,
+        agents: observation.machine.agents,
+        projects: observation.projects,
+        tasks: observation.tasks.map((task) => ({ ...task, fleetId: fleetTaskID(machineID, task.id) }))
+      }
+    } catch (cause) {
+      return {
+        key: endpoint,
```
giuliastro, line 95: `key` carries two different identities depending on which branch produced the row, and neither is used for the deduplication the module exists to do.

**Duplicate machines survive.** Grouping happens on `machineEndpointKey`, a normalized `host:port` string. Two profiles reaching the same daemon by different names — `localhost:4097` and `192.168.1.64:4097`, or a `.local` hostname and its IP — normalize to two distinct strings, so they form two groups, so `discoverFleet` contacts the daemon twice and returns two `FleetMachine` rows. Both then set `key: machineID` (line 83) and both carry the same one, so the result is two rows with the same key and the same agents, projects and tasks.

That combination — one daemon appearing as two machines, with a duplicated React key — is the exact failure the header comment says this module prevents, only one level up from where the guard was placed. And it is not hypothetical: a laptop reachable as `localhost` from itself and by LAN address from a phone is the normal setup for this project.

The daemon publishes a stable machine identity precisely so this is solvable. Endpoint grouping is a good cheap pre-pass — keep it — but the authoritative pass has to be after discovery: fold rows that resolve to the same `machine.id`, union their `profileIds`/`profileNames`, and keep one config (preferring the one that answered).

**Identity is not stable across reachability.** An online row is keyed by `machineID`, an unreachable one by `endpoint` (line 95). So when a machine drops, its key changes, and any list keyed on it unmounts the row and mounts a new one — selection, expansion and scroll position are lost at the exact moment the user is looking at the machine that just went offline. It also means the offline row cannot be matched back to the online row it replaced.

Endpoint is the only identity available before a successful discovery, so it should be the *stable* one: key every row by endpoint, and carry `machineId` as a separate field that is `null` until known. Then the row is continuous across reachability changes and machine identity is what you fold duplicates on, which is what it is for.

Neither case is covered by `fleet-model.test.mjs` today — the four `discoverFleet` tests each use one endpoint per machine. Two profiles with different hosts resolving to the same `machine.id`, and one machine flipping online → unreachable across two calls, would pin both.

---

web/src/fleetModel.ts
```diff
@@ -0,0 +1,115 @@
...
+  let lastError: unknown
+  for (const profile of machineProfiles) {
+    try {
+      return { profile, observation: await discover(profile.config) }
+    } catch (cause) {
+      lastError = cause
+    }
+  }
+  throw lastError ?? new Error("No usable profile is configured for this machine")
+}
```
giuliastro, line 67: The fallback works, but it reports the wrong failure.

Profiles are tried in stored order and `lastError` wins, so the message that reaches the user is whatever the *final* profile said — not necessarily the informative one. Concretely: profile A holds stale credentials (`HTTP 401: the server rejected these credentials`) and profile B has a typo in the host (`Cannot reach workstaton.local:4097`). The machine is reported unreachable because of the typo, and the credential problem — the one the user can actually fix, and the reason the fallback engaged at all — is discarded.

Two small changes make the behaviour match the intent stated in the comment above `discoverFleet`:

- collect all failures, not just the last, and surface the most actionable one — a 401 tells you more than a connect failure, since it proves something is listening;
- name the profile in the error, because with several profiles per endpoint "this machine is unreachable" does not tell you which saved credential to go fix.

Worth handling now rather than after the Fleet UI exists: once a row just says "unreachable", the reason is the only thing standing between the user and a working machine, and it will be read on a phone with no console to check.

---

web/src/fleetClient.ts
```diff
@@ -0,0 +1,40 @@
...
+    if (!machine) throw new Error("Harness machine daemon is unavailable")
+    const [projects, tasks] = await Promise.all([
+      taskClient.listProjects(config),
+      taskClient.listTasks(config)
+    ])
+    return { machine, projects, tasks }
+  })
+}
+
+export async function launchFleetTask(machine: FleetMachine, input: {
```
giuliastro, line 18: `launchFleetTask` and `TaskLaunchDialog.start()` are now two implementations of the same sequence, and this one is the better of the two.

The dialog (from #167) does: create → prepare worktree if `isolated && kind === "git"` → launch. This does the same, plus it verifies the agent is actually `available`/`configured` on the target machine and that the project exists there before creating anything — checks the dialog skips because it trusts the state it loaded at mount.

Two paths that must agree about worktree defaults, agent validity and ordering will drift; the giveaway is that they already disagree on the isolation default (`isolated !== false` here, `isolated` there — the same value, expressed as opposite defaults).

Since #167 is the parent branch, the cheap resolution is to land `launchFleetTask` as *the* launch function and have the dialog call it with a single-machine `FleetMachine`, rather than keeping the dialog's copy until the Fleet UI arrives. The dialog gains the two validity checks for free, and the pre-flight failures become "this agent is no longer available on that machine" instead of a raw daemon error after three round trips.

---

giuliastro (commented):
## Review of #168 — Multi-machine Fleet foundation

Stacked on #167; reviewed only what this branch adds on top (`fleetModel.ts`, `fleetClient.ts`, `fleet-model.test.mjs`, the `taskClient.listTasks` addition and the test-script change). My #167 comments are not repeated here.

**Verdict: the right model, and the isolation properties are correct. Two identity problems to settle before a UI is built on it — cheap now, expensive once rows are rendered and keyed.**

### What is right

- **Failure isolation is the property that matters and it holds.** `Promise.all` over groups where every group catches internally means one dead laptop yields one `unreachable` row and never rejects the fleet load. The test covers it. This is the thing that would have made a naive implementation useless on a real network, and it was got right first time.
- **Profile fallback within an endpoint** is a genuinely good idea I would not have thought to ask for: several agent profiles on one daemon, one with stale credentials, and the machine still comes up. It matches how profiles actually accumulate.
- **`fleetTaskID` encodes both halves**, so a machine id containing a separator cannot produce an ambiguous composite. Small thing, done correctly.
- **Sorting online-before-offline, then by name**, is the right default for a list whose purpose is "where can I put work".
- **`launchFleetTask` validates before it mutates** — agent availability and project membership are checked before a task row is created, so a stale UI cannot leave orphan draft tasks on the daemon.
- Deferring the Fleet UI was the correct call. This is a model with tests and no screen, which is exactly the reviewable unit.

### What to settle first

1. **Endpoint grouping does not deduplicate machines** (comment on `fleetModel.ts:83`). Two profiles reaching one daemon by different names — `localhost` and the LAN address, the ordinary case for a laptop plus a phone — produce two rows carrying the *same* `machineID` as `key`. One daemon, two machines, duplicate keys: the failure the module's own comment sets out to prevent, one level above where the guard sits.
2. **`key` means `machineID` when online and `endpoint` when not** (same comment). A machine going offline changes its identity, so its row is destroyed and recreated rather than updated. Endpoint is the identity that exists in both states; machine id is the identity you fold duplicates on. Swapping those two roles fixes both problems at once.
3. **The fallback reports the last error rather than the useful one** (comment on `fleetModel.ts:55`), which can hide a 401 behind an unrelated connect failure.
4. **Two launch paths now exist** (comment on `fleetClient.ts:18`), already disagreeing on how the isolation default is expressed.

### On sequencing

Stopping before the Fleet UI, and before arbitrary test execution and PR creation, was the right place to stop — and the reason you gave for it is the correct one. Running tests and opening PRs on a user's behalf is not a UI problem, it is an execution-policy problem: what is allowed to run, with which credentials, with what confirmation, and what a phone-sized confirmation dialog can honestly convey about a command it is about to execute on a machine the user is not sitting at. That deserves its own designed slice, not an incremental commit.

One thing worth planning for while the model is still malleable: the fleet load fans out to every configured machine on every refresh, each doing a discovery plus two list calls. With three or four machines, one of them asleep, that is a slow screen governed by the slowest timeout. The model should probably expose per-machine loading state and let rows arrive independently rather than resolving one `Promise.all` — otherwise the fleet view is as slow as its worst machine, and the offline one is always the worst.

Note on the test script: `test:profiles` now chains `fleet-model.test.mjs` behind `&&`. It does run in CI, but a suite whose name gives no hint it exists is easy to lose. A `test:fleet` script added to the `pr-checks.yml` list would cost one line and stay findable.

PR #170, opened 2026-08-14
feat(tasks): expose backend-neutral Git review summary

giuliastro:
Superseded by #172, which carries the backend-neutral task diff/review summary together with the task-first client and integration fixes.

Closing this PR so review and eventual merge happen through one consolidated path only.

### DIFF
--- a/bridge/src/task-review.js
+++ b/bridge/src/task-review.js
@@ -1,0 +1,51 @@
+function text(result) {
+  return String(result?.stdout ?? "")
+}
+
+function count(value) {
+  if (value === "-") return null
+  const parsed = Number.parseInt(value, 10)
+  return Number.isFinite(parsed) ? parsed : null
+}
+
+export function parseNumstat(output) {
+  return output.split(/\r?\n/).filter(Boolean).flatMap((line) => {
+    const [additions, deletions, ...pathParts] = line.split("\t")
+    const path = pathParts.join("\t")
+    if (!path) return []
+    return [{ path, additions: count(additions), deletions: count(deletions), untracked: false }]
+  })
+}
+
+export async function inspectTaskDiff(workspace, worktreeManager) {
+  const status = await worktreeManager.inspect(workspace)
+  const sourceHead = text(await worktreeManager.runGit(["-C", workspace.source, "rev-parse", "HEAD"])).trim()
+  const tracked = parseNumstat(text(await worktreeManager.runGit([
+    "-C", workspace.path, "diff", "--numstat", sourceHead, "--"
+  ])))
+  const trackedPaths = new Set(tracked.map((file) => file.path))
+  const untracked = text(await worktreeManager.runGit([
+    "-C", workspace.path, "ls-files", "--others", "--exclude-standard"
+  ])).split(/\r?\n/).filter(Boolean).filter((path) => !trackedPaths.has(path)).map((path) => ({
+    path,
+    additions: null,
+    deletions: null,
+    untracked: true
+  }))
+  const files = [...tracked, ...untracked].sort((left, right) => left.path.localeCompare(right.path))
+  const knownAdditions = files.reduce((total, file) => total + (file.additions ?? 0), 0)
+  const knownDeletions = files.reduce((total, file) => total + (file.deletions ?? 0), 0)
+
+  return {
+    managed: status.managed,
+    source: workspace.source,
+    sourceHead,
+    branch: workspace.branch,
+    dirty: status.dirty,
+    fileCount: files.length,
+    additions: knownAdditions,
+    deletions: knownDeletions,
+    hasUnknownLineCounts: files.some((file) => file.additions === null || file.deletions === null),
+    files
+  }
+}
--- a/bridge/src/task-review.js
+++ b/bridge/src/task-review.js
@@ -8,10 +8,18 @@ function count(value) {
   return Number.isFinite(parsed) ? parsed : null
 }
 
+function records(output) {
+  return output.includes("\0") ? output.split("\0").filter(Boolean) : output.split(/\r?\n/).filter(Boolean)
+}
+
 export function parseNumstat(output) {
-  return output.split(/\r?\n/).filter(Boolean).flatMap((line) => {
-    const [additions, deletions, ...pathParts] = line.split("\t")
-    const path = pathParts.join("\t")
+  return records(output).flatMap((record) => {
+    const firstTab = record.indexOf("\t")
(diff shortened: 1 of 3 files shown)


bridge/src/task-review.js
```diff
@@ -0,0 +1,51 @@
...
+    return [{ path, additions: count(additions), deletions: count(deletions), untracked: false }]
+  })
+}
+
+export async function inspectTaskDiff(workspace, worktreeManager) {
+  const status = await worktreeManager.inspect(workspace)
+  const sourceHead = text(await worktreeManager.runGit(["-C", workspace.source, "rev-parse", "HEAD"])).trim()
+  const tracked = parseNumstat(text(await worktreeManager.runGit([
+    "-C", workspace.path, "diff", "--numstat", sourceHead, "--"
+  ])))
```
giuliastro, line 33: **A renamed file comes back as one entry whose `path` is not a path.**

`diff.renames` has defaulted to true since Git 2.9, so `git diff --numstat` emits rename pairs as a single composite line. I ran it against a scratch repo — rename `old.txt` → `new.txt`, then append a line:

```
$ git diff --numstat $BASE --
1	0	old.txt => new.txt
```

That parses into `{ path: "old.txt => new.txt", additions: 1, deletions: 0 }`. Two consequences, both reaching the client:

- the file list shows a string no client can resolve — it cannot be opened, fetched or matched against anything;
- `trackedPaths` then holds that composite, so the untracked dedup at line 27 stops working for either side of the rename;
- and the counts describe the *rename-adjusted* delta (`1/0`) rather than the work done, which reads as a one-line change for what may be a wholesale move.

`--no-renames` gives honest, resolvable entries for the same commit:

```
$ git diff --no-renames --numstat $BASE --
4	0	new.txt
0	3	old.txt
```

That is the right default for a review payload: a rename genuinely is a delete plus an add as far as "what does this task's work consist of" is concerned, and every path is real. If rename pairing is wanted later it should be added deliberately, with `-z` so the two paths arrive as separate fields instead of inside one string.

---

bridge/src/task-review.js
```diff
@@ -0,0 +1,51 @@
...
+  const trackedPaths = new Set(tracked.map((file) => file.path))
+  const untracked = text(await worktreeManager.runGit([
+    "-C", workspace.path, "ls-files", "--others", "--exclude-standard"
+  ])).split(/\r?\n/).filter(Boolean).filter((path) => !trackedPaths.has(path)).map((path) => ({
```
giuliastro: **Any non-ASCII filename arrives C-quoted, escapes and surrounding quotes included.**

`core.quotePath` defaults to true, and `WorktreeManager.runGit` passes no `-c` overrides. Same scratch repo, an untracked file named `perché.md`:

```
$ git ls-files --others --exclude-standard
"perch\303\251.md"
```

That literal 18-character string — quotes included — is what lands in `files[].path` and goes out over `/v1/tasks/:id/diff`. For an Italian, French, German, Spanish or Chinese repository this is not an edge case, it is most of the file names. The same applies to the numstat paths on line 24.

The fix is one flag on both calls:

```
$ git -c core.quotePath=false ls-files --others --exclude-standard
perché.md
```

Better still, set it once in `WorktreeManager.runGit` so every current and future git call in the daemon is UTF-8 clean — `inspect()` only counts porcelain lines today, so nothing else is affected, but the next command that reads a path will hit exactly this.

Two related notes while you are in here:

- A path containing a literal tab stays quoted even with `quotePath=false`, and `parseNumstat` splits on `\t`. The `pathParts.join("\t")` on line 15 shows the case was anticipated, but a quoted path is still not the real one. `-z` on both commands solves quoting and separators together and is the durable answer; if you take `-z`, the parser splits on `\0` and the `join` goes away.
- Neither case has a test. A rename and an accented filename would be two short additions to `task-review.test.js`, and they are the two that will actually be hit.

---

giuliastro (commented):
## Review of #170 — Git review summary

Read `task-review.js`, the finish-server routing change and the new tests, then reproduced the git behaviour in a scratch repository rather than reasoning about it from memory.

**Verdict: correct shape, correct scope, two real bugs in the path handling — both proven, both one flag to fix.**

### What is right

- **Read-only, and structurally so.** No `worktreeManager.cleanup`, no store write, no state change; the route is a pure projection. For the first piece of the review loop that is the correct place to start.
- **It slots into the existing server without disturbing it.** One route constant, one `inspectDiff` flag folded into the existing guard, and `(resultMatch ?? diffMatch ?? finishMatch)` extended in the obvious way. The reconciliation await and the authentication call are inherited rather than re-implemented — no fourth copy of `authenticate`.
- **The non-worktree branch returns the same shape with zeroes** instead of a 404 or a special case, so a client renders one component for both. That consistency was the right instinct in #164 and is right again here.
- **Binary handling is honest.** `count("-") → null` plus `hasUnknownLineCounts` says "this file changed, the line count is not a meaningful number" rather than silently reporting 0 and understating the diff.
- **The diff is taken in the worktree against the source head**, so uncommitted work is included. For "what has this task actually done" that is the right base — a client would otherwise show nothing until the agent commits.
- **`--exclude-standard` and the `trackedPaths` dedup** mean the untracked list is what a person would call untracked, not everything gitignored.
- **Allowing the route while the task is still running**, unlike `/finish`, is a deliberate and correct asymmetry: looking is safe, finalizing is not. The test names it as a point-in-time view.

### The two bugs

1. **Renames produce a composite non-path** (comment on line 22): `git diff --numstat` emits `old.txt => new.txt` as a single field, which is unresolvable by any client, breaks the untracked dedup, and reports rename-adjusted counts. `--no-renames` gives two honest entries.
2. **Non-ASCII filenames arrive C-quoted** (comment on line 26): `core.quotePath` defaults to true, so `perché.md` goes out as `"perch\303\251.md"`. On an Italian or Chinese repository that is not an edge case, it is most of the file list. `-c core.quotePath=false`, ideally set once in `runGit`.

Both are invisible in the current tests because every fixture uses ASCII names and no rename.

### Where this leaves the loop

`diff` is the honest first slice of `run → diff → tests → review → PR`, and stopping here was right — the next two steps are where it stops being a read. Running a project's tests means executing arbitrary code from a repository on a machine the user is not sitting at, and opening a PR means acting under their identity on an external service. Both need an explicit policy about what may run, under whose credentials, and what a phone-sized confirmation can honestly convey about a command it is about to execute somewhere else. Worth designing before either is built, and this PR is the right thing to have shipped first.

Once the two path fixes land I would merge this ahead of #167/#168 — it is independent, it is green, and nothing else depends on it.

PR #172, opened 2026-08-14
feat: consolidate task-first control plane, fleet foundation and review summary

giuliastro:
## Summary

Consolidates the task-first client, fleet client foundation, backend-neutral Git review summary, and the integration fixes found during Android manual testing into one reviewable PR.

This supersedes #167, #168, #170 and the temporary integration PR #171. It is intentionally **draft** and must not be merged until Android manual testing is complete.

## Task-first workflow

- expose New Task alongside the existing New Session workflow;
- discover machine projects and available agents;
- create tasks, prepare isolated Git worktrees by default, and launch the selected agent;
- keep task execution represented by the existing session UI after launch;
- surface task, project, worktree and launch errors in the dialog.

## OpenCode / machine-daemon integration

- allow `harness-remote-daemon --backend opencode` without requiring an ACP backend;
- keep the managed OpenCode process internal while exposing one public machine-daemon endpoint;
- resolve the machine-level endpoint separately from an existing OpenCode session endpoint: a saved OpenCode profile on 4096 can discover the standard Harness machine daemon on 4097, while a profile already on 4097 stays there;
- never treat ordinary OpenCode connectivity as proof that the machine daemon exists;
- normalize native/Capacitor task payloads, including wrapped and stringified JSON responses;
- provide endpoint-oriented errors instead of misleading `invalid JSON` errors when a machine API is not available.

## UI cleanup

- Task dialog now uses the same shared wizard/footer action layout as New Session;
- removed one-off button widths and inline action sizing;
- project/agent fields use a responsive grid;
- worktree isolation is visually separated without introducing a separate design language;
- on narrow screens Refresh uses the compact icon slot while New Task and New Session remain labeled, balanced actions.

## Fleet foundation

- group saved profiles by physical daemon endpoint;
- preserve stable machine identity when reachable and endpoint identity when offline;
- isolate failures across machines;
- aggregate agents, projects and tasks per machine;
- namespace task identity at the fleet layer;
- support explicit task placement on a selected machine.

No Fleet UI is introduced in this slice.

## Review summary

- add read-only `GET /v1/tasks/:id/diff` inspection for managed worktrees;
- compare task work against the source checkout HEAD;
- support committed, tracked, untracked, binary, Unicode and tab-containing paths;
- use rename-safe, NUL-delimited Git parsing;
- expose aggregate file/addition/deletion counts without mutating task state.

## Validation

PR checks now explicitly include a `test:tasks` suite covering the machine-endpoint resolution, Android/native payload normalization and Task dialog UI invariants. The consolidated source state passed:

- TypeScript build and web regression suites, including `test:tasks`;
- bridge tests on Linux;
- bridge tests on macOS;
- bridge tests on Windows;
- Android debug APK build and signature verification.

Manual Android verification is still required before merge.


giuliastro:
Current gate status after re-review against #181/#173:

**Do not enable `TASK_LAUNCH_ENABLED` yet. Do not merge this PR yet.**

1. #181 is now narrowed to the bridge prompt-acknowledgement path: acknowledgements are stored one-per-session, so queuing prompt B overwrites prompt A's ack. A late ACP echo of A can then become a real user chunk and, especially when the failed turn emits no assistant chunk, the echo of B can append to the same user chunk. This matches the reported concatenation/title corruption. The fix needs a regression test covering queued prompts + first-turn failure/no assistant echo.

2. #173 is still incomplete for the stated requirement. The branch adds session-less model lookup and a PI-specific loader (`pi --list-models`), which is good for PI. But OMP, Claude Code and Codex still have no guaranteed pre-session catalog on a fresh agent with zero sessions; they can only reuse a catalog learned from an ACP session. Therefore New Task cannot yet guarantee model choice at task start for every supported harness.

Required before this draft can be enabled/merged:
- fix #181 and add the queue/echo regression;
- define a truthful pre-session model-discovery contract per agent (real session-less catalog where supported, explicit unsupported/default-only state otherwise), with zero-session coverage;
- then enable `TASK_LAUNCH_ENABLED`;
- rerun the full required matrix: TypeScript/build + all 12 web suites, bridge Linux/macOS/Windows, debug APK build;
- preserve the legacy URL invariant for profiles without `agentId` (`web/src/server-config.test.mjs`).

No merge performed.

giuliastro:
Model-selection implementation has been replaced with the agent-level design agreed in review:

- removed the PI-only `pi --list-models` path;
- `GET /v1/agents/:id/models` refreshes the selected agent when New Task opens;
- managed HTTP agents (OpenCode) are queried directly on every refresh;
- ACP agents use one dedicated, prompt-less catalog session per agent, persisted and reused across New Task opens and daemon restarts (not create/destroy per task);
- the catalog session is kept out of the user session list;
- cached models are fallback-only and are marked stale when refresh fails;
- POST task launch performs a second fresh validation; if the chosen model disappeared, launch returns a conflict instead of silently falling back;
- regression tests cover zero-user-session bootstrap, reuse across refresh/restart, model removal, HTTP refresh, and launch gating.

`TASK_LAUNCH_ENABLED` remains false. Current head is `b1bd400`; full PR checks are running. The first run exposed an expected wrapper-wiring test that still assumed the pre-catalog server chain; that test has been updated and macOS bridge tests are now green on the new head. No merge until review + complete matrix.

giuliastro:
## Manual Android status — NOT READY

The 2026-08-14 real-device test found several P0 failures in the task/connection path. This PR must remain draft and `TASK_LAUNCH_ENABLED` must remain false in normal builds.

Confirmed failures:

- New Task model refresh can block for a very long time then fail with `Cannot reach localhost:4097. timeout`;
- reopening after the timeout can show cached models immediately, but freshness/completeness is unclear;
- Start returns to the session list before the newly created task/session is visible; it appears only many seconds later;
- opening the launched task/session can show the model area in warning/yellow state with no models loaded;
- machine setup is still backend/profile-first in places: first-run OpenCode flow and later PI flow are inconsistent;
- a daemon-backed machine still effectively requires duplicated per-agent client configuration;
- generated daemon credentials must not change on ordinary restart.

Roadmap/issue cleanup made from these findings:

- #143 reopened as P0 Universal Daemon/auth/discovery stabilization;
- #178 rewritten around one machine-first wizard;
- #180 rewritten around one saved machine with discovered agents underneath;
- #145 rewritten as an end-to-end task reliability gate;
- #173 rewritten around bounded/fresh model catalog + persisted run configuration;
- #133 now has an explicit P0 stabilization gate;
- #182 abandoned/closed; #181 remains isolated.

Do not add more feature surface to #172 until the above contract is stable. The next code work should reduce complexity and make the existing path deterministic rather than expand it.

giuliastro:
Archiving this Harness 3 work while `main` is restored to the last known-good pre-Harness-3 baseline (`923d62a`). The code remains preserved on `agent/task-control-plane-consolidated` and the full archive/index lives on `archive/harness-3-2026-08-15` in `docs/HARNESS_3_ARCHIVE_2026-08-15.md`. Do not merge this PR as-is into restored `main`; future Harness 3 work should be reintroduced in small, independently validated slices.

Issue #181, opened 2026-08-14
Two separate prompts render as one merged message, and the merge becomes the session title


giuliastro:
Reported from the Android build against PI through the machine daemon.

Two prompts were sent, `ciao` and then `ciai`. PI's own terminal shows them as two turns — the first answered with a 401 `CreditsError`, the second answered normally — and PI's reasoning for the second turn reads:

> The user said "ciao" and "ciai" - just greetings in Italian.

so the agent received two distinct prompts. The app shows **one** user bubble reading `ciaociai`, and that concatenation also became the session title.

## Where it is not happening

`AcpService.#recordPrompt` creates a new message per prompt, with its own `randomUUID()` and its own `time.created`:

```js
messages.push({ info: { id: messageID, role: "user", sessionID, time: { created: Date.now() } }, parts: [...] })
```

The queue keeps entries separate too — each carries its own `messageID` — and the title is derived from the *first* user message, not from a running concatenation. So the daemon holds two messages and the merge happens after it, either in what the client reconciles or in what it renders.

The most suspicious neighbours, not yet confirmed as the cause:

- `hasMatchingUserMessage` in `App.tsx` reconciles an optimistic user message against the server list by **exact text equality**, which is already fragile: two identical prompts in one session are indistinguishable.
- `groupRenderedMessages` merges consecutive non-user messages into one run. User messages are documented as always starting a fresh group, so this should not apply — worth confirming rather than assuming.

## Why the title matters as much as the bubble

The title is derived once from the first user message and cached. If the first message is already a merge, the session is permanently named after a string the user never typed, and renaming is the only escape.

## Reproduction

1. Connect to a machine daemon with an ACP agent.
2. Send a prompt that fails at the agent (an expired key or exhausted credits produces a 401 like the one above).
3. Send a second, different prompt.
4. The transcript shows one bubble containing both texts concatenated, and the session is titled with that concatenation.

The failed first turn may or may not be necessary; that is the sequence that produced it.

giuliastro:
[...]

PR #182, opened 2026-08-14
fix: preserve PI message boundaries from native history

giuliastro:
## Status

This draft is **abandoned** after real-device testing.

The native PI-history loader did split the original `ciao` / `ciai` pair, but it also widened PI session behavior and produced duplicate/corrupted transcript state on Android. The branch was reset to `main` rather than layering more reconciliation work on top of it.

#181 remains open and will be fixed separately with a smaller change in the ACP replay/reconciliation/message-boundary path, with regression coverage and real-device validation.

Do not merge this PR.

### DIFF
--- a/bridge/src/harness-profiles.js
+++ b/bridge/src/harness-profiles.js
@@ -1,3 +1,4 @@
+import { findExecutable } from "./launcher.js"
 import { createCodexHistoryLoader } from "./codex-session-history.js"
 import { createOmpHistoryLoader } from "./omp-session-history.js"
 import { OMP_EXTENSION_ACTION_PROVIDERS } from "./extension-actions.js"
@@ -44,6 +45,7 @@ export const HARNESS_PROFILES = {
     // default failed with `notarget` when a release outran its own tarball in the registry.
     command: process.platform === "win32" ? "npx.cmd" : "npx",
     args: ["-y", "@automatalabs/pi-acp@0.2.5"],
+    adapterCommand: "pi-acp",
     permissionMode: "allow",
     preserveListedTimestamps: true,
     reloadOnHistoryRefresh: false,
@@ -68,6 +70,7 @@ export const HARNESS_PROFILES = {
     // Pinned to avoid the `notarget` scenario that PI hit: an unpinned default failed when a
     // release appeared in the registry index before its tarball could be fetched.
     args: ["-y", "@agentclientprotocol/claude-agent-acp@0.63.0"],
+    adapterCommand: "claude-agent-acp",
     permissionMode: "allow",
     preserveListedTimestamps: true,
     reloadOnHistoryRefresh: false,
@@ -94,6 +97,7 @@ export const HARNESS_PROFILES = {
     // Pinned to avoid the `notarget` scenario that PI hit: an unpinned default failed when a
     // release appeared in the registry index before its tarball could be fetched.
     args: ["-y", "@agentclientprotocol/codex-acp@1.1.14"],
+    adapterCommand: "codex-acp",
     permissionMode: "allow",
     // The adapter offers `api-key` before `chat-gpt`; the former demands CODEX_API_KEY or
     // OPENAI_API_KEY, while a `codex login` leaves ChatGPT credentials the `chat-gpt` method
@@ -127,4 +131,20 @@ export function harnessProfile(id) {
   const profile = HARNESS_PROFILES[id]
   if (!profile) throw new Error(`Unsupported backend: ${id}`)
   return profile
+}
+
+/**
+ * The harness and its ACP adapter are two different installations. `pi` from the project's own
+ * installer puts the harness on PATH and no adapter with it, so detecting the harness and then
+ * assuming `npx` can fetch an adapter is how a machine with PI installed ends up unable to run PI.
+ *
+ * An adapter already on PATH is preferred over fetching one: it is what the user installed, it
+ * starts without a network round trip, and it sidesteps environments where `npx` cannot link a
+ * binary — which is exactly what happens under proot on Android.
+ */
+export function resolveAcpLaunch(profile, { find = findExecutable } = {}) {
+  if (!profile.adapterCommand) return { command: profile.command, args: [...profile.args], source: "harness" }
+  const installed = find(profile.adapterCommand)
+  if (installed) return { command: installed, args: [], source: "path" }
+  return { command: profile.command, args: [...profile.args], source: "npx" }
 }
--- a/bridge/src/acp-client.js
+++ b/bridge/src/acp-client.js
@@ -34,6 +34,7 @@ export class AcpClient extends EventEmitter {
   #agentInfo
   #promptCapabilities = {}
   #stderr = ""
+  #stderrPartial = ""
 
   constructor({ command = "omp", args = ["acp"], permissionMode = "deny", preferredAuthMethod, spawnProcess = spawn } = {}) {
     super()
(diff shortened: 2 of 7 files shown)


bridge/src/harness-profiles.js
```diff
@@ -47,6 +48,9 @@ export const HARNESS_PROFILES = {
     args: ["-y", "@automatalabs/pi-acp@0.2.5"],
     adapterCommand: "pi-acp",
     permissionMode: "allow",
+    // PI's native JSONL keeps exact message records and branch parentage. Prefer it for history
+    // instead of inferring message boundaries from optional ACP messageIds during session/load.
+    historyLoader: createPiHistoryLoader(),
```
giuliastro: **Adding `historyLoader` to PI changes far more than history reconstruction, and none of it is covered here.**

`historyLoader` is not a passive source of messages. `acp-service.js` branches on `this.#historyLoader && !this.#ownedSessions.has(sessionID)` in seven places — lines 238, 276, 305, 314, 495, 568 — and each is a behaviour switch:

- line 238 and 276 mark sessions **external** in what is reported to the client;
- line 314 makes `todos()` return `[]` outright for any session this process did not create;
- lines 495 and 568 change the prompt and load paths.

Until this PR, PI had no loader and none of those branches fired. After it, every PI session the daemon did not itself create takes the external path. That is not a subset of "message boundaries are preserved" — it is a different mode of operation for the harness the user is actually running today.

The two tests here exercise the JSONL parser directly and never construct an `AcpService`, so the entire question of what the loader does to a live PI session is untested. Given that #181 was found by using PI on a phone, and the fix ships to the same person, the regression surface deserves at least one test that mounts the loader on a service and checks that an owned session still reports todos, still prompts, and is not marked external.

Worth stating plainly in the PR body too: this changes how PI sessions are represented, not just how their history is read.

---

bridge/src/pi-session-history.js
```diff
@@ -0,0 +1,147 @@
...
+      // Fall back to the authoritative header id before giving up.
+      for (const file of files) {
+        const header = await sessionHeader(file)
+        if (header?.type === "session" && header.id === sessionID) {
+          sessionFiles.set(sessionID, file)
+          return file
+        }
+      }
+      return undefined
+    } catch (error) {
```
giuliastro: **A lookup that fails rescans the whole tree and opens every file, and the failure is never cached.**

`sessionFiles.set` runs only on success. So for any session id this loader cannot resolve — an OpenCode session, an ACP session with no native file, a stale id — every call does:

1. `readdir(sessionRoot, { recursive: true })` over the entire PI session tree;
2. if no filename matches, `sessionHeader()` on **every** `.jsonl` found, each an `open` + 8 KiB read + `JSON.parse` + `close`.

A user with a few hundred PI sessions therefore pays a few hundred file opens per history request, repeatedly, and the daemon is being run on a phone under proot where that is not cheap. The happy path caches; the path that costs the most does not.

Two changes make it bounded:

- cache the negative result too, so an id that has no native file is resolved once (a `Map<string, string | null>` covers both);
- cache the directory listing with a short lifetime rather than re-walking it per lookup, since it is the same answer for every id resolved in the same moment.

While you are here: the cache never expires either, so a session file that is rotated or moved keeps resolving to a path that no longer exists. `loadPiHistory` would then throw out of `createReadStream` rather than returning `[]` like the other failure paths do.

---

bridge/src/pi-session-history.js
```diff
@@ -0,0 +1,147 @@
...
+          role,
+          sessionID,
+          time: { created: Number.isFinite(created) ? created : Date.now() }
+        },
+        parts
+      })
+    }
+    return messages
+  }
+}
```
giuliastro: Two silent losses in the final loop.

**`if (parts.length === 0) continue` drops whole turns.** `messageParts` only recognises `text`, `thinking` and `image`. An assistant turn that produced only tool calls — the ordinary shape of a coding agent doing work — yields no parts and vanishes from the transcript entirely. The reconstructed history then reads as if the agent answered nothing between two user turns, which is a different corruption from the one being fixed but the same kind: a message that existed does not appear.

Either keep the envelope with no parts and let the renderer decide, or map the unrecognised content types to something visible. Dropping is the one option that cannot be noticed downstream.

**`Date.now()` as the timestamp fallback.** A record with a missing or unparseable `timestamp` is stamped with the moment it happened to be read, so on replay a set of them all collapse to the same instant and sort after everything real. PI's profile sets `preserveListedTimestamps: true`, which says the timestamps in the listing are meant to be trusted — a synthesised one quietly violates that. Inheriting the previous message's time, or leaving it absent for the caller to handle, both fail more honestly than a value that looks real.

---

giuliastro (commented):
## Review of #182 — PI message boundaries from native history

Head `0168088`, 212 additions across three files, bridge only. Read the loader in full, the profile change, both tests, and the seven call sites in `acp-service.js` that branch on the presence of a history loader.

**Verdict: the parser is careful and well-reasoned, but I am not convinced it fixes the reported bug, and it changes how PI sessions are handled in ways nothing here tests.**

### What is well done

- **Reconstructing the parent chain instead of flattening the file** is the right call, and the comment explaining why the last appended entry is the only branch signal available is honest about its own limit.
- **The session id is regex-guarded** before it is used, even though it only ever reaches an `endsWith` comparison rather than a path join. Defensive in the right direction.
- **`sessionHeader` reads a bounded 8 KiB and closes in `finally`**, and every parse failure degrades to skipping the line rather than failing the load.
- The branched-history test is the one I would have asked for.

### The reason to hold

**This fixes replayed history, and #181 was reported live.** The bug was two prompts sent one after another, with a failed turn between them, rendering as one bubble. `AcpService.#recordPrompt` creates a separate message per prompt with its own id — I checked that while filing #181 — so on the live path the daemon already holds two messages and the merge happened somewhere after it. This PR changes what `session/load` reconstructs.

That may well be the same path, if the client reloads history between turns. But the PR asserts the cause rather than demonstrating it, and the tests are unit tests of the JSONL parser, so nothing here shows the reported sequence producing one bubble before and two after. Without that, "Closes #181" is a hypothesis.

The cheapest way to settle it: reproduce on the device — two prompts, failed turn between — and confirm whether the merged bubble appears before any history reload. If it does, the client is still merging and this PR does not close the issue; if it only appears after a reload, this is the right fix and the PR body should say so.

### Also before merge

1. **The profile change is not a small one** (comment on `harness-profiles.js:53`). Giving PI a `historyLoader` flips seven `!ownedSessions.has(...)` branches in `AcpService`, including one that makes `todos()` return empty. PI is the harness in daily use right now; none of that is covered.
2. **The lookup rescans and reopens everything on a miss, and never caches the miss** (comment on `pi-session-history.js:61`). On a phone with a few hundred sessions this is hundreds of file opens per request, repeatedly.
3. **Two silent losses** (comment on `pi-session-history.js:128`): an assistant turn made only of tool calls is dropped entirely, and a missing timestamp becomes `Date.now()` on a profile that declares `preserveListedTimestamps`.

### One question I cannot answer from here

`PI_CODING_AGENT_DIR` and the `~/.pi/agent/sessions` layout are taken as given. I have no PI installation to check them against, and if either is wrong the loader silently returns `[]` and PI history quietly stops working — the same failure shape as "no native file", so it would not look like a bug. Worth confirming against a real installation and noting in the PR how it was confirmed.

Scoping the change to the bridge, and keeping it draft until the full matrix has run, are both right.

---

giuliastro:
Real-device result on the combined #172 + #182 test build (PI): the native-history change fixes the first half of #181 — the original `ciao` / `ciai` prompts are now rendered as two separate user messages. However the session is still corrupted afterwards. After the assistant response, the app temporarily appends another `ciaociai`, then another copy of the previous assistant response, then another `ciaociai`; after a few seconds the duplicate assistant response disappears, but an extra trailing `ciaociai` remains.

So #182 is **not sufficient to close #181**. This strongly suggests we now have two sources being reconciled: the corrected native PI history plus ACP/live/snapshot replay, with semantic dedup failing for the concatenated replayed user chunk. The fact that one duplicate assistant response disappears after a later refresh while the extra `ciaociai` survives is especially consistent with merge/dedup/reconciliation rather than PI persistence itself.

Next fix should target the AcpService/app merge path so that when native history is authoritative for an external PI session, replay/snapshot messages cannot re-append semantically equivalent history under different IDs — and specifically must not preserve the old synthetic `ciaociai` chunk after native `ciao` + `ciai` have been loaded. Add an integration regression reproducing this exact sequence before marking #181 closed.

PR #184, opened 2026-08-14
fix: make daemon credentials explicit or ephemeral

giuliastro:
## Scope

Small first stabilization slice for #143.

The credential policy is now intentionally simple:

- explicit `--username` + `--password` are used exactly as supplied and stay stable because the user chose them;
- environment credentials behave the same way;
- if credentials are omitted, the daemon generates a fresh random password for that process, prints it, and does **not** persist it;
- generated passwords therefore rotate on restart by design;
- no generated client password is stored on disk.

This PR also keeps the npm-bin direct-invocation fix so `harness-remote-daemon` actually runs when npm exposes it through a symlink, and prints the active username/password clearly on startup.

## Compatibility

No profile URL semantics, task behavior, session behavior or legacy direct-server behavior is changed. The launcher keeps its existing random-credential behavior.

## Validation

Regression coverage checks:
- ephemeral credentials are regenerated;
- explicit environment credentials are preserved;
- explicit CLI credentials are left untouched;
- npm-bin symlink invocation is recognized.

## Follow-up within #143

This does **not** claim #143 complete. The next slice is all-agent discovery/routing + lifecycle/reason reporting.

Do not merge until CI and manual validation pass.


giuliastro:
Manual real-device/Termux validation: PASS for credential persistence. The daemon was started, stopped, and started again from the same state directory; the generated password remained unchanged across restart. Keeping this PR draft until CI completes and the remaining #143 daemon-discovery/lifecycle work is handled separately.

giuliastro:
Archived with the Harness 3 work. The code remains on `agent/stabilize-machine-daemon-p0`; the archive/index is `archive/harness-3-2026-08-15:docs/HARNESS_3_ARCHIVE_2026-08-15.md`. Closing without merge while `main` is restored to the last known-good pre-Harness-3 baseline.

PR #185, opened 2026-08-14
fix: make New Task model discovery fast and non-blocking

giuliastro:
## Scope

Focused #145 stabilization slice on top of #172.

This addresses the real-device failures where New Task waited a very long time for the model list, then failed the whole dialog, and a later reopen sometimes showed a cache with unclear freshness.

### Changes
- render the New Task form as soon as machine + projects are known;
- move model discovery into a separate background state: `loading | fresh | stale | unavailable`;
- cap client model refresh requests at 5 seconds on Android/browser/desktop;
- cap daemon ACP/HTTP catalog refresh work at 5 seconds;
- **reuse the same ACP process used by ordinary PI sessions instead of starting a second model-only ACP process**;
- warm that shared ACP process in the background after the daemon starts listening, so model discovery follows the same PI session/configOptions path that already worked before task work;
- keep one hidden reusable catalog session only to obtain pre-task `configOptions`, without changing normal session model behavior;
- stale model catalogs remain visibly marked;
- unavailable model discovery no longer replaces the entire task dialog;
- explicit Refresh control retries only the model catalog;
- disable Start while a model-capable agent has no trustworthy catalog;
- keep the dialog visible while worktree/session launch is still running;
- keep final launch-time model revalidation/no-fallback behavior unchanged.

### Real-device finding
The first Termux test proved the new endpoint was reachable but the previous implementation timed out at `Starting pi model catalog timed out after 5 seconds`. That exposed the unnecessary second ACP startup. This revision removes that duplicate PI ACP path and deliberately reuses the mature ordinary-session path instead.

### Deliberately not included
- no machine-first/profile UX redesign;
- no PI replay changes;
- no default enabling of New Task;
- no merge into #172 until manual Android validation.

Targets `agent/task-control-plane-consolidated` so this PR stays reviewable as a focused stabilization diff.

### DIFF
--- a/bridge/src/agent-model-catalog.js
+++ b/bridge/src/agent-model-catalog.js
@@ -1,6 +1,8 @@
 import { mkdir, readFile, writeFile } from "node:fs/promises"
 import path from "node:path"
 
+export const MODEL_CATALOG_TIMEOUT_MS = 5_000
+
 function splitModelValue(value, fallbackProviderID) {
   const separator = value.indexOf("/")
   return separator > 0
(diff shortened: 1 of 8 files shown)


giuliastro:
Archived with the Harness 3 work. The code remains preserved on `agent/stabilize-new-task-models` and is indexed in `archive/harness-3-2026-08-15:docs/HARNESS_3_ARCHIVE_2026-08-15.md`. Closing without merge; any future model/task work should restart from restored stable `main`.

PR #192, opened 2026-08-15
fix(bridge): surface ACP provider error messages

giuliastro:
## Summary

Surface actionable ACP provider errors that adapters place in `error.data.message` instead of dropping them behind a generic `Internal error`.

PI currently returns rate limits as:

```json
{"message":"Internal error","data":{"errorKind":"rate_limit","message":"provider rate limit"}}
```

The bridge already preserves `data.details` for adapters such as Codex; this extends the same behavior to `data.message`, so the existing `session.error` path can show the real provider failure in the app.

## Scope

- one bridge-only error formatting change
- no session/model/runtime behavior changes
- no changes to `main` beyond this PR

## Expected UI result

Instead of silently ending / exposing only a generic error, the session receives an actionable error such as:

`Internal error: provider rate limit`

This PR is intentionally narrow and is not merged automatically.

### DIFF
--- a/bridge/src/acp-client.js
+++ b/bridge/src/acp-client.js
@@ -9,14 +9,13 @@ const REQUEST_TIMEOUT_MS = 30_000
 const STDERR_KEPT_CHARS = 600
 
 /**
- * JSON-RPC only promises a human-readable `message`, and Codex spends it on a bare "Internal
- * error", putting the part worth reading — "thread <id> already has an active writer" — in
- * `data.details`. Dropping that left the app showing `{"error":"Internal error"}` for a refusal it
- * could otherwise have explained.
+ * JSON-RPC only promises a human-readable `message`, and adapters sometimes spend it on a bare
+ * "Internal error", putting the part worth reading in `data.details` or `data.message`. Dropping
+ * that left the app unable to explain actionable failures such as provider rate limits.
  */
 function acpErrorMessage(error) {
   const message = error?.message ?? "ACP adapter request failed"
-  const details = error?.data?.details
+  const details = error?.data?.details ?? error?.data?.message
   return typeof details === "string" && details && !message.includes(details) ? `${message}: ${details}` : message
 }
 
--- a/bridge/src/acp-client.js
+++ b/bridge/src/acp-client.js
@@ -15,8 +15,10 @@ const STDERR_KEPT_CHARS = 600
  */
 function acpErrorMessage(error) {
   const message = error?.message ?? "ACP adapter request failed"
-  const details = error?.data?.details ?? error?.data?.message
-  return typeof details === "string" && details && !message.includes(details) ? `${message}: ${details}` : message
+  const details = [error?.data?.details, error?.data?.message].find(
+    (value) => typeof value === "string" && value
+  )
+  return details && !message.includes(details) ? `${message}: ${details}` : message
 }
 
 export class AcpClient extends EventEmitter {
--- a/bridge/test/acp-provider-error.test.js
+++ b/bridge/test/acp-provider-error.test.js
@@ -1,0 +1,57 @@
+import assert from "node:assert/strict"
+import { EventEmitter } from "node:events"
+import test from "node:test"
+import { AcpClient } from "../src/acp-client.js"
+
+class FakeChild extends EventEmitter {
+  killed = false
+  stdout = new EventEmitter()
+  stderr = new EventEmitter()
+
+  constructor() {
+    super()
+    this.stdout.setEncoding = () => undefined
+    this.stderr.setEncoding = () => undefined
+    this.stdin = {
+      writable: true,
+      write: (line, callback) => {
+        const request = JSON.parse(line)
+        if (request.method === "initialize") {
+          this.respond({ jsonrpc: "2.0", id: request.id, result: { agentInfo: { name: "pi" }, authMethods: [] } })
+        } else if (request.method === "session/prompt") {
+          this.respond({
+            jsonrpc: "2.0",
+            id: request.id,
+            error: {
+              code: -32603,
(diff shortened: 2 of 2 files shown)


bridge/src/acp-client.js
```diff
@@ -9,14 +9,13 @@ const REQUEST_TIMEOUT_MS = 30_000
...
- * `data.details`. Dropping that left the app showing `{"error":"Internal error"}` for a refusal it
- * could otherwise have explained.
+ * JSON-RPC only promises a human-readable `message`, and adapters sometimes spend it on a bare
+ * "Internal error", putting the part worth reading in `data.details` or `data.message`. Dropping
+ * that left the app unable to explain actionable failures such as provider rate limits.
  */
 function acpErrorMessage(error) {
   const message = error?.message ?? "ACP adapter request failed"
-  const details = error?.data?.details
+  const details = error?.data?.details ?? error?.data?.message
```
giuliastro: A non-string `data.details` shadows `data.message`.

`??` only falls through on `null`/`undefined`, so an adapter that sends both — e.g. `data: { details: { code: 429 }, message: "provider rate limit" }` — binds `details` to the object, the `typeof details === "string"` guard on the next line fails, and the session still gets a bare `Internal error`. That is exactly the case this PR is trying to fix, just with a structured `details` alongside the message.

Picking the first *string* of the two avoids it:

```suggestion
  const details = [error?.data?.details, error?.data?.message].find(
    (value) => typeof value === "string" && value
  )
```

(The `&& details` check on the next line then becomes redundant, but harmless.)

---

bridge/src/acp-client.js
giuliastro, line 1: The new `data.message` branch has no test coverage.

Reverting line 18 back to `error?.data?.details` alone keeps all 11 tests in `bridge/test/acp-client.test.js` green, so the PI rate-limit behaviour this PR adds can regress silently. The `data.details` branch does have a dedicated test (`bridge/test/acp-client.test.js:161`) — worth adding the mirror case with `data: { errorKind: "rate_limit", message: "provider rate limit" }` asserting `Internal error: provider rate limit`.

---

giuliastro (commented):
Bridge-only change to `acpErrorMessage`, low risk and it does what the description claims. The single call site (`bridge/src/acp-client.js:206` → `session.error` in `bridge/src/acp-service.js:517/540`) doesn't string-match on the error text, so nothing downstream regresses, and the existing tests pass.

Two notes inline: one correctness edge (a non-string `data.details` shadows `data.message`) and the missing test for the new branch.

---

bridge/src/acp-client.js
giuliastro: Fixed in 9fb8d05: the formatter now selects the first non-empty string from `data.details` and `data.message`, so structured `details` no longer hides an actionable provider message.

bridge/src/acp-client.js
giuliastro: Added regression coverage in 1d38e41 for PI's rate-limit shape, including the edge case where `data.details` is structured and `data.message` contains the actionable text.

bridge/test/acp-provider-error.test.js
```diff
@@ -0,0 +1,57 @@
+import assert from "node:assert/strict"
+import { EventEmitter } from "node:events"
+import test from "node:test"
+import { AcpClient } from "../src/acp-client.js"
+
+class FakeChild extends EventEmitter {
```
giuliastro, line 6: Non-blocking nit: this re-implements scaffolding that already exists.

`bridge/test/acp-client.test.js` has a `FakeChild(handler)` plus `fakeSpawn()` and `respondToHandshake()` helpers, and its test at line 161 (`"keeps the detail an adapter hides behind a generic error message"`) already covers `acpErrorMessage` — this new case is that test's sibling. Dropped in there it's about ten lines and no second fake to keep in sync as `AcpClient`'s handshake evolves:

```js
test("prefers the first string detail when data.details is structured", async () => {
  const client = new AcpClient({
    spawnProcess: fakeSpawn((child, request) => {
      respondToHandshake(child, request)
      if (request.method === "session/prompt") {
        child.respond({
          jsonrpc: "2.0",
          id: request.id,
          error: {
            code: -32603,
            message: "Internal error",
            data: { details: { code: 429 }, errorKind: "rate_limit", message: "provider rate limit" }
          }
        })
      }
    })
  })
  await client.start()
  await assert.rejects(client.request("session/prompt", {}), /Internal error: provider rate limit/)
  client.close()
})
```

Worth it only if you agree the two cases belong together — the coverage itself is already correct either way.

---

giuliastro (commented):
Both earlier findings verified fixed against `1d38e41`:

- `9fb8d05` picks the first *string* detail, and correctly drops the now-redundant `typeof` guard.
- `1d38e41` is a real regression guard, not just a green test — restoring the old `error?.data?.details` line makes it fail, and reverting that makes it pass again.

Full bridge suite: 89/89 passing. One non-blocking nit inline on the new test file; nothing blocking from my side.

---

bridge/src/acp-client.js
giuliastro: Thanks — agreed the shared scaffolding would be a little leaner, but this is non-blocking and the isolated regression test is intentionally narrow around the PI provider-error shape. Keeping it separate avoids touching the broader ACP client test file for this fix; coverage is correct and no production behavior changes.

bridge/test/acp-provider-error.test.js
giuliastro: Agreed this could share the existing test scaffolding. Since the comment is explicitly non-blocking and the isolated regression test is correct, I'm keeping this PR minimal rather than touching the broader ACP client test file again. The coverage remains focused on PI's provider-error shape.

PR #196, opened 2026-08-15
fix(client): read bare OpenCode 2 response bodies

### LINKED ISSUE #195: Cannot create a new session on OpenCode 2 backend: "Cannot read properties of undefined (reading 'directory')"
## Summary

Creating a new session against an OpenCode 2 backend fails with `Cannot read properties of undefined (reading 'directory')`. The same error surfaces in the new-session folder picker and in the create-session flow when a project directory is selected.

## Root cause

`GET /api/location` on the v2 server answers **bare** — `{"directory": "...", "project": {...}}` — without the `{ data, location?, cursor? }` envelope every other v2 route used by this client returns. `opencode2Api.loadPath` unwrapped `.data` and then read `.directory`, which threw on `undefined`.

Two sibling routes share the mismatch and silently degrade instead of crashing:

- `GET /api/health` (bare) — the connection test loses `version`
- `GET /api/project/current` (bare) — the project dashboard resolves empty

Verified against the live v2 server (`v0.0.0-next-17444`): all other endpoints the OpenCode 2 client calls (`/api/session*`, `/api/model*`, `/api/vcs*`, `/api/agent`, `/api/command`, `/api/skill`, `/api/form/request`) are properly enveloped.

## Reproduction

1. Add an OpenCode 2 server in Settings and test the connection.
2. Open the new-session folder picker, or create a session in a project directory.

## Fix

`web/src/opencode2-client.ts` now reads the raw body for the three bare-response routes. See the linked PR.

nitsuga:
## Summary

Fixes #195 — creating a new session on the OpenCode 2 backend crashed with `Cannot read properties of undefined (reading 'directory')`.

## Root cause

`GET /api/location` answers **bare** (`{"directory": ..., "project": ...}`) without the `{ data, location?, cursor? }` envelope the v2 client assumed for every route. `opencode2Api.loadPath` unwrapped `.data` and then read `.directory`, throwing on `undefined` — the folder picker and the create-session flow both call it.

Two sibling routes share the mismatch and silently degraded:

- `GET /api/health` — connection test lost `version`
- `GET /api/project/current` — project dashboard resolved empty

All other endpoints the client calls are enveloped (verified against the live v2 server).

## Changes

- `loadPath`, `health`, `loadProjectCurrent` now read the raw body via `v2Raw` instead of unwrapping `.data`
- Module comment documents the three bare-response routes

## Testing

- `npx tsc --noEmit -p tsconfig.app.json` — clean
- `npm run test:opencode2` (Node 24), `test:ui`, `test:model`, `test:settings` — all pass


nitsuga:
Closing — this should target the author's own fork (nitsuga/harness-remote), not upstream. Reopened there.

Issue #197, opened 2026-08-16
Harness Remote 3.0 master plan: Project -> Native Sessions -> Cross-Agent Continuity


giuliastro:
> **COMPLETED — Harness Remote 3.0.0 released on 2026-08-30.**
>
> This issue is retained as the historical 3.0 decision/release log. Current development starts from `main`; the RC/checkpoint instructions below are historical and must not be used as current release instructions.

## Purpose

This issue is the canonical Harness Remote 3.0 product, architecture and release plan.

## Current state

Harness Remote 3.0 has reached **release-candidate** status on:

```text
checkpoint/v3-session-first-working-2026-08-25
```

The current RC includes the Session-first migration and the later stabilization/fix sequence through PR #314.

For release planning, we now assume desktop/web and real-harness behavior are functionally acceptable unless new evidence proves otherwise.

The remaining release-critical validation is **Android/mobile testing on the exact RC**.

`main` remains the stable 2.x line until the final RC is accepted and a dedicated 3.0 release PR is prepared.

`archive/harness-3-2026-08-15` remains a protected recovery baseline.

## Product thesis

Harness Remote is not another coding agent and should not rebuild capabilities already owned by modern coding harnesses.

> Your sessions. Any coding agent. Any device.

The 3.0 user model is Session-first:

```text
Machine
  Project
    Native Session
    Native Session
    ...
```

A Native Session is the real Session owned by OpenCode, Codex CLI, Claude Code, OMP, PI or another supported harness.

A Project is the real selected working directory/repository.

Task, Run and Conversation may remain temporarily as internal compatibility implementation, but the user must not need those abstractions to discover, observe, resume or continue a native Session.

## Session-first rules

Harness Remote must preserve native Session identity and semantics.

A Session started in the native harness should be discoverable in Harness Remote where the harness exposes enough information.

A Session started in Harness Remote must be a real native harness Session.

Observation and writer ownership are separate capabilities. Reading a Session must not silently steal native writer ownership.

Harness Remote should represent per-harness capabilities honestly:

- discover/list;
- lookup by native Session ID;
- observe/read;
- create;
- resume/continue;
- writer ownership/takeover;
- Stop/cancel;
- rename/delete;
- live events;
- model catalog;
- model variants/reasoning controls.

Do not invent a universal behavior where the harness does not expose it.

## Cross-agent continuation

Switching coding agent remains a first-class 3.0 capability, but it is built on native Sessions.

Example:

```text
[...]

PR #200, opened 2026-08-16
integration: validate nitsuga TaskDesk stabilization work

giuliastro:
## Purpose

Controlled TaskDesk/Harness 3 integration tracked in #197.

**Base is `archive/harness-3-2026-08-15`, not `main`. This PR remains draft and must not be merged into the archive branch. `main` is untouched.**

## Important lineage correction

Audit of upstream PR #172 established that the full task-first/New Task client was **never merged into the archived Harness 3 line**. It remained in `agent/task-control-plane-consolidated` and was intentionally hidden at the end of that experiment (`TASK_LAUNCH_ENABLED = false`). nitsuga continued the merged Harness 3 infrastructure line, not that complete New Task UI.

The integration strategy is therefore:

1. archived Harness 3 baseline;
2. generic stability ideas adapted from nitsuga;
3. selective restoration of the task-first pieces from upstream #172;
4. isolated real-harness testing;
5. only later consider normal-app exposure.

## Stabilization primitives

- adapted dependency-free session mutation coordinator;
- no global lock: session/lane scoped arbitration;
- abort/permission/question can coexist with run work;
- model/agent discovery stays outside mutation locking;
- explicit reset/recovery and copied contexts;
- `catalog-request-guard` for stale model/agent responses scoped by profile/config/session/directory/request id.

No coordinator or catalog guard has been wired into `App.tsx` yet.

### Coordinator semantics after review

- Explicitly passing the currently selected session is still context-bound; navigation invalidates its result.
- Targeted work on another session may survive session navigation only while `profileID` and `configKey` remain unchanged.
- Server/profile changes invalidate all outstanding results, including targeted metadata work.
- Fork-generation tracking from the fork was intentionally not retained. The current coordinator prevents concurrent same-session run-lane work while a lease is active, but **once a lease has been released it cannot detect a later fork**. Any future Compact/Fork integration that needs post-release invalidation must add an explicit generation mechanism rather than assuming the coordinator already supplies it.

## TaskDesk endpoint and model block

Selectively restored from upstream #172:

- TaskDesk-only machine-daemon endpoint resolution separate from a saved direct OpenCode endpoint (e.g. OpenCode 4096 vs daemon 4097);
- legacy `discoverMachine()` remains one request to `/v1/machine` on the configured endpoint, so the existing connection wizard does not probe 4097 or change payload semantics;
- native/stringified payload normalization is limited to the TaskDesk discovery path;
- machine task client;
- machine/agent-level model catalogs for ACP and managed HTTP agents;
- durable prompt-less ACP catalog session instead of creating a probe session per New Task open;
- catalog sessions are filtered from public `/session`, `/v1/sessions`, `/experimental/session`, and `/session/status` results;
- `GET /v1/agents/:id/models`;
- task model persistence;
- fresh model validation immediately before launch;
- selected-model application to ACP and HTTP/OpenCode task sessions.

### Catalog latency and launch policy

Model discovery has an **8 second overall budget** for both ACP and managed HTTP catalogs. This budget wraps startup/load/fetch, so a stalled harness cannot leave New Task waiting for the old 90-second ACP request timeout.

Pre-launch validation is deliberately fail-closed:
- selected model no longer exists → HTTP 409;
- the agent/catalog cannot be freshly verified (including timeout/transient failure) → HTTP 503 and the task does not launch.

The ACP model catalog currently uses a **second lazy ACP client/process** so its durable catalog session cannot interfere with the user-facing ACP service/session state. That isolation has a real process/memory cost and must be measured during real-harness testing before promotion.

## Isolated TaskDesk test surface

The normal application still does **not** expose New Task. The integration surface is:

`/?taskdesk-test=1`

It is now **DEV-build only** and loaded with a dynamic import. Production builds ignore the query flag. The page renders the restored TaskLaunchDialog outside `App.tsx`, using the normally saved active server profile, and allows switching among agents exposed by the machine daemon.

The intended local sequence is documented in `docs/TASKDESK_LOCAL_TEST_PLAN.md`.

## Architecture guardrail

TaskDesk had already begun extracting UI/business logic from the oversized `App.tsx`. New task/model/session work must continue that decomposition. The restored Task launch dialog, task client, machine payload logic and request guards are separate modules; `App.tsx` remains untouched in this PR.

## Attribution

The session-state/stale-response ideas are adapted from Eric Schneider / `nitsuga`'s continuation of Harness 3. The task-first UI, machine endpoint resolution and agent model catalog block are selectively restored from this repository's own unmerged PR #172. Provenance is recorded here and in #197.

## Validation

Automated CI is used as a checkpoint after each slice, but promotion requires real browser + real harness testing and later Android testing. The PR is a validation vehicle, not a merge target.

Refs #197


giuliastro:
Superseded by #202 after renaming the active Harness Remote v3 / TaskDesk branch to `v3/taskdesk`. The new PR points to the same commit history and still targets `archive/harness-3-2026-08-15`. No changes were promoted to `main`.

PR #201, opened 2026-08-16
perf(bridge): trim common prefix before LCS merge replay

Baylar55:
`mergeReplay` built a full `(n+1)×(m+1)` Uint32Array LCS table on every ACP replay — on multi-thousand-message sessions that's tens of MB of transient allocation and the hot spot of a session load.

- Trim the shared prefix before running LCS, so the table covers only the divergent middle.
- Short-circuit the common append-only replay case (`previous` fully contained in `replayed`) to O(n) with no table at all.
- Signature matching stays cache-free — `messageSignature` computes signatures fresh on every merge, so messages mutated in place during streaming can never be compared against a stale signature. (Correction: an earlier revision of this description claimed the patch removed a cross-call signature cache; no such cache ever existed, in this patch or on `main` — the bullet was a leftover from an early draft.)
- Add a regression test for in-place mutation and a seeded differential test that proves the trimmed merge is identical to the previous full-matrix implementation.

All 93 bridge tests pass.

### DIFF
--- a/bridge/src/acp-service.js
+++ b/bridge/src/acp-service.js
@@ -73,39 +73,61 @@ function semanticHistorySignature(messages) {
   })))
 }
 
-function mergeReplay(previous, replayed) {
+/** Exported for testing only. */
+export function mergeReplay(previous, replayed) {
   if (previous.length === 0) return replayed
   if (replayed.length === 0) return previous
   const left = previous.map(messageSignature)
   const right = replayed.map(messageSignature)
-  const common = Array.from({ length: left.length + 1 }, () => new Uint32Array(right.length + 1))
-  for (let leftIndex = left.length - 1; leftIndex >= 0; leftIndex -= 1) {
-    for (let rightIndex = right.length - 1; rightIndex >= 0; rightIndex -= 1) {
-      common[leftIndex][rightIndex] = left[leftIndex] === right[rightIndex]
+
+  let prefix = 0
+  const maxPrefix = Math.min(previous.length, replayed.length)
+  while (prefix < maxPrefix && left[prefix] === right[prefix]) {
+    prefix += 1
+  }
+
+  if (prefix === previous.length) {
+    return [...previous, ...replayed.slice(prefix)]
+  }
+
+  const midLeft = left.slice(prefix)
+  const midRight = right.slice(prefix)
+
+  const common = Array.from({ length: midLeft.length + 1 }, () => new Uint32Array(midRight.length + 1))
+  for (let leftIndex = midLeft.length - 1; leftIndex >= 0; leftIndex -= 1) {
+    for (let rightIndex = midRight.length - 1; rightIndex >= 0; rightIndex -= 1) {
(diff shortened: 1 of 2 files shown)


giuliastro (commented):
Thanks — the optimization itself looks sound, and the differential coverage is exactly what I’d want for a replay merge change. Trimming the common prefix and short-circuiting the append-only case should materially reduce transient LCS allocation on long sessions.

One thing to clarify before merge: the PR description says it removes a cross-call signature cache that could return stale signatures after in-place mutation, but I don’t see that cache removal in the actual diff. Please either update the description to match the current patch or point out where that change lives.

Also, I don’t currently see CI/check runs attached to this head commit. Please make sure the repository checks run green on this exact SHA before merge.

Baylar55:
Thanks for the careful read — both points addressed:

**Description:** updated. The signature-cache bullet was a leftover from an early draft — no cross-call signature cache ever existed, in this patch or on `main` (signatures are computed fresh per merge; the only `Map` is the per-call `remainingBySignature` inside `mergeExternalHistory`). The description now says exactly that, and the in-place mutation regression test is in `bridge/test/external-history-dedup.test.js`.

**Checks on this SHA:** I can't attach them from this branch, and I believe it's structural rather than an oversight: `pr-checks.yml` only exists on `integration/nitsuga-taskdesk` (#200) — not on `main` — so a PR based on current `main` has no `pr-checks` workflow in its ref to trigger, and `workflow_dispatch` on this branch returns 403 without admin. Copying the file into this PR verbatim would also run red, because it references web scripts that likewise arrive with #200 (`test:agent-runs`, `test:mutation-coordinator`, `test:catalog-guard`, `test:machine-payload`, `web/scripts/set-debug-app-id.mjs` are absent from a main-based tree).

The full bridge suite (93 tests) passes locally on this SHA. To get real checks attached I see three options:

1. #200 lands first — I then update this branch and `pr-checks` fires on the synchronize push.
2. An admin dispatches `pr-checks.yml` on `perf/merge-replay-prefix-trim` directly.
3. I cherry-pick `pr-checks.yml` plus the missing script bits from #200 into this PR, if you'd rather not order the merges that way.

Tell me which you prefer and I'll do it.

giuliastro (approved):
Thanks for clarifying both points. The description now matches the patch, and the CI situation makes sense given the current main baseline. I do not want to couple this focused bridge optimization to #200 or pull TaskDesk-specific workflow pieces into this PR. Given the narrow scope, the differential coverage, and the full 93-test bridge suite passing on this SHA, this is good to merge. Nice work.

giuliastro:
Merged, thanks Baylar55. I also wanted to close the loop on the CI point. Your explanation was correct: the current main baseline does not contain the PR checks workflow that previously existed, while #200 is an isolated TaskDesk validation branch and should not become a dependency for this focused bridge change. I did not want to pull TaskDesk-specific workflow pieces into this PR just to manufacture attached checks. Given the narrow two-file scope, the seeded differential test against the previous implementation, and the full 93-test bridge suite passing on this exact SHA, I was comfortable approving and merging it as-is. Thanks for the careful follow-up and for the optimization.

Issue #253, opened 2026-08-18
TaskDesk 3.0: machine-centric home and Tasks view


giuliastro:
## Goal

Build the first user-facing TaskDesk 3.0 milestone on top of the released Harness Remote 2.11 baseline.

Target branch for implementation PRs: `v3/taskdesk`.

This issue is the first execution slice from #197.

## Product outcome

Replace the current session/server-first mental model with a machine-centric entry point where Tasks are a first-class object while ordinary Sessions remain directly accessible.

Target information architecture:

```text
Machine
  Projects
  Tasks
  Sessions
  Agents
```

The user should be able to open TaskDesk and immediately understand:

- which machine they are controlling;
- whether it is reachable;
- which agents/harnesses are available;
- which Tasks exist and their state;
- how to open ordinary Sessions without confusing them with Tasks.

### Machine-centric shell/home

- [ ] Make the connected machine the primary context instead of individual harness server profiles.
- [ ] Present discovered agents under the machine without exposing managed/internal ports.
- [ ] Show machine health in a compact, actionable way.
- [ ] Preserve compatibility with existing saved connection/profile data during the transition.
- [ ] Keep direct session access available.

### Tasks view

- [ ] Add a real Tasks view separate from Sessions.
- [ ] Source it from the existing task/machine APIs instead of inventing client-only task state.
- [ ] Show at minimum title/prompt summary, project, agent, model, lifecycle status and last activity where available.
- [ ] Define a stable mapping for current backend task states to initial UI states.
- [ ] Handle empty, loading, unavailable and error states explicitly.
- [ ] Make task rows/cards navigable toward the upcoming Task detail milestone without prematurely implementing the whole detail screen here.

### Navigation / terminology

- [ ] Make Tasks and Sessions visually and semantically distinct.
- [ ] Avoid presenting the incomplete TaskDesk flow as finished 3.0 functionality.
- [ ] Keep the UI usable on desktop and mobile layouts.

## Architecture constraints

- Do not add unrelated state/business logic back into the oversized `App.tsx` when a focused component/hook/controller is appropriate.
- Reuse the existing machine/task clients and daemon APIs where possible.
- Harness-specific details stay behind existing backend/profile abstractions.
- Do not expose managed OpenCode's internal loopback port.
- Do not break the 2.11 ordinary-session workflow.

## Non-goals for this slice

- Full New Task flow.
- Task detail/result/diff UI.
- Finish/review/cleanup.
- Subagent/fork tree.
[...]

PR #254, opened 2026-08-18
feat(taskdesk): add universal multi-agent session workspace

giuliastro:
## What changed

Reworks the first TaskDesk 3.0 UI slice around native sessions instead of introducing a separate Task dashboard.

- aggregates native sessions from all agents exposed by each saved machine daemon;
- keeps Session as the primary user-facing unit of work;
- treats project, harness, model and machine as session metadata rather than navigation hierarchy;
- adds global search plus All, Working, Needs You, Idle and Pinned views;
- groups and filters sessions by project and machine;
- opens the real native conversation and continues it through the existing prompt API;
- shows native diff, VCS branch, todos, permission requests and agent questions when the harness exposes them;
- adds New Session with Machine, Project, Harness, Model and Initial request selection;
- adds an experimental Continue with flow that creates a native session on another harness and hands over recent context without pretending session IDs are portable;
- keeps the existing 2.11 interface available as Classic during validation;
- preserves connection editing from the new workspace;
- uses local copy-owned UI primitives so the new shell does not depend on a new component runtime.

## Product direction under test

The goal of this PR is to test Harness Remote as a universal control plane over coding harnesses, not as another coding harness and not as a separate task manager.

Native harnesses continue to own their agent loops, tools, sessions and harness-specific behavior. Harness Remote normalizes discovery, navigation, status, attention and control across them.

## Validation

Type-check/build, web regressions and bridge tests on Linux, Windows and macOS must stay green. Browser testing with real Codex, Claude Code, OpenCode, Oh My Pi and PI sessions is the next release gate.

This PR remains draft until the new workspace is manually validated.

Refs #253
Refs #197

### DIFF
--- a/bridge/src/acp-service.js
+++ b/bridge/src/acp-service.js
@@ -231,6 +231,7 @@ export class AcpService {
   #replaying = new Set()
   #historyLoader
   #ownedSessions = new Set()
+  #adoptedSessions = new Set()
   #acpOpenSessions = new Set()
   #promptAcknowledgements = new Map()
   #titles = new Map()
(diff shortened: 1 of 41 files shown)


giuliastro:
Blocking bug found during real-machine testing.

The original infinite-loading race was real, but browser diagnostics exposed a second root cause in the same path: UniversalWorkspace was requesting `GET /v1/machine` with stale credentials and receiving `401 Unauthorized`, while Classic continued to work.

Two issues were involved:

1. Global refreshes could overlap every 4s and invalidate the foreground load before it cleared `loading`.
2. UniversalWorkspace collapses saved profiles by endpoint before discovery. When multiple saved profiles point to the same daemon, the first stored profile could win even if it had an old/empty password; the active profile with the credentials the user just configured could be discarded.

Fixes now on `feature/taskdesk-machine-home`:

- prevent overlapping global refreshes;
- bound per-agent native session loading to 12s;
- add browser timeouts to `/v1/machine` and machine API requests;
- put the active profile first when building the UniversalWorkspace machine list, so its current credentials win over stale duplicate profiles.

Latest fix commit: `025373586479833019ee30707487151a1f660e2c`.

Observed browser evidence before the credential-priority fix:
`GET http://localhost:4099/v1/machine net::ERR_ABORTED 401 (Unauthorized)`

PR remains draft and must not be merged until real-machine validation passes.

PR #261, opened 2026-08-19
fix(taskdesk): harden unified shell navigation and recovery

giuliastro:
## What changed

Follow-up hardening after the unified TaskDesk shell landed in #260.

### Unified product navigation

- makes `Open Session` from Task review open the exact native Session linked to the current Run;
- applies the same exact-session navigation from Questions and Permission requests in `Needs You`;
- prevents a transient wrong-session render by honoring the focus request during the first native-session refresh;
- clears conflicting Session filters when navigating from Task/attention context;
- keeps Tasks and Sessions inside the same persistent TaskDesk shell.

### Restart recovery

- adds regression coverage for ambiguous ACP state, temporary HTTP inspection failures and unknown HTTP session state;
- preserves the existing Run identity instead of failing or duplicating work when the daemon cannot yet prove a terminal state.

### OOM and performance hardening

A real Android test exposed a Node heap OOM after about 99 minutes of daemon uptime. The failure showed more than 4 GB of live heap. Investigation found that the v3 Sessions surface could turn lightweight index operations into full transcript work.

This PR now:

- keeps `/experimental/session` and `/session/status` metadata-only for ACP harnesses, so listing Sessions no longer restores every transcript snapshot into daemon memory;
- honors message `limit` responses and caps a message page at 500 items;
- prevents daemon-mode `limit=1` preview requests from materializing a complete ACP journal;
- removes the 18-session transcript preview fanout from the Sessions UI entirely;
- slows background reconciliation from 4 s to 10 s and selected detail refresh from 2.5 s to 5 s;
- prevents overlapping Session detail refreshes;
- pauses repeating refresh work while the app/webview is hidden;
- stops the Tasks polling tree while viewing Sessions or Classic;
- stops Task detail polling as soon as its Task view is no longer visible;
- loads only the Task detail data required by the active tab rather than messages, diff, todos and VCS on every detail refresh;
- adds `/v1/diagnostics` with RSS, heap, external memory and uptime for real-device soak testing;
- adds bridge and web regression tests specifically guarding these performance invariants.

This is deliberately not an `--max-old-space-size` workaround. The goal is to remove unbounded work and transcript retention from index/preview paths rather than merely increasing V8's crash threshold.

### Design pass

- applies the requested Design Taste review to hierarchy, typography, density, responsive behavior and removal of remaining nested-app chrome;
- keeps the established dark developer-tool visual language rather than introducing a separate Sessions design system.

## Validation

Performance-hardening HEAD `a034f47a27e8720e2d74ed2f3b6fba523d89d97c` passes:

- TypeScript and production web build;
- web regression suites, including TaskDesk performance guards;
- bridge tests on Linux;
- bridge tests on macOS;
- bridge tests on Windows;
- desktop application menu verification on macOS;
- Android debug APK build;
- Android APK signature verification.

Android artifact: `harness-remote-debug-v2.11.1-a034f47`, built from the exact HEAD above.

The PR remains draft until a real daemon soak test confirms that memory plateaus under the workload that previously triggered the OOM.

## Safety

Targets `v3/taskdesk` only. `main` is untouched.

Refs #197
Refs #253


giuliastro:
Performance hardening candidate updated to `e4f1a97240e5f0315497b560768537998fbfaa56`.

Additional guard: `bridge/test/taskdesk-memory-pressure.test.js` simulates 160 historical sessions with deliberately unreadable snapshots. The test verifies that `/experimental/session`, `/session/status`, and daemon-mode `limit=1` preview paths remain metadata-only and do not restore transcript snapshots or open ACP sessions.

Validation is green on Linux, macOS, Windows, web build/regressions, desktop menu verification, Android debug APK build, and APK signature verification.

Artifact: `harness-remote-debug-v2.11.1-e4f1a97` from workflow run `32289134885`.

Keeping the PR draft until a real daemon soak test confirms memory plateaus under the workload that previously triggered the OOM.

giuliastro:
Latest candidate: `755ddaa1e551ee7c0a2a0458461af1e224901233`.

Real Android testing confirmed the OOM hardening is a major improvement: after about 709 seconds of interactive use the daemon reported about 42 MB heap used and 142 MB RSS instead of the previous multi-GB growth.

This follow-up also incorporates the mobile UX findings from that test:

- mobile Sessions is now a drill-in flow: Session list first, full conversation second, explicit Back to Sessions;
- mobile Task detail is a dedicated focused page rather than a compressed split view;
- embedded desktop/browser Sessions now keeps only Session list + conversation as persistent panes; the internal navigation is removed and the inspector is on demand;
- Session ordering uses real live daemon activity without loading transcript bodies;
- Continue remains same durable Task, new numbered Run, new native Session;
- previous Runs are explicitly reviewable instead of exposing only the latest outcome;
- repeated ACP prompt echoes are filtered so one submitted question is not rendered twice, while two genuinely separate identical prompts are preserved.

Validation for this exact HEAD is green on web build/regressions, bridge tests on Linux/macOS/Windows, desktop menu verification, Android debug APK build, and APK signature verification.

Android artifact: `harness-remote-debug-v2.11.1-755ddaa`, workflow run `32339604654`.

Keeping the PR draft until this mobile navigation candidate is checked on a real device. `main` remains untouched.

giuliastro:
New mobile test findings:

1. Mobile create actions are missing. Root cause is confirmed: the base mobile CSS hides the last TaskDesk topbar button at <=680px, which removes `New Task`; embedded Sessions also hides UniversalWorkspace navigation, where `New Session` currently lives. Both create actions must remain directly reachable on mobile.

2. Session detail can jump to another Session after briefly showing the one the user selected. Root cause is confirmed in `UniversalWorkspace.refreshAll()`: a prior `focusSessionRequest` is still given priority on every refresh, so a request originally used for Task -> Session navigation can override a later manual Session selection. The focus request must be consumed once, then ordinary refreshes must preserve the user's current selection.

3. Current `Continue Task` semantics need a product correction, not just UI polish. Today `continue()` calls a fresh launch, which creates a new Run and a new native Session, then sends only the continuation prompt into that new Session. It preserves the Task workspace but not the prior conversational context. That makes it a fresh attempt in the same workspace, not a true conversational continuation.

Proposed product semantics:
- Task = durable goal, workspace and review lifecycle.
- Session = native harness conversational context.
- Run = one execution/turn recorded under the Task and may reference the same Session as an earlier Run.
- Continue = same Task, same native Session, new Run/turn so context is preserved.
- New Run / Retry fresh = explicit new Session.
- Fork = explicit new Session with carried context if we add that workflow later.
- Ad-hoc `New Session` remains valid for work that does not need Task lifecycle.

Keeping #261 focused on the two confirmed mobile UI/navigation bugs. The continuation semantic change should be implemented and tested as a separate backend/product change rather than slipped into this UI candidate. `main` remains untouched.

giuliastro:
Latest mobile/navigation candidate: `b877b151ca6a9684ba71e37900ac1ed16a5c5f57`.

Follow-up from device testing:

- restores a visible `New Task` action on the mobile Task list;
- restores a visible `New Session` action on the mobile Sessions list while delegating to the existing creation flows;
- protects an explicitly chosen mobile Session from being replaced by a stale background focus/refresh selection;
- folds these guarantees into the TaskDesk UI regression suite.

Validation for this exact HEAD is green on TypeScript/web build, web regressions, bridge tests on Linux/macOS/Windows, desktop menu verification, Android debug APK build and APK signature verification.

Android artifact: `harness-remote-debug-v2.11.1-b877b15`, workflow run `32347182945`.

The product direction for Tasks is now tracked in #263: Tasks remain because they will become the durable cross-harness orchestration unit, with intelligent Continue/Handoff choosing harness, model, role and inspectable Task Context.

PR #262 has also been realigned on this HEAD and is mergeable again as the independent performance track. `main` remains untouched.

PR #262, opened 2026-08-20
perf(taskdesk): move session refresh toward event-driven bounded state

giuliastro:
## Purpose

Parallel performance track for TaskDesk, intentionally separated from the UI/UX hardening in #261.

This branch starts from the current #261 candidate and targets `feature/taskdesk-unified-shell`, not `main` and not `v3/taskdesk` directly.

Planned scope:

- prefer event-driven refresh where the daemon already exposes session events;
- retain polling only as a fallback and reconciliation path;
- add bounded transcript caching and eviction so opened conversations cannot accumulate without limit;
- introduce real message pagination for long native session histories instead of materializing an entire transcript;
- keep session listing and status metadata-only;
- preserve the `/v1/diagnostics` soak-test path and memory-pressure regression coverage.

Initial commit adds the explicit architecture and acceptance gates for this independent track.

Safety:

- no `main` changes;
- no direct `v3/taskdesk` target;
- no merge until #261 UI behavior and this performance track both pass their own validation gates;
- mobile navigation redesign and memory/cache changes stay in separate commits and PRs.

Refs #197
Refs #253


giuliastro:
Superseded by the integrated candidate in #271, which has now been merged into `v3/taskdesk`. The exact #262 branch HEAD is an ancestor of that candidate (0 commits left behind), so no code from this PR is being discarded. Remaining frontend event-driven refresh / older-message paging work stays tracked in #266.

PR #268, opened 2026-08-20
fix(taskdesk): UI audit, navigation repair, appearance and language

### LINKED ISSUE #265: TaskDesk parallel track: UI audit, polish and bug fixing
## Purpose

Run an independent UI-focused TaskDesk pass in parallel with the performance and multi-harness backend tracks.

Work branch:

`fix/taskdesk-ui-audit-polish`

Base branch:

`feature/taskdesk-unified-shell`

Do not touch `main` and do not merge this work directly into `v3/taskdesk`.

## Scope

Audit the current TaskDesk interface as a real product, not as a collection of screens. Find, reproduce and fix UI bugs and rough edges across desktop and mobile, while preserving the current TaskDesk information architecture.

Focus on:

- Tasks, Sessions, Needs You, Projects, Agents and Machines navigation;
- New Task and New Session discoverability and reliability;
- Task detail navigation, close/back behavior, tabs and actions;
- exact Session navigation from Task detail and attention items;
- mobile master/detail behavior and back navigation;
- responsive layout at narrow, medium and desktop widths;
- overlapping, clipped, duplicated or inaccessible controls;
- typography, spacing, density, hierarchy and visual consistency;
- loading, empty, offline and error states;
- keyboard/focus behavior and obvious accessibility regressions;
- terminology consistency between Task, Run and Session;
- ensuring Tasks and Sessions feel like one TaskDesk product instead of two applications.

Use the existing Design Taste principles already adopted by the project. Prefer a dense professional developer-tool UI over decorative landing-page styling.

## Non-goals

Do not redesign the backend task model.
Do not work on transcript caching, event transport, pagination or daemon memory architecture.
Do not implement multi-harness Task Context or handoff semantics.
Do not modify Classic 2.x behavior except to prevent a TaskDesk regression.
Do not merge the PR.

## Required workflow

1. Check out `fix/taskdesk-ui-audit-polish` and confirm it starts from the current `feature/taskdesk-unified-shell` candidate.
2. Read #197, #253 and PR #261 before changing code.
3. Run the current web app locally and inspect both desktop and mobile widths.
4. Create a short reproducible bug list before fixing anything. Prioritize functional navigation bugs first, then usability, then polish.
5. Fix issues directly on this branch in focused commits.
6. Add or update regression tests for every functional bug fixed.
7. Run TypeScript/build, web regressions and relevant bridge tests.
8. Open or update a draft PR targeting `feature/taskdesk-unified-shell` only.
9. In the PR body, list every bug found, what was fixed, what remains, and which items still require real Android validation.

## Acceptance criteria

- no overlapping or inaccessible primary controls at common desktop and mobile widths;
- New Task and New Session remain clearly distinct and reachable;
- selecting/opening a Task always produces an obvious visible result;
- Task detail can be entered and exited reliably on mobile and desktop;
- Open Session reaches the exact intended Session;
- Sessions and Tasks share one coherent TaskDesk shell;
- no hidden duplicate navigation chrome remains in the TaskDesk Sessions view;
- no new polling or backend performance work is introduced here;
- automated UI regressions are green;
- PR remains draft pending manual Android validation.

Refs #197
Refs #253
Refs #261

giuliastro:
## What this is

An independent product-level UI pass over the TaskDesk shell, run against a local machine daemon at desktop, narrow-desktop, tablet and phone widths. Every functional bug below was reproduced first, then fixed, then covered by a test.

Two additions were requested during the pass and are included: a Settings surface for appearance and language, and the same layered back navigation Classic already has for Android's hardware back and edge-swipe gesture.

Targets `feature/taskdesk-unified-shell`. `main` is untouched.

### Functional

1. **The Sessions view froze the app at any width above the phone breakpoint.** Both TaskDesk DOM installers observed `attributes` on `document.body` and mutated the DOM from the observer callback. `classList.remove()` writes the `class` attribute even when the token is absent, so once `.td3-sessions-embedded` was mounted the observer re-entered itself without end. Reproduced by clicking Sessions at 1280px: the renderer was pegged and the tab never recovered. This is why the view worked on a phone (the other branch ran) and not on a desktop.

2. **New Task was removed by `display: none` at every width at or below 1220px.** `.td3-topbar > .td3-button:last-child { display: none }` appears in all three media blocks. Measured `display: none` at 1100, 900, 760 and 375px. A phone got a replacement button injected into the page heading; 681–1220px got nothing at all.

3. **New Session was unreachable everywhere except a phone.** It lived only in `.uw-nav`, which `v3-polish.css` hides with `display: none !important` inside the TaskDesk shell, and in the no-selection welcome state — which is unreachable in practice because the refresh auto-selects the first session whenever any exists.

4. **One "Open Session" pinned the Session selection for the rest of the session.** `refreshAll` preferred `focusSessionRequest.sessionID` on *every* refresh and the request was never cleared, so each 10s cycle discarded the user's manual choice. The `appliedFocusRequest` guard only covered a separate effect. The mobile installer's `restoreManualSelection` — which programmatically re-clicked the card React had just deselected — was fighting this rather than fixing it.

5. **An open Task detail replaced itself with a loading spinner every few seconds.** The detail effect listed `selected?.task.updatedAt` in its dependencies and did a non-silent reload, and a running Task moves `updatedAt` on every poll. Review, Conversation or Diff blanked repeatedly and lost scroll position.

6. **A machine that stopped answering closed the Task you were reading.** One failed discovery emptied that machine's task list for a cycle, the selection was dropped, and the open detail collapsed to "Select a Task".

7. **Machines and Manage machines were unreachable on a phone.** `nav button:nth-child(n+7) { display: none }` hid Machines as the seventh item and `.td3-sidebar-bottom { display: none }` hid Classic and Manage machines. After first run there was no way to add, edit or remove a machine.

8. **Permission and question replies swallowed failures.** Fired from inline handlers with no `catch` and no busy state: a rejected reply produced an unhandled rejection, no message, and buttons that stayed live for double submission.

9. **Nothing closed on Escape, and Android back left the app from any depth** — including with a modal open. Classic implements this; TaskDesk did not.

10. **The appearance preference did nothing.** The theme effect lives inside `App.tsx`, so TaskDesk — which boots first — never applied `data-theme` at all, and every v3 surface was authored as literal dark hexes regardless. There was also no language selector and no translated strings, while Classic ships four languages.

11. **Keyboard focus was invisible.** `outline: 0` is set on buttons, selects and text fields across both shells with no `:focus-visible` replacement, so the navigation, toolbar, machine picker, filters, detail tabs and every form control gave no indication of focus.

### Usability and layout

12. **The Tasks list reserved a dead 500px column with no Task open.** The unified sheet was imported from a component, which put it near the front of the bundle; `.td3-tasks-layout` then won on order against the equally specific `.td3-tasks-layout-unified`. Measured `750px 500px` at 1440px.

13. **`.td3-shell button` outranked every single-class component rule.** At (0,1,1) it beat `.td3-task-row`, so the Tasks table had been rendering at the shell's 13px with every cell at full text strength instead of 11.5px with muted metadata — the title/metadata hierarchy was flattened and the table was looser than designed. Because `taskdesk-v3.css` loads after `universal-workspace.css`, the same reset also overrode the Sessions pane's button colours from a different sheet.

14. **Nine selectors rendered at 7–7.5px** — diff counters, todo markers, avatar labels, attention-card actions, an inspector link — because the two sheets that exist to raise those sizes had missed them.

15. **Machine scope was ignored by Overview, Needs You, Projects, Agents and the Tasks badge**, while the Tasks list honoured it, and the harness strip described a single machine while labelled "All machines".

16. **The Sessions view lost Refresh and connection settings** (`.uw-top-actions` is hidden in the embedded shell) and its primary action was a redundant "Tasks" button duplicating the sidebar.

17. **A 760px window left 336px for the conversation** and its composer, next to a 360px Session list.

18. **No offline or loading state for Tasks** — an unreachable machine rendered as "No Tasks match this view", indistinguishable from an empty filter.

## What was fixed

**The DOM-injection layer is gone.** `taskdesk-mobile-navigation.ts` and `taskdesk-run-history.tsx` are deleted. The phone master/detail pane is React state on the shell and the class the stylesheet keys off is rendered from it; the Session back control lives in the Session header; reviewing an earlier Run opens a real dialog using the machine, task and agent the shell already holds instead of scraping a session id out of a `<dt>` and re-listing every machine's tasks. Nothing observes or repairs the rendered tree any more, which is both the fix for (1) and the reason (4)'s workaround is no longer needed.

**Create actions are one persistent toolbar action chosen by the surface** — New Task on Task-centric views, New Session on Sessions — and the *label*, not the control, collapses at narrow widths. This fixes (2) and (3) and removes the duplicate "Tasks" button from (16).

**Selection is stable.** A focus request is applied once (4). An unreachable owner keeps the selection instead of dropping it (6). Only the Task, tab and pane visibility trigger a visible reload; activity refreshes silently (5).

**Phone navigation is four destinations plus an explicit More sheet** carrying Overview, Agents, Machines, Settings, Manage machines and Classic, so nothing is hidden by position (7).

**Attention replies report failure** on the page and disable while in flight (8).

**One ordered back stack** serves Escape, Android's hardware key and the edge-swipe-back gesture the webview delivers as the same event: run review → More → Settings → Continue → New Task → mobile Session detail → Task detail → back to Tasks, and only then exit (9).

**Settings for appearance and language** (10). The preference is shared with Classic under the keys Classic already used, applied once at startup, and the whole TaskDesk shell is translated into English, Italian, Traditional Chinese and Simplified Chinese. `Task` and `Run` deliberately stay untranslated as product object names; `Session` follows Classic's existing wording.

**The palette is real.** All 470 colour literals across the TaskDesk shell and the embedded Sessions pane now resolve through one token set, so a theme is a redefinition of that set. The tokens that mattered most were the lifts: a dark surface is raised with translucent white and a light surface with translucent ink, so a hard-coded `rgba(255,255,255,.03)` paints an invisible band on white whatever else the palette says.

**A visible focus ring** from the accent already used for selection, drawn inside the box for full-bleed rows and the phone tab bar so a scroller cannot clip it (11). Dialogs now focus the field they exist to collect rather than their header close button.

**Layout and density**: the TaskDesk stylesheets are ordered base-to-override in one place (12); the control resets are wrapped in `:where()` so they normalise without outranking components (13); the sub-8px text is raised to a 9.5px floor (14); every page honours machine scope, and Needs You accounts for items on machines outside it rather than dropping them (15); the tablet conversation gets 400px instead of 336px (17); Tasks has explicit loading and "no machine is reachable" states with a Retry action (18).

`100vh` also becomes `100dvh`, the topbar's phantom fifth grid column is gone, the nav uses real icons instead of mixed glyphs, and the icon rail hides labels accessibly instead of with `font-size: 0`.

## How it was verified

Driven against a local stub of the machine daemon serving five Tasks across five harnesses, seven Sessions, a pending question and a pending permission — at 1440, 1100, 900, 760 and 375px.

Measured rather than eyeballed, because the theme work needed evidence: for every rendered text node the real background chain is composited and the WCAG contrast ratio computed. That is what found the 1.2:1 active navigation label, the 3.5:1 primary button, the 4.0:1 brand monogram and the two text tiers sitting at 3.4–4.1:1 — and it is how the result is stated: **both themes now pass AA on every measured node** (Tasks surface 82 nodes, Sessions pane 150 nodes, 0 failures each), with no rendered text below 10px.

Behaviour confirmed in the running app: Sessions no longer freezes; a manual Session choice survived three refresh cycles after an Open Session navigation; an open Task detail showed zero loading spinners across 128 samples over 32s with a Task whose `updatedAt` moved every poll; Escape unwound run review → Task detail; the phone drilled list → conversation → back; the phone Task detail is a full page whose action bar ends exactly at the tab bar; Machines is reachable from the More sheet; a failing permission reply surfaced the daemon's own message; offline showed the new state and Retry recovered.

## Tests added

- rewrote the two suites that covered the mobile behaviour — they asserted the *shape of the injection code*, which is why they stayed green while the app froze, and now assert the behaviour;
- `taskdesk-appearance.test.mjs`: every dark token has a light value and vice versa, no sheet hard-codes a colour outside the palette, the resets stay at zero specificity, no authored text below 9.5px, the stylesheet load order is base-to-override;
- the back stack's dismissal order is asserted as an ordered list, so a reordering is a test failure;
- an open Task detail survives a poll and an unreachable owner;
- attention replies report failure instead of dropping it;
- every TaskDesk string resolves in all four languages with matching placeholders;
- also fixed a pre-existing assertion in `taskdesk-home.test.mjs` that could only ever match on an LF checkout, so it had been silently failing on Windows.

## Validation

TypeScript, production web build, all web suites (`test:ui`, `test:settings`, `test:i18n`, `test:platform`, `test:config`, `test:model`, `test:events`, `test:profiles`, `test:markdown`, `test:attachments`, `test:agent-runs`, `test:mutation-coordinator`, `test:catalog-guard`, `test:machine-payload`, `test:completion-audio`, `test:completion-timing`, `test:native-response`) and bridge tests on Windows (261/261) all pass.

## Still needs real Android validation

- the `backButton` listener: the hardware key and the edge-swipe-back gesture were wired the way Classic wires them and verified through the Escape path that shares the same stack, but not on a device;
- theme and language switching in the Android WebView, including `prefers-color-scheme` following the system setting for "As device";
- `100dvh` against the Android URL bar and the soft keyboard;
- the phone tab bar and the full-page Task detail on a real screen and with real touch targets;
- CJK rendering and label wrapping in the bottom tab bar at Traditional and Simplified Chinese.

Two limits of this environment worth flagging: the browser pane could not composite frames, so all layout evidence is measured geometry and computed style rather than screenshots; and viewport changes did not emit `resize`/media-query `change` events, so responsive behaviour was verified by reloading at each width. The `useMediaQuery` hook listens to both events for that reason.

## Not touched

Daemon memory architecture, transcript caching, SSE/event transport, message pagination, Task Context and multi-harness handoff semantics are all left to the parallel tracks. No polling was added and the existing performance guards still pass. Classic 2.x behaviour is unchanged; it picks up an appearance or language change made in TaskDesk on its next mount, which is worth a follow-up if live sync is wanted.

Refs #197
Refs #253
Refs #261
Closes #265


giuliastro:
Superseded by the integrated candidate in #271, now merged into `v3/taskdesk`. The exact #268 branch HEAD is an ancestor of the candidate (0 commits left behind), and the final waiting-indicator follow-up was validated on top of it in the integrated branch. Closing this component PR without a separate merge.

Issue #287, opened 2026-08-22
Harness Remote 3.0 backend audit: transport reliability, model catalogs and subscription lifecycle


giuliastro:
## Purpose

This is the next release gate for Harness Remote 3.0 after the conversation-first UI beta in PR #286.

The goal is not to add more product features. The goal is to audit the backend and every harness adapter deeply enough that real conversations, model catalogs and live transport are trustworthy on desktop, web and Android.

### 1. Mobile conversations disconnect or stop unexpectedly

On a local network, Android conversations sometimes stop, disconnect or show red server errors while the harness may still be working.

Audit the full path:

`mobile UI -> Capacitor HTTP/event transport -> machine daemon -> harness adapter -> native Session`

Check request timeouts, stream/event lifetime, reconnect behavior, accepted-prompt recovery, cancellation, polling/reconciliation races, stale error propagation and daemon-side timeouts.

Do not mask real failures by only increasing timeouts. Determine which operation timed out, whether the harness accepted the prompt, whether the native Session kept running, and how the UI should recover.

### 2. Model catalogs are inconsistent across harnesses

OMP and PI are configured with broadly similar model access, but Harness Remote currently exposes noticeably different catalogs. PI often fails the first time it is selected, reports a disconnection, leaves the model picker unusable, then works after switching to another harness and back.

OpenCode exposes model options/variants such as reasoning levels where available, while other harnesses often do not.

Audit model discovery per adapter from source to UI:

- where each harness obtains its model list;
- whether it is live, cached, static, provider-derived or inferred;
- provider/model IDs and normalization;
- default model selection;
- variants/options such as low, medium, high and equivalent harness-specific capabilities;
- stale catalog handling;
- first-selection race conditions;
- lazy harness startup and model discovery ordering;
- cache keys and invalidation when switching harness/machine;
- errors that leave the picker disabled or stuck;
- whether equivalent model access is represented consistently without inventing capabilities a harness does not expose.

Produce an explicit capability matrix for OpenCode, Codex, Claude, PI and OMP showing exactly what Harness Remote can discover and control for each one.

### 3. OpenCode listener leak warning

The backend repeatedly logs:

```text
MaxListenersExceededWarning: Possible EventTarget memory leak detected. 11 event listeners added
```

Treat this as a real lifecycle bug until proven otherwise.

[...]

PR #286, opened 2026-08-22
beta(v3): conversation-first universal agent control plane

giuliastro:
## Goal

Harness Remote 3.0 uses the product model:

`Project -> Conversation -> native Sessions`

Harness Remote stays above coding agents instead of recreating them.

## Current beta direction

- Conversation is the primary user-facing unit.
- Native harness Sessions remain the source of conversation truth.
- A Conversation can continue across OpenCode, Codex, Claude, OMP and PI Sessions.
- `Continue with` changes coding agent/model from the same Conversation and shows the transition in history.
- Native Session continuity is inspectable in Sessions.
- Changes reflects the real Project workspace.
- Normal Conversations run in the real Project directory, with no hidden automatic worktree.
- Machines, projects, coding-agent discovery, model catalogs, Settings, permissions/questions, Stop and live refresh are retained.
- Classic and the separate Advanced Native Sessions product entry points are removed from the 3.0 shell.
- Android uses explicit Conversations, Machines and Settings navigation.
- Chat keeps response-start feedback, live Activity and local tool errors without turning one failed tool into a failed whole turn.

## Current validated beta head

`8683f4e79d91431a4903ec0e4d8e30c331091ce7`

Automated checks on this head are green:

- production TypeScript + Vite build: PASS
- full web regression suite: PASS
- conversation-first regressions: PASS
- mobile navigation/layout regressions: PASS
- bridge tests on Linux/macOS/Windows: PASS
- desktop application menu verification: PASS
- Android debug APK build: PASS
- APK signature verification: PASS

This is now a usable UI beta baseline, not a release candidate.

## Backend release gate

Real testing has exposed backend reliability issues. These are now the next blocking phase and are tracked canonically in #287.

Priority problems:

1. Android/local-network conversations sometimes disconnect or show red server errors while the native harness may still be working.
2. Model catalogs and model options are inconsistent across OpenCode, OMP, PI and other harnesses. PI frequently fails on first selection and only recovers after switching away and back.
3. OpenCode repeatedly emits `MaxListenersExceededWarning`, which must be treated as a listener/subscription lifecycle bug until disproved.

The next session must audit the backend end-to-end rather than adding more UI features.

Required work includes:

- request/prompt timeout policy and accepted-prompt recovery;
- event/SSE/ACP subscription ownership, reconnection and cleanup;
- native Session create/resume/prompt/Stop routing;
- model catalog source, caching, startup ordering and invalidation per harness;
- model/provider IDs, variants and reasoning options;
[...]

PR #288, opened 2026-08-22
fix(v3): backend reliability and harness adapter audit

giuliastro:
## Scope

Backend and harness-adapter reliability audit for the conversation-first Harness Remote 3.0 beta.

Canonical product plan: #197  
Backend audit: #287  
Parallel pre-merge audit: #289  
Beta baseline: #286

This PR targets `feature/conversation-control-plane-rc1` and remains **DRAFT**. It does not touch `main`, `v3/taskdesk`, or `archive/harness-3-2026-08-15`.

### Android transport recovery

The native Android SSE reader previously allowed an infinite read wait. After backgrounding, sleep, Wi-Fi handoff or temporary network loss, a dead TCP stream could remain blocked while the native harness Session continued.

Fix:
- finite native SSE stall watchdog;
- existing reconnect/backoff is reused;
- accepted prompts are reconciled rather than blindly resent;
- regression prevents the infinite read timeout from returning.

Real-machine OpenCode background/foreground and short network interruption recovered without duplicate prompts or permanent disconnect.

### OpenCode event ownership

Downstream reconnect churn could create multiple upstream OpenCode `/global/event` subscribers.

Current ownership:

`OpenCode /global/event -> one daemon-owned upstream SSE -> N downstream clients`

Diagnostics expose upstream/downstream/reconnect state and regression coverage requires the upstream count to plateau at one. Listener limits were not raised.

### PI cold model discovery

Cold ACP adapter startup/auth/model discovery can outlive one browser/mobile request. Discovery is now daemon-owned and single-flight; the model endpoint can return bounded `202 loading` while the same operation continues and clients poll it.

Real-machine result: PI selected first after cold daemon restart loaded on the first attempt and completed a simple Conversation prompt.

### ACP catalog lifetime

Historical prompt-less technical Sessions are no longer loaded as current model membership after daemon restart. Persisted technical Session ids are retained only so they remain hidden. Discovery starts from a fresh technical Session for the current adapter lifetime.

### Model discovery timeout contract

ACP discovery now has one end-to-end budget instead of several independent long waits. Optional variant probing is bounded separately so slow enrichment cannot make valid base models unusable.

Diagnostics expose phase, duration/error and variant-probe completeness.

### Conversation availability is independent from model-catalog warming

- New Conversation can start while models are still loading.
- Existing Conversation history/chat remains usable if model discovery is slow or unavailable.
- Switching harness clears the previous harness model immediately.
- Continue with no verified current model sends `model: null`, explicitly requesting the harness-native default instead of accidentally reusing an old persisted selection.

### Harness-specific options are preserved without fabrication

Only runtime-advertised controls are emitted:
- OpenCode: provider-advertised variants;
- OMP: `thinking` when advertised;
- PI: `thinkingLevel` or compatible runtime alias when advertised;
- Codex: `reasoning_effort` when advertised;
- Claude: no invented low/medium/high control.

### Continue and Stop recovery

Continue now persists a durable Run acceptance boundary and `clientRequestId` before slow Session/model/context work. Ambiguous client retries converge on the accepted Run rather than creating duplicate native prompts.

Stop reconciles authoritative Work Thread state after a lost response, so a successful native abort is not overwritten by a stale transport error.

### Managed OpenCode retry

A transient lazy-start failure does not permanently circuit-break the managed OpenCode host. Later authenticated use can retry startup.

## Deepseek / native-filter correction

A real-machine report initially suggested that Harness Remote was exposing a removed Deepseek/OpenCode Zen model. A speculative follow-up layer then launched a second native `pi` / `omp` model command as membership authority.

Further testing on both Linux and Windows showed the same Deepseek v4 free entry in the native PI and OMP catalogs as well, so the original signal was likely transient/provider-side rather than proof that ACP membership was stale.

The speculative native PI/OMP filtering layer has therefore been removed completely. Harness Remote does **not** launch a second native harness process for membership filtering and does not use model-name blacklists.

The tree for that area is back to the simpler ACP-based architecture from `7509db6`.

## Diagnostics

`/v1/diagnostics` exposes safe control-plane state including:
- machine/agent/process state;
- ACP pending RPC count/method/Session/age;
- listener/subscriber counts;
- in-flight HTTP requests and duration;
- event fanout and reconnect counters;
- active/queued Sessions and loads;
- transcript/cache/snapshot state;
- model catalog source, age, phase, in-flight status and error.

Prompt bodies, credentials and generated auth values are excluded.

## Frozen audit baseline

The current frozen shared baseline for the final parallel audit is:

`56c5c9d925045fbef542650c6d564ce502c72758`

PR checks are green on this line. The native-filter experiment has been fully removed.

Two independent branches now start from this exact SHA under #289:
- `audit/v3-capability-contract`
- `audit/claude-v3-product-ui`

Do not add unrelated work directly to this frozen branch while the parallel audit is running.

## Remaining architecture gate

The model list is not the complete capability contract. Following the #287 discussion, the final backend audit must explicitly represent per-harness capabilities such as transport/protocol, tool-call representation, model/variant metadata and lifecycle guarantees.

Where harness configuration can be project-specific, discovery/cache identity must include validated project/cwd in addition to machine + harness. This will be implemented on `audit/v3-capability-contract`, without spawning duplicate native harness processes.

## Remaining real-machine gate

Before promotion:
1. review both #289 workstreams independently;
2. validate a known-live/default model on OpenCode, PI, OMP, Codex and Claude;
3. exercise long turns and Stop;
4. repeat harness switching;
5. restart daemon and verify Session recovery/continuity;
6. verify diagnostics plateau for requests, subscribers, caches/listeners and OpenCode `upstreamStreams <= 1`;
7. run Android smoke tests from the exact final candidate SHA;
8. require green CI on that exact SHA.

## Merge rule

Do **not** merge yet.

#286 remains DRAFT. #288 remains DRAFT. After both #289 workstreams are reviewed, accepted changes should be integrated into a fresh candidate and tested before anything moves toward the historical `v3/taskdesk` branch.

Refs #197
Refs #286
Refs #287
Refs #289


giuliastro:
Additional confirmed finding after the main PR description update:

### Android native SSE could remain silently dead forever

`LiveEventsPlugin` used `HttpURLConnection.setReadTimeout(0)`. After Android sleep/background, Wi-Fi handoff or a brief local-network loss, a stale TCP connection can stop delivering bytes without immediately closing. The native reader could therefore block forever while the native harness Session continued working. That explains a class of "conversation stopped/disconnected" reports that cannot be fixed by increasing model/request timeouts.

The audit branch now uses a 30s native SSE read watchdog. The daemon emits a heartbeat every 10s, so three missed heartbeat windows are treated as a dead stream and the plugin falls back into its existing reconnect/backoff loop. A regression explicitly forbids returning to `setReadTimeout(0)`.

This is separate from the Continue/Stop HTTP reconciliation work: HTTP mutations now reconcile authoritative Run state, while the native SSE transport now recovers from silent dead sockets.

Also added:
- `bridge/scripts/backend-audit-probe.mjs` / `npm run audit:probe`, a non-destructive real-machine probe for sanitized diagnostics and the five model catalogs;
- `docs/V3_BACKEND_AUDIT_REAL_HARNESS_VALIDATION.md`, the exact cold-PI, reconnect/background, Stop, duplicate-prompt, restart and capability-matrix release gate.

Latest automated head remains green for type-check/web regressions and bridge tests on macOS/Windows/Linux; the head Debug APK is being produced by CI. PR #288 remains DRAFT and must not merge before the documented real-harness gate.

giuliastro:
Real-machine gate update:

PI cold-start validation passed on the target machine. After a fresh daemon start, selecting PI first loaded the model catalog on the first attempt, with no disconnect and no switch-away/switch-back workaround. A simple PI Conversation prompt also completed successfully.

PR remains DRAFT. Next gate is Android background/foreground + short Wi-Fi interruption during an active Conversation, then diagnostics plateau verification.

giuliastro:
Real-machine model-list validation exposed another release-blocking bug before merge.

Harness Remote was showing removed OpenCode Zen models on OpenCode, PI and OMP even though the corresponding native harness pickers no longer showed them. Selecting a stale entry produced provider HTTP 400 `Model is unavailable`.

The audit branch now fixes the discovery source instead of blacklisting model names:
- ACP: do not reuse a persisted technical catalog Session as an authoritative model source across daemon restarts. Persisted ids are retained only so technical Sessions stay hidden. The current adapter lifetime discovers from one fresh prompt-less Session.
- OpenCode: prefer runtime provider inventory (`/provider`, then `/api/provider`) and use `/config/providers` only as compatibility fallback when runtime routes are unavailable.
- Existing model-specific variant probing remains runtime-advertised only.
- New regressions lock stale-session rejection and runtime-provider selection.

Head: `eaec2fac02edf7f2b1bc7259e1525c33c9a0513c`.

macOS/Windows bridge tests and Linux web/bridge checks are green; Debug APK is building. PR remains DRAFT and must not merge until the corrected model lists are validated against the native harnesses.

giuliastro:
Real-machine model-discovery regression diagnosed and corrected before merge.

The reported failures were two different UI paths:

- Existing PI Conversation: model loading failed with `spawn pi ENOENT`.
- New Conversation creation on other harnesses: model picker could remain loading for over a minute.

The regression was caused by a post-`eaec2fa` native PI/OMP model-filter experiment plus an existing timeout-contract flaw. The native filter spawned a second `pi --list-models` / `omp --list-models` process after ACP discovery, duplicating the model authority and adding a PATH dependency. That layer is removed entirely.

The corrected design is now:

- PI: `@automatalabs/pi-acp@0.5.0` ACP `configOptions` are the model source. No second `pi` process is spawned.
- OMP: `omp acp` config options are the model source. No second OMP process is spawned.
- Codex/Claude: their ACP adapter config options remain authoritative.
- OpenCode: runtime `/provider` or `/api/provider`, with `/config/providers` only as compatibility fallback.
- Persisted technical ACP Session ids stay hidden but their historical config is never loaded as current membership.

Timeout/liveness changes:

- ACP discovery now has one end-to-end 90s budget across adapter initialize/authenticate and technical Session creation.
- Optional model-specific variant probing is bounded to 10s total and 2s per probe. It cannot invalidate or indefinitely delay the base model list.
- diagnostics expose current discovery phase plus variant-probe total/completed/incomplete/error.

Conversation behavior changes:

- New Conversation creation is not blocked by model discovery. If the picker is still warming, Start uses the harness-native default.
- Opening an existing Conversation remains usable if its model catalog is slow or unavailable; model-catalog failure is no longer a general chat error.
- switching harness clears the prior harness model immediately.
- Continue uses an explicitly verified catalog model when available; otherwise it sends `model: null`, which means native default. This is intentionally different from omitting `model`, because the backend interprets omission as reuse of the prior Run model and could otherwise resurrect a removed provider model.

Test-context correction: the APK used for the regression report was from `eaec2fa`, while the daemon branch had already advanced six commits. Those later backend commits contained the native PI/OMP filter, so frontend and backend were not on the same SHA. The next manual gate will use matching APK/backend revisions.

Final current head: `7509db658e6422ab6dffbdda56174f225f0b0ce0`.

macOS/Windows bridge suites, Linux/web bridge suite, web build/type-check and regressions are green. The matching Debug APK is building. PR remains DRAFT and must not merge before the corrected real-harness model lists and Conversations are revalidated.

giuliastro:
Real-harness stale-model fix revised again after confirming the native CLI refresh surfaces.

Current head: `208f53ce5ba0af614aa446d5c75220b77c6bb188`.

PI and OMP model membership is now gated by their refreshed native inventories:
- PI: installed `pi` path -> `pi update --models` -> `pi --list-models`.
- OMP: installed `omp` path -> `omp models refresh --json`.
- ACP is metadata-only for defaults/config options and cannot reintroduce a model absent from the native inventory.
- Native refresh failure does not fall back to ACP-only membership.
- Regression covers the reported stale `deepseek-v4-flash-free` case directly.

Automated status on this head: Linux/web type-check, web build/regressions, Linux bridge tests, Windows bridge tests and macOS bridge tests all pass. Android debug artifact is still building.

Next gate is browser-first real-machine validation because this model-discovery bug is daemon/UI-shared and does not require Android transport to reproduce. PR remains DRAFT and is not to be merged.

giuliastro:
Stale-model root cause correction after real-machine reproduction

`deepseek-v4-flash-free` still appeared in TaskDesk for an existing PI Conversation and OMP New Conversation even though the native PI/OMP pickers no longer show it.

The previous assumption that a fresh ACP `session/new` was authoritative for PI/OMP model membership was incorrect. Both harnesses have dynamic provider catalogs whose native picker refresh path can be fresher than the ACP snapshot.

Head `9955ffa8f8dcbd59ef098f95728a6e0dd069f8a6` now uses:

- PI: absolute installed `pi`, `pi update --models`, then `pi --list-models` as membership authority.
- OMP: absolute installed `omp`, `omp models refresh --json` as membership authority.
- ACP: metadata/default/variant enrichment only, never permission to add a model absent from native live membership.
- Native refresh/list failure: fail closed, no fallback to ACP-only membership.

Regression coverage explicitly puts `deepseek-v4-flash-free` in stale ACP configOptions and verifies that it cannot reach the returned picker when native membership omits it. OMP native thinking efforts also remain authoritative over ACP variants.

Web/Linux, macOS and Windows jobs are green on this head. Browser real-machine validation is the next gate. The PR remains DRAFT and is not ready to merge.

Separate follow-up after this focused validation: make model discovery/cache Project-directory scoped end to end instead of using the daemon's first configured root.

giuliastro:
Superseded by the current 3.0 release candidate. The checkpoint contains this reliability work and subsequent fixes. Closing the draft; no code is being discarded.

Issue #290, opened 2026-08-23
Repository branch hygiene after Harness Remote 3.0 parallel audit


giuliastro:
## Goal

Reduce obsolete branch clutter after the Harness Remote 3.0 release without losing recovery points or unique diagnostics.

This is repository hygiene, not a product refactor.

## Current canonical baselines

- `main`: official Harness Remote 3.x stable line;
- tag `v3.0.0`: immutable official 3.0 release point;
- `checkpoint/v3-session-first-working-2026-08-25`: retained release-development checkpoint/reference;
- `archive/harness-3-2026-08-15`: protected historical recovery baseline.

The old pre-release survivor assumptions around `v3/taskdesk`, `feature/conversation-control-plane-rc1`, `fix/v3-backend-reliability-audit` and audit branches are no longer the canonical release topology. They may be deleted only after their unique commits are classified against current `main`/checkpoint/archive.

## Protected / historical keep

Never delete or rewrite without an explicit separate decision:

- `main`;
- `checkpoint/v3-session-first-working-2026-08-25`;
- `archive/harness-3-2026-08-15`;
- any `archive/*` or `backup/*` branch that still contains unique recovery material.

## Classification

For every remaining remote branch classify against **current `main` first**, then checkpoint/archive when historically appropriate:

1. **PROTECTED**: canonical/recovery branch or branch backing current open work.
2. **SAFE_TO_DELETE**: no open PR and no unique commits not already reachable from a surviving baseline.
3. **UNIQUE_WORK_REVIEW**: unique commits remain and need an explicit keep/delete decision.
4. **HISTORICAL_KEEP**: deliberate recovery/diagnostic snapshot.
5. **AMBIGUOUS**: ancestry or purpose is not yet proven.

Do not infer safety from branch names or closed PR state.

## Phase 1: final inventory

For each branch record:
- branch name;
- tip SHA;
- last commit date/message;
- related PR if identifiable;
- nearest surviving baseline;
- ahead/behind or unique commit count;
- classification;
- reason.

The earlier 144-branch audit was useful but its survivor set predates the final Session-first release. Refresh the inventory against current `main`.

## Phase 2: deletion

Delete only branches classified `SAFE_TO_DELETE`.

Never force-update refs to manufacture ancestry. Use normal remote deletion, then prune and publish the final remaining inventory.

## Current status

- Harness Remote 3.0 release: complete;
- obsolete open PR cleanup: complete;
[...]

Issue #289, opened 2026-08-23
Harness Remote 3.0 parallel pre-merge audit: capability contract + product/UI


giuliastro:
## Purpose

Run the final pre-merge audit in two independent workstreams from the same frozen baseline, then integrate only reviewed changes.

Frozen shared baseline:
- branch: `fix/v3-backend-reliability-audit`
- SHA: `56c5c9d925045fbef542650c6d564ce502c72758`
- PR #288 remains DRAFT

Do not modify the frozen baseline directly while the two audits are in progress.

## Product naming

The product is **Harness Remote 3.0**, not TaskDesk. The user model is conversation-first:

`Project -> Work Thread / Conversation`

A Work Thread is a persistent work conversation that can continue across harnesses. Do not redesign the product as a task manager. Existing internal `task*` names/routes may remain temporarily when renaming them would create risky churn; new user-facing language should prefer Conversation / Work Thread.

## Workstream A: capability contract / backend

Branch: `audit/v3-capability-contract`

Scope:
- make the per-harness capability matrix the real contract, not a normalized model list alone;
- include transport/protocol, model discovery source, tool-call representation, reasoning/variant metadata, Session lifecycle guarantees, reconnect/Stop/resume guarantees, and diagnostics where observable;
- make model/capability discovery correctly project/cwd scoped where harness configuration can be project-specific;
- cache/single-flight identity must include machine + harness + validated project/cwd where appropriate;
- carry project/cwd end to end from UI to daemon without accepting arbitrary filesystem paths outside configured roots;
- preserve the already-fixed reliability behavior from #287/#288;
- add regressions and diagnostics for first-selection races and cache isolation.

Avoid broad UI redesign in this branch.

## Workstream B: independent product/UI/UX audit

Branch: `audit/claude-v3-product-ui`

Start by reading in full:
- #197 canonical product direction;
- #287 backend/reliability audit;
- draft PR #286;
- draft PR #288 and latest discussion;
- current code at frozen baseline `56c5c9d925045fbef542650c6d564ce502c72758`.

Audit before changing code. Focus on:

### Product model and information architecture
- Is Project -> Work Thread / Conversation obvious and coherent?
- Remove TaskDesk/task-manager language from user-facing UI where it no longer matches the product.
- Preserve persistent Conversation continuity and harness switching.

### UX/UI quality
- desktop, web and Android layouts;
- typography, spacing, hierarchy, responsiveness and accessibility;
- New Conversation flow;
[...]

PR #291, opened 2026-08-23
audit(v3): independent product and UI review

giuliastro:
## Purpose

Independent Harness Remote 3.0 product/UI audit running in parallel with the final backend capability-contract workstream.

Coordination: #289
Handoff: `docs/V3_CLAUDE_PRODUCT_UI_AUDIT.md`
**Findings register: `docs/V3_PRODUCT_UI_AUDIT_FINDINGS.md`** — severity, status, terminology inventory, proposals not implemented.

Base is the frozen backend-audit baseline `56c5c9d925045fbef542650c6d564ce502c72758`.

This PR must remain DRAFT while the audit is in progress.

## How this was audited

Static reading of the live component tree, plus a real browser. The app was built, served, and driven in Chromium (Playwright) against a stub machine daemon serving five coding agents, two projects and three Conversations in working / failed / cancelled states. Measurements at 2560→320 CSS px in both themes. The stub and driver scripts are throwaway audit tooling and are not committed; the behaviours they exposed are pinned by regression suites instead.

## What the running shell actually is

Worth stating, because the tree does not make it obvious:

```
main.tsx → StandaloneUniversalWorkspace → ConversationWorkspace
         → ConversationDetail → WorkThreadConversation
         → TaskDeskConversation → TaskDeskMessageContent
```

Everything else under `web/src/components/`, plus `App.tsx`, is unreachable from the running app — ~12 900 lines. See finding P1.

## Fixes in this branch

| Commit | Sev | What |
|---|---|---|
| `875b9ed` | High | Composer keystrokes no longer `JSON.stringify` every Run, nor write `localStorage` synchronously |
| `5146661` | High | The 10s poll no longer rebuilds the object graph, flickers offline machines, or moves a Conversation backwards |
| `2c3db67` | Med | A missing model catalog is a resolved state, not a red failure |
| `b5c217c` | High | Escape, Tab trap and focus restoration for every dialog; `window.confirm` gone; no keyboard-over-form on Android |
| `f7032bb` | Med | Changes stops flickering; real ARIA tabs; inline rename instead of `window.prompt` |
| `2b50fa8` | Med | Pane separator survives `pointercancel` and is keyboard operable; New Conversation says *which* machine is blocked and why |
| `d6b197e` | Med | The composer is named for AT; touch devices are no longer told to press Ctrl |
| `bea15d9` | High | A disconnected machine is visible on a phone; attention surface announced; manifest copy |
| `36f1469` | High | Crash recovery clears the configuration 3.0 actually loads |
| `3373121` | High | A phone can reach "Needs attention" |
| `4ee49fd` | Med | A Conversation whose launch failed is not lost |
| `31f67d7` | — | The client-invented timeline role is renamed away from the old product |
| `2232c55` | High | The top bar action group no longer paints over the breadcrumb |
| `bc31256` | High | The closed conversation drawer is out of the tab order |
| `4ccaa8b` | High | 10px type floor, 76ch prose measure, activity label out of CSS, and a failed/cancelled Conversation no longer reports "Ready" |

### The four with hard measurements

**Top bar overlap (`2232c55`).** `minmax(360px, 1fr)` plus `min-width: 0` on `.tdw-top-actions` let the button row overflow its grid area leftward and draw on the breadcrumb. Measured: the machine-health pill overlapped the breadcrumb at **every width from 1200 to 1600** — the whole common laptop range — and at none of them after.

**Closed drawer in the tab order (`bc31256`).** `opacity: 0` + `pointer-events: none` hides a panel from the mouse only. **5 of 23 tab stops** on the default desktop screen landed inside it. 0 after; 5 again once it is open.

**Conversation state lied (`4ccaa8b`).** A failed or cancelled Conversation reported **"Ready"** in its header while its own list card said "Needs attention" or "Stopped" — the detail view was the one hiding the problem, and a cancelled Conversation's interruption was not visible anywhere. That is the direct opposite of #197's "a real interruption remains visible".

**Crash recovery cleared the wrong keys (`36f1469`).** The `ErrorBoundary` exists so a poisoned setting cannot brick the next launch — on Android, so a crash is not an unrecoverable black screen. Its reset list held only the 2.x profile keys; 3.0 boots from `harness-remote.workspace.machines.v1`, which was absent. A machine entry that crashed the render reproduced the crash on every launch while the button cleared keys nobody reads.

## Not implemented — decisions for you

1. **Delete the dead shells.** ~12 900 unreachable lines, and **12 of the 32 regression suites assert against them**, so green CI overstates confidence in the shipped shell. Two more test files are never executed by any npm script, and one of them **fails** today. Mechanical, but wants its own PR after #286/#288.
2. **Decide what Settings → Language means.** `i18n.ts` has four languages and ~300 keys; only the settings modal calls `t()`. Choosing Italian translates six labels. Either localize the shell (~80 new keys × 4 languages) or hide the control. Not removed unilaterally — it is a product call.
3. Compact the mobile Continue-with/Model toolbar (73px of permanent chrome on a phone), reduce the 145px empty composer, add a copy affordance to code blocks, replace the Android back-button `querySelector` coupling with explicit state, and note the minimum WebView version (`:has()`).

## Task / TaskDesk terminology

**No rendered string in the live shell says Task or TaskDesk** — pinned by a regression. One safe rename made: the synthetic timeline role `"taskdesk"` → `"conversation-event"` (client-invented, never crosses the harness boundary). Everything else is deliberately deferred with a stated reason: `/v1/tasks` routes and `taskClient`/`MachineTask` (backend contract, #287/#292), 89 `tdw-*` DOM classes and 32 `taskdesk-*` filenames (mechanical post-merge commit), and `harness-remote.taskdesk.draft.*` storage keys (need a migration, not a rename). Full table in the findings doc.

## Backend dependencies

**None.** No change to model discovery, capability/cache ownership, ACP/HTTP transport, native Session lifecycle, prompt idempotency, Stop/reconnect semantics, diagnostics or OpenCode event fanout. Two changes came close and were kept client-side; one observation is logged for #292 (`listAgentModels` single-flights but keeps no result cache, so every Conversation mount re-requests — a capability-contract call, not a UI one).

## Tests

11 new suites wired into `npm run test:v3-product-ui` and into `pr-checks.yml`, including 8 behavioural unit tests over the new `workspace-runtime-merge` module and a stylesheet floor check that fails if text under 10px is reintroduced. Three existing suites updated where the behaviour they protected moved.

Green on this head: `tsc -b`, `vite build`, and every suite in `pr-checks.yml`.

## Boundary

Do not independently redesign backend model discovery, transport, native Session lifecycle, idempotency, Stop/reconnect semantics or diagnostics owned by #287 and `audit/v3-capability-contract`.

## Safety

- never touch `main`;
- never modify `v3/taskdesk` directly;
- never delete or rewrite `archive/harness-3-2026-08-15`;
- do not modify the frozen baseline branch;
- do not merge this PR.

Refs #197
Refs #287
Refs #289
Refs #290
Refs #286
Refs #288


giuliastro:
## Audit progress 1 — read-through complete, first four fixes landed

Read in full before touching code: #197, #287 (including the six latest comments), #289, #286, #288, this PR and `docs/V3_CLAUDE_PRODUCT_UI_AUDIT.md`. Branch is based on the frozen baseline `56c5c9d925045fbef542650c6d564ce502c72758`; no backend contract owned by #287 / #292 has been touched.

### What the 3.0 shell actually is

For the record, because it is not obvious from the tree: the live entry point is `main.tsx` → `StandaloneUniversalWorkspace` → `ConversationWorkspace` → `ConversationDetail` → `WorkThreadConversation` → `TaskDeskConversation`. Everything else under `web/src/components/` (`universal-workspace.tsx`, `taskdesk-v3.tsx`, `taskdesk-v3-unified.tsx`, `taskdesk-workspace.tsx`, `taskdesk-home.tsx`, `work-thread-detail.tsx`) plus `App.tsx` (5 443 lines) is unreachable from the running app. That is ~11 k lines of unreferenced UI, and several of the existing regression suites still assert against those dead files. Documented below as a finding, not changed here.

### Fixes in this batch

| # | Severity | Area | Fix | Commit |
|---|---|---|---|---|
| 1 | **High** | Typing latency | Composer keystrokes no longer JSON-serialize every Run, and no longer write `localStorage` synchronously per character | `875b9ed` |
| 2 | **High** | Re-render / stale state | The 10s workspace poll no longer rebuilds the whole object graph, no longer flickers offline machines, and can no longer move a Conversation backwards | `5146661` |
| 3 | **Medium** | Model catalog UX | A missing catalog is presented as a resolved state, not a failure | `2c3db67` |
| 4 | **Medium** | Accessibility / mobile | Every dialog gets Escape, a Tab trap and focus restoration; `window.confirm` removed; no keyboard-over-form on Android | `b5c217c` |

**1. Typing latency (`875b9ed`).** `WorkThreadConversation` computed `taskConversationSignature(task)` — a `JSON.stringify` over every Run — unconditionally on every render, and the draft effect wrote `localStorage` on every keystroke. Both sat directly on the input path, so cost scaled with conversation length. The signature is now memoized on the Task object identity; the draft is debounced at 400 ms and flushed when the conversation is left, so nothing is lost. Storage failures no longer escape into typing. The "working" clock also kept a 1 s interval alive while the agent was idle.

**2. Workspace poll churn (`5146661`).** Every 10 s the poll replaced `runtimes` with freshly allocated objects even when the payload was byte-identical, so `conversations` / `projects` / `agents` all changed identity and the open Conversation, its toolbar and its transcript re-rendered on a loop. Same effect flipped an already-known-offline machine back to `loading`, so the sidebar cycled "Connecting…" → "Machine offline" forever. And a slow `listTasks` could overwrite a Conversation the open detail view had already reconciled — a visible Working → Ready → Working flicker. Reconciliation now lives in a pure, unit-tested `workspace-runtime-merge.ts`: identity is reused when a value is structurally unchanged, and a locally-known Conversation is kept only when it is *strictly newer* than the polled one. Server order stays authoritative and a newer server record always wins. This is client-side reconciliation only — no transport, polling-cadence or Session-lifecycle semantics changed.

**3. Model catalog presentation (`2c3db67`).** Two inconsistent treatments of the same recoverable condition. New Conversation surfaced a model-discovery failure through the modal's red `role="alert"` block, which directly contradicts the rule established in #287 that discovery is *not* an acceptance prerequisite — the conversation starts fine on the harness-native default. And `ModelPicker` disabled itself while still showing the "Choose model" placeholder whenever the catalog came back empty, which reads as a broken control with no explanation ("errors that leave the picker disabled or stuck", #287 §2). The picker now reads *Harness default / Chosen by the coding agent*, carries the discovery error as its tooltip and drops the chevron; the modal shows a quiet field note and keeps the red block for a real Start failure. Both surfaces now use the same wording.

**4. Dialog accessibility and mobile input (`b5c217c`).** The four dialogs (New Conversation, Settings, Machines, mobile Settings) were mouse-only: Escape did nothing, focus stayed behind the dialog so Tab kept walking the workspace underneath, and closing one dropped focus onto `<body>`. A shared `useDialogDismiss` hook now owns Escape, a Tab trap and focus restoration. Escape ownership is explicit: an open model picker inside a dialog consumes it first (previously Escape on the picker also threw away the filled-in New Conversation form), and a dialog on top no longer also collapses the conversation list behind it. Two mobile fixes in the same surfaces: removing a machine used `window.confirm`, which the Android WebView renders as a bare system alert over the app — now an inline Keep/Remove confirmation; and New Conversation autofocused its textarea unconditionally, raising the on-screen keyboard over the machine/project/agent/model selectors it sits below. Autofocus is now limited to fine pointers, with Ctrl/Cmd+Enter to start for keyboard users.

### Tests

New: `workspace-runtime-merge.test.mjs` (8 behavioural unit tests over the real merge functions), `v3-conversation-input-performance.test.mjs`, `v3-dialog-accessibility.test.mjs` — wired as `npm run test:v3-product-ui` and added to `pr-checks.yml`. `settings-regression.test.mjs` updated for the inline machine-removal confirmation.

Green locally on this head: `tsc -b`, `vite build`, and all of `test:i18n config ui settings model events profiles markdown attachments agent-runs mutation-coordinator catalog-guard machine-payload platform completion-audio completion-timing native-response v3-product-ui`.

### Still open — next batches

Working through: Changes-tab reload flicker, conversation-tab ARIA, `window.prompt` rename on Android, the pane resizer's `pointercancel` leak and missing keyboard support, "no coding machine is ready" not saying *why*, composer labelling and the misleading Ctrl/Cmd+Enter hint on touch devices, and the fact that the Settings **Language** control is effectively a no-op in the 3.0 shell. Full findings table, the Task/TaskDesk terminology inventory and any backend dependencies for #289 will follow in the next update.

No backend dependency identified so far. This PR stays DRAFT.

---

giuliastro:
## Audit progress 2 — browser-verified findings, and the register is up

The PR body is now the running summary and `docs/V3_PRODUCT_UI_AUDIT_FINDINGS.md` is the full register (severity, status, the commit that closed each finding, the terminology inventory, and the proposals I deliberately did not implement).

### Method change worth flagging

The second half of this audit was done in a real browser, not by reading source. I built the app, served it, and drove it in Chromium against a stub machine daemon serving five coding agents, two projects and three Conversations in working / failed / cancelled states, then measured at 2560 → 320 CSS px in both themes. The stub and the driver scripts are throwaway and are not committed; every behaviour they exposed is pinned by a regression suite instead.

That is what turned up the four findings below. None of them were visible from reading the code, and none would have been caught by the existing suites.

### What the browser found

**The top bar painted over itself at every common laptop width (`2232c55`).** `.tdw-top-actions` is `min-width: 0` with `justify-content: flex-end` inside a `minmax(360px, 1fr)` column, so the button row overflowed its own grid area to the *left*. Measured with box intersection: the machine-health pill overlapped the breadcrumb at **1200, 1280, 1366, 1440 and 1600** — and at none of them after making the column `auto`, which is what the ≤1180px rule below it already did. No width scrolls horizontally.

**22% of the desktop tab journey was invisible (`bc31256`).** The conversation drawer is hidden with `opacity: 0` + `pointer-events: none`, which hides it from the mouse only. Driving actual Tab presses: **5 of 23 stops** landed inside it — its close button, the search field, the empty-state New conversation button, the resizer. 0 after adding `visibility`; 5 again once the drawer is open, so nothing was lost.

**The detail view contradicted the list about failed and cancelled Conversations (`4ccaa8b`).** Driving the stub's failed and cancelled Conversations: the list card said "Needs attention" and "Stopped", the header pill said **"Ready"** for both. The pill only knew working / pending-question / ready. For the cancelled one the interruption was not visible anywhere in the detail view — the direct opposite of #197's "a real interruption remains visible". It now derives from the Conversation status and carries the harness error as its tooltip.

**Readability, measured rather than asserted (`4ccaa8b`, `89d4040`).** Over a third of the visible desktop text rendered below 10px, including the daemon's own connection error at 8.5px. Contrast was already fine — 7.4:1 to 16:1 in dark, 4.9:1 to 18:1 in light — so this is purely size, and every declaration under 10px in the fifteen stylesheets the running shell loads is now at 10px. Agent prose ran to a 991px line, ~146 average characters against a comfortable 45–90; capped at 62ch, which measures 611px and ~90. User messages are excluded: they are short and framed by their own bubble. A stylesheet floor check fails CI if sub-10px text is reintroduced.

### Also in this batch

`36f1469` crash recovery cleared only the 2.x profile keys while 3.0 boots from `harness-remote.workspace.machines.v1` — a poisoned machine entry reproduced the crash on every launch while the button cleared keys nobody reads. `3373121` gives the phone a way to reach "Needs attention", which was unreachable because the filter section is `display: none` below 780px. `4ee49fd` stops losing a Conversation whose `createTask` succeeded but whose `launch` failed. `31f67d7` renames the one TaskDesk identifier that carries no protocol risk.

### Two things I am not deciding for you

**The dead shells.** `App.tsx` and thirteen other modules — ~12 900 lines — are unreachable from `main.tsx`, and **12 of the 32 regression suites assert against them**. Two more test files are never executed by any npm script, and `taskdesk-focus-layout.test.mjs` **fails today** against a heading structure `taskdesk-workspace.tsx` no longer has. Green CI therefore overstates confidence in the shipped shell by a meaningful margin. The cleanup is mechanical but conflicts with anything in flight against those files, so it wants its own PR after #286 and #288.

**Settings → Language.** `i18n.ts` carries four languages and ~300 keys; only the settings modal calls `t()`. Choosing Italian translates six labels and leaves the product in English. Either localize the shell — roughly 80 new keys across four languages, real translation work — or hide the control. I did not remove it unilaterally because that is a product call, not a bug fix.

### Backend

Still no dependency. Nothing here touched model discovery, capability or cache ownership, ACP/HTTP transport, native Session lifecycle, prompt idempotency, Stop/reconnect semantics, diagnostics or OpenCode event fanout. One observation logged for #292 rather than acted on: `taskClient.listAgentModels` single-flights in-flight requests but keeps no result cache, so every Conversation mount re-requests its agent's catalog. Whether that should be cached, and on what key given the project/cwd scoping agreed in #287, is a capability-contract decision.

### Tests

11 new suites under `npm run test:v3-product-ui`, wired into `pr-checks.yml`, including 8 behavioural unit tests over the new `workspace-runtime-merge` module. Three existing suites updated where the behaviour they protected moved rather than disappeared. Green locally on this head: `tsc -b`, `vite build`, and every suite in `pr-checks.yml`.

Still DRAFT.

---

giuliastro:
## Audit progress 3 — two late findings, and one of my own fixes was wrong

### A select stopped looking like a select (`4c69ca6`)

The base `select` rule in `styles.css` draws the dropdown chevron with `background-image`. Three v3 rules set the `background` **shorthand**, which resets `background-image` to `none`. Result: **Machine, Project, Coding agent, Theme, Language and Continue with all rendered as plain boxes**, visually indistinguishable from text inputs — while the Model control sitting immediately beside them kept its chevron, so the New Conversation form had two different affordances for the same kind of choice.

What identifies this as an unintended regression rather than a design decision: `.tdw-agent-control select` still reserves `padding: 0 25px 0 8px` — right padding for an arrow it had stopped drawing.

Fixed by using `background-color` in the three rules, so the base `background-image` survives. Confirmed in the browser: computed `background-image` goes from `none` to the chevron gradients.

### My own Escape-ownership fix did not work (`173858b`)

Worth calling out plainly, because it is a lesson about the shape of this repo's test suite.

In `b5c217c` I added Escape ownership so that dismissing the model picker would not also throw away the New Conversation form behind it. I asserted the guard in a regression test, the test passed, and I moved on. It was wrong.

`NewConversationModal` renders one of two dialogs and therefore calls `useDialogDismiss` **twice**. The branch that is not rendered has a null ref — and that unused instance still registered a document Escape listener with no container to check, so it closed the dialog regardless of what the rendered one decided. The static assertion could not see this, because the guard really was in the source; it just never ran.

Only a keyboard walkthrough in a real browser found it. After the fix, driving actual key presses:

- Enter on the launcher opens the dialog and focus lands in the first field;
- **0 of 24 Tab presses** leave the dialog;
- Escape closes it and returns focus to the button that opened it;
- Escape on an open model picker closes **only** the picker, and the typed draft survives;
- a second Escape closes the dialog.

The general point, which matters beyond this one bug: this repository's regressions are source-text assertions. They are cheap, fast and genuinely useful for pinning intent, but they verify that code *was written*, not that it *works* — and they cannot distinguish a guard that runs from a guard that is merely present. That is worth remembering when reading the green checkmark on any of these suites, mine included.

## Status

21 commits on the branch, 15 of them fixes. Findings register at `docs/V3_PRODUCT_UI_AUDIT_FINDINGS.md`, running summary in the PR body.

CI on head `173858b`: Type-check and regressions ✅, Bridge tests on Ubuntu / macOS / Windows ✅, Application menu on macOS ✅. The Debug APK job is still building at the time of writing — every earlier head on this branch built it successfully, and nothing in this batch touches Android-specific code, but it is worth a glance before the integration review.

Still DRAFT. Still no backend dependency.

---

giuliastro:
CI is fully green on head `173858b`: Type-check and regressions ✅, Bridge tests on Ubuntu / macOS / Windows ✅, Application menu on macOS ✅, Debug APK ✅. The APK note in my previous comment is resolved.

---

giuliastro:
Integration update: the browser-audited product/UI changes in #291 are accepted into the isolated Harness Remote 3.0 pre-merge candidate. I did not merge this source PR into a survivor branch.

The combined candidate keeps the #291 responsive/focus/performance/state-fidelity fixes and overlays #292's backend/capability work. In the two shared model-discovery call sites, the integration explicitly preserves #292 scope: New Conversation uses `{ projectId: projectID }`, and an existing Conversation uses `{ workThreadId: task.id }`.

DRAFT #294 validates the combined candidate. Dead-shell deletion, broad internal TaskDesk/task renaming and full localization remain intentionally outside this stabilization candidate.

giuliastro:
Audit work is already incorporated into the surviving 3.0 line and is fully behind the current release candidate. Closing the draft as completed/superseded.

PR #300, opened 2026-08-24
fix(bridge): prevent duplicate last message on OMP refresh (#299)

### LINKED ISSUE #299: Repeated message in the Android app
I have a harness-remote server on my MBP. When interacting with OMPnvia the server, whenever I go from the servers view to the server, the last message I have in the session gets duplicated.

Baylar55:
Fixes #299

**Bug:** Navigating servers view -> server on Android (and any client with ?refresh=1) duplicated the trailing assistant message on every visit. Screenshots showed Yes--we're in the pfBlockerNG repository growing from 5->6 copies, each with distinct Thought for ... duration.

**Root cause:**
- HARNESS_PROFILES.omp relies on leaf-gated omp-session-history.js:69 (activeSessionLeaf===undefined -> []) falling back to ACP session/load replay - replay is intentional, so reloadOnHistoryRefresh must stay true (unlike pi/codex).
- messageSignature (bridge/src/acp-service.js:68) role+text join included ephemeral reasoning/tool text. Live streaming (agent_thought_chunk -> reasoning{time:{start:now}} + tool_call) vs persisted/replayed history therefore diverged though visible text was identical.
- mergeReplay LCS zero-tie at acp-service.js:154 common[l+1][r] >= common[l][r+1] (0>=0) emitted [...prefix, live, replay] -> duplicate. join empty erases part boundaries, so split vs glued text must stay equal - separator must be empty (verified via repro: 3 msgs, dup=true with newline).

**Fix (bridge/src/acp-service.js):**
- messageSignature -> visible text only (filter type==text) + error, join empty preserved. Ignores reasoning/tool/time/ids - matches semanticMessagePart:81 intent but keeps mergeReplay lenient (strict semanticMessageSignature would worsen).
- Tail reconciliation after LCS: if remaining tails are single assistant with same visible sig, keep one; if tails same-length and role+visible-sig pairwise equal, keep previous (belt).
- healPoisonedSnapshot - dedup consecutive assistant visible duplicates. Wired in mergeReplay (entry/exit), #restoreSnapshot (heals mergeFragmentedPiSnapshot + marks dirty for lazy rewrite) and #persistSnapshot (never persist duplicate). Existing poisoned snapshots self-heal on next open.
- 247/247 bridge tests pass (see Tests commit); preserves user-separated legitimate repeats (healer checks role==assistant only).

**Trade-off documented:** consecutive identical assistant envelopes are assumed impossible under ACP (every turn begins with user_message_chunk), so healing is safe.

**Verification:** end-to-end repro scenarios B/D no longer duplicate; poisoned [user,live,replay] heals to one.

**Tests (3ab7937):**
- mergeReplay unit cases in external-history-dedup.test.js: live reply with streamed reasoning vs text-only replay -> one copy (thought preserved); journal-split text parts vs glued replay part -> equal (guards the empty separator); failed turn stays distinct from an identical successful reply; poisoned snapshot input heals; user-separated repeated replies never collapse.
- Differential test reference signature updated to visible-text+error so it no longer passes vacuously against the old all-parts semantics.
- End-to-end reopen test via ThoughtfulOmpAcp fake: live turn streams agent_thought_chunk, session/load replays text only -> reopening must not duplicate and must keep the thought.
- Mutation-checked: the 4 bug-dependent tests fail against 41dfe88 (pre-fix), including the end-to-end one at "reopening must not duplicate"; F-case/repeat guards pass on both by design.

### DIFF
--- a/bridge/src/acp-service.js
+++ b/bridge/src/acp-service.js
@@ -66,7 +66,32 @@ function mergeFragmentedPiSnapshot(messages) {
 }
 
 function messageSignature(message) {
-  return `${message?.info?.role ?? ""}\u0000${(message?.parts ?? []).map((part) => part?.text ?? "").join("")}`
+  // Visible text only – reasoning/tool/file are ephemeral between live stream and persisted/replayed
+  // history. Including them made live vs replayed assistant turns with same answer diverge and get
+  // appended as duplicates on every refresh (see omp-session-history.js:69 fallback).
+  const visibleText = (message?.parts ?? [])
+    .filter((part) => part?.type === "text")
+    .map((part) => part?.text ?? "")
+    .join("")
+  const error = message?.info?.error?.message ? `\u0000${message.info.error.message}` : ""
+  return `${message?.info?.role ?? ""}\u0000${visibleText}${error}`
+}
+
+function isConsecutiveAssistantDuplicate(prev, next) {
+  return prev?.info?.role === "assistant"
+    && next?.info?.role === "assistant"
+    && messageSignature(prev) === messageSignature(next)
+}
+
+function healPoisonedSnapshot(messages) {
+  if (!Array.isArray(messages) || messages.length < 2) return messages
+  const healed = []
+  for (const msg of messages) {
+    const prev = healed.at(-1)
+    if (prev && isConsecutiveAssistantDuplicate(prev, msg)) continue
+    healed.push(msg)
+  }
+  return healed.length === messages.length ? messages : healed
 }
 function stableSemanticValue(value) {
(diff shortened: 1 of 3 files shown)


giuliastro (approved):
Reviewed the fix against #299 and the current main baseline. The root cause and reconciliation strategy are consistent with the observed OMP refresh behavior, and the added regression coverage exercises live reasoning vs replayed text, poisoned snapshots, legitimate repeated replies, failed turns, and end-to-end reopen behavior. CI is green across bridge tests, web regressions, Windows/macOS, and the Android debug build. Looks good to merge. Thanks for the thorough fix and tests!

PR #301, opened 2026-08-25
v3: redesign the Session-first workbench UI

gervaso-assistant:
## Summary

Redesigns the Harness Remote 3.0 Session-first surface as a native coding-agent workbench rather than a Task dashboard or generic chat UI.

The product hierarchy remains:

```text
Machine -> Project -> Native Session
```

Native Sessions still open through the existing HR3 controller and the single shared `WorkThreadConversation -> TaskDeskConversation` renderer. This PR does not add a second transcript, composer, reasoning model, timeline, or lifecycle implementation.

### Session Home

- explicit Machine -> Project -> Session hierarchy;
- scalable search plus All / Live / Attention filters;
- harness filter with real per-agent counts;
- collapsible Projects with selected-Session auto-expand;
- selected Session highlighting;
- harness identity, external origin, Working / Needs attention / Stopped / Ready and relative time;
- native parent/child Session tree when `parentID` is present, with cycle and missing-parent safety;
- bounded accessible labels for pathological historical titles;
- compact native diff summaries (`+N -N · N files`) without transcript fanout;
- Project grouping sourced from the machine Project catalog, with native directory fallback;
- responsive New Session flow grouped by Machine and Project.

### Session detail and conversation

- compact Machine / Project / harness / native-agent context header;
- native agent mode and safe permission-policy summary without exposing rule patterns;
- token, cost and changed-file metadata when the harness reports them;
- mobile-focused full-screen Session detail;
- persistent Ready / Working / Needs attention / Stopped state on mobile;
- Home state mirrors the already-authoritative selected Session controller immediately, avoiding a stale Ready row while the detail is Working;
- professional coding-tool transcript styling while preserving the mature renderer, paging, Activity, tool, reasoning, copy and scrolling behavior;
- composer and Send / Stop controls remain inside mobile safe areas.

### Model picker

- provider/model/variant search retained;
- catalog-confirmed Free filter;
- Free / Paid, context and pricing metadata;
- rendering bounded to 100 base model families with search guidance for larger catalogs;
- mobile safe-area sheet and 44px touch targets.

## Real-harness validation

Validated against the running local machine daemon with real native data:

- OMP: 25-26 Sessions, 3,198 catalog options;
- PI: 15-16 Sessions, 2,156 catalog options;
- Codex: 6 Sessions, 35 catalog options;
- OpenCode: 79-80 Sessions, 865 catalog options;
- 125 historical Sessions on initial load, 127 after dedicated PI/OpenCode validation Sessions;
[...]

PR #298, opened 2026-08-24
v3: Session-first native discovery and safe control path

giuliastro:
## Scope

Session-first implementation slice for #297, built on the validated #294 candidate.

This PR keeps native coding-agent Sessions as the primary resource. It does not replace the mature HR3 chat renderer and does not make Task, Conversation or Run part of the Session-first product model.

### Product model

`Machine -> Project -> Native Session`

A Session discovered from Codex, PI, OMP, Claude or OpenCode remains that harness-owned native Session. Harness Remote can observe it and, where the harness permits, deliberately claim and continue that exact Session. No replacement Session is created for same-Session continuation.

### What it adds

- per-agent native Session discovery through the machine daemon;
- global/paginated Session listing first, stable Session listing fallback;
- status enrichment that cannot make Session discovery fail;
- isolation between harnesses so one broken adapter does not hide Sessions from the others;
- composite Session identity using machine + agent + native Session id + directory;
- Project-first compact Session navigation with bounded 5-row disclosure;
- the existing `TaskDeskConversation` transcript/composer reused directly for native Sessions;
- existing HR3 transcript paging, message merging, live refresh and foreground recovery reused rather than reimplemented;
- explicit observe -> claim -> write semantics for ACP Sessions;
- ACP lightweight discovery never infers writer ownership from missing metadata;
- dedicated native Session claim endpoint, separate from the model catalog API;
- writer-lock refusal keeps the Session observe-only;
- successful ACP claim retains ownership so the first prompt does not load/acquire the Session twice;
- durable native Session prompt idempotency using `clientRequestId` and a machine-scoped operation ledger;
- pending operations persisted before native dispatch and accepted operations persisted before HTTP success;
- sequential and concurrent retries converge on one native prompt;
- ambiguous delivery is retained as uncertain and is never blindly replayed after restart;
- unresolved prompt request ids survive WebView reload;
- live lightweight Session status reconciliation in the open Session view;
- capability-gated native Stop using the existing HR3 Stop control;
- ACP Stop allowed only after an explicit successful Session claim;
- OpenCode Stop routed to its native abort endpoint;
- Stop uses the same durable operation ledger, including concurrent retry deduplication and uncertain-delivery handling;
- Stop idempotency is scoped to the active user turn so an old accepted Stop cannot swallow a future turn's Stop;
- read-only per-agent native model discovery through the daemon without claiming the Session;
- current native model and variant restored when reported by the harness;
- model and variant included in the durable prompt mutation identity so retries cannot silently change model or effort;
[...]

Issue #297, opened 2026-08-24
Harness Remote 3.0 Session-first UX migration without 3.0 regressions


giuliastro:
## Purpose

Define and implement the Session-first product model from #197 without regressing the backend, ACP, model, transport, Android or UI reliability work already completed in the 3.0 line.

This is **not** a rollback to the 2.x implementation.

The 2.x application is reference material for the familiar native-Session mental model and for useful session listing/history behaviors only. The implementation base must remain the stabilized 3.0 line.

## Non-negotiable rule

> Session-first is a product-model/UI migration on top of Harness Remote 3.0, not a port of Harness Remote 2.x.

Any 2.x code or behavior must be individually justified before reuse. If 3.0 has a newer or safer implementation, the 3.0 implementation wins.

## 3.0 reliability floor that must not regress

The Session-first work must preserve at minimum:

### ACP and harness startup

- installed ACP binaries on PATH remain preferred;
- pinned package fallbacks remain correct:
  - PI: `@automatalabs/pi-acp@0.5.0` / `pi-acp`;
  - Claude: `@agentclientprotocol/claude-agent-acp@0.63.0` / `claude-agent-acp`;
  - Codex: `@agentclientprotocol/codex-acp@1.1.14` / `codex-acp`;
- OMP remains on its ACP path;
- OpenCode remains on the managed HTTP path;
- transient lazy-start failure must remain retryable.

### Model discovery

- machine + harness scoped model state;
- OpenCode runtime provider inventory as primary source, with legacy config fallback only where required;
- fresh ACP technical Session per adapter lifetime for catalog authority rather than reusing stale historical catalog Sessions;
- daemon-owned single-flight discovery;
- bounded loading/polling and total ACP discovery budget;
- optional variant probing must not make base models unusable;
- adapter-native model controls/variants remain preserved, including OMP thinking, PI thinkingLevel, Codex reasoning_effort and OpenCode provider-advertised variants;
- no fabricated Claude reasoning levels;
- model state is cleared on harness/machine switch so stale choices cannot leak;
- model catalog failure must not make an otherwise usable native Session unavailable;
- when no verified explicit model is available, the harness-native default is preferred over a stale persisted model id.

### Transport and lifecycle

- OpenCode managed global event fan-out remains one daemon-owned upstream subscription for downstream clients;
- reconnect/background churn must not create unbounded listeners/subscriptions;
- Android native SSE stall watchdog and reconnect/backoff behavior remain intact;
[...]

PR #294, opened 2026-08-23
integration(v3): pre-merge candidate validation

giuliastro:
## Harness Remote 3.0 integrated pre-merge candidate

This DRAFT PR validates the integrated Harness Remote 3.0 candidate against the frozen backend reliability baseline. It is a validation surface, not an authorization to merge.

Candidate branch: `integration/v3-premerge-candidate-2026-08-23`

Frozen base: `56c5c9d925045fbef542650c6d564ce502c72758`

Product model: `Project -> Conversation -> Native Sessions`

## Stabilized backend/model behavior

The ACP model catalog remains intentionally machine + harness scoped during this stabilization. Project/cwd is not the authority for ACP catalog membership. OpenCode continues to use its runtime provider inventory.

The earlier real-machine failure where PI model discovery returned `ACP adapter exited (127) ... @automatalabs/pi-acp@0.5.0: not found` was fixed by invoking pinned scoped packages through their actual published binaries:

- PI: `npx --yes --package=@automatalabs/pi-acp@0.5.0 pi-acp`
- Claude: `npx --yes --package=@agentclientprotocol/claude-agent-acp@0.63.0 claude-agent-acp`
- Codex: `npx --yes --package=@agentclientprotocol/codex-acp@1.1.14 codex-acp`

Installed adapter binaries on PATH are still preferred. OMP remains `omp acp`; OpenCode remains on its managed HTTP path. Exact fallback-command regressions cover PI, Claude and Codex.

## Mobile and conversation UI stabilization

Real-device and browser feedback has now been folded into the integration branch.

### PR #295

Merged mobile keyboard/status polish:

- portrait Android no longer falls into short-landscape styling when the soft keyboard resizes the WebView;
- New Conversation stays anchored and usable with the keyboard open;
- mobile Refresh is reachable;
- Host/Username disable unwanted autocapitalization/autocorrect;
- harness state labels/colors are consistent and accessible;
- search and composer keyboard hints are corrected.

### Android background/foreground recovery

A real-device Claude test exposed a lifecycle gap: the native SSE reader can keep consuming while Android suspends the WebView, leaving the renderer stale when the app returns even though the native Session continued and completed correctly.

The shared live-refresh layer now listens to Android `appStateChange` plus `visibilitychange` / `pageshow`, deduplicates them, and forces an immediate authoritative Conversation + transcript + attention reconciliation on foreground. This is harness-agnostic and applies to Claude, PI, Codex, OMP and OpenCode paths that use the shared Conversation surface.

### PR #296

[...]

PR #302, opened 2026-08-25
v3: fix the Session-first model lifecycle by reusing the stabilized v3 backend

giuliastro:
Draft. Do not merge.

Fixes the release-blocking model-change failure class on the Session-first line by routing the native Session through the conversation-first backend contracts from #288 / #294 instead of the parallel implementation that grew around it. No rewrite, no second catalog, no second lifecycle manager. The UI integrated by #301 is untouched.

Base: `feature/v3-session-first-read-path` @ `54f38f0ed4bdc1f40b955730fef98fe895930a7e`.

## The reasoned diff

Comparing this tree with `56c5c9d9` (#288) and `f76c67b1` (#294): the mature backend is almost entirely intact. `agent-model-catalog.js`, `taskClient.ts`, `api.ts` and the ACP load/prompt/queue machinery are byte-identical to #294. The regression is not in that backend — it is in the thin Session-first layer added on top, which re-implemented four things the mature backend already did correctly.

**1. The variant was applied before the model, so it was silently discarded.**

`promptSession` issued its own `session/set_config_option` for the variant, then called `AcpService.prompt`, which applies the model. A harness legitimately resets dependent controls when the model changes, so the variant was thrown away. `TaskLauncher.resumeSession` in #294 already had the correct sequence — `setModel` first, variant after — so the variant now travels through `AcpService.setModel`/`prompt`. That is the one place which already waits for real `configOptions`, orders the model before its variant, and defers both to *dequeue* rather than mutating the configuration of a turn that is still running. The daemon no longer sends raw adapter requests of its own.

Measured on a real daemon over real ACP stdio. Before — requested `thinking=high`:

```
t=15  set_config_option thinking=high
t=16  set_config_option model=claude-opus-4-8   <- resets thinking to off
t=17  session/prompt        -> SERVED thinking=off
```

After:

```
t=15  set_config_option model=claude-opus-4-8
t=16  set_config_option thinking=high
t=17  session/prompt        -> SERVED thinking=high
```

**2. One refused prompt bricked the Session permanently — the reported headline symptom.**

The durable pending-delivery record exists so a retry after a lost response converges on the same daemon ledger entry instead of duplicating a turn. It was also kept when the daemon had *definitively* refused the request, and it never expired. Since the guard compares the stored model against the requested one, **a model change is exactly what makes the next request differ** — so one refused prompt made every later prompt fail, across reloads, because the record lives in `localStorage`.

The claim server answers 202 for the genuinely ambiguous cases and only throws when the mutation was never dispatched, so a real HTTP status is proof of refusal: the record is cleared then. A transport failure still leaves delivery unknown, is still remembered, and retrying the same prompt still reuses the same request id. Ambiguous records now also expire, so the native transcript rather than a stale browser record has the last word.

**3. Sending waited out the catalog's whole discovery budget.**

Resolving an explicit model had no bound of its own, so a cold ACP adapter could hold the prompt request for the catalog's full 90s. From the client that is indistinguishable from a wedged Session. Sending now reuses the ceiling the model route already applies to a client-facing request (8s), then delivers the requested model without variant enrichment. Discovery keeps its full budget and stays owned by the daemon's single-flight catalog, so the degradation is transient: measured 8.0s instead of 25s against a deliberately slow adapter, with the following prompt applying the variant in 8ms once warm. No timeout was raised to hide anything.

Relatedly, model discovery failure no longer makes a usable Session unusable. Only an authoritative `model_unavailable` answer about the user's explicit choice stays a conflict.

**4. Shared client state was rewritten globally, and stale daemon state outlived its cause.**

- The native-Session adapter reassigned `taskClient.listAgentModels` for every consumer in the app in order to strip a synthetic scope back off. The daemon owns one catalog per machine and harness — its capability contract says `cacheScope: machine` — so the surface now passes that scope explicitly through a new `WorkThreadConversation` prop. The Work Thread scope remains the default for Task-backed conversations, and the catalog effect keys on the scope's *value* rather than a caller's object identity.
- Writer claims were remembered across an adapter exit, even though ownership lives inside one live adapter process. Stop then skipped the reload it needs and failed against a Session the new process had never opened. Claims are dropped when that agent's adapter exits, and only for that agent.
- Opening a Session waited for model enrichment before mounting the controller, so a slow read left the surface on `Loading Session into the v3 controller...` indefinitely. The transcript mounts first; the recovered model refines it afterwards and cannot overwrite a selection the user has since made.

## Diagnostics

Only what was missing. Pending ACP requests named their Session for prompts alone, so a stuck `session/load` or `session/set_config_option` was indistinguishable from anything else — every session-scoped request now reports its Session and `configId`. The inactivity watchdog still applies to prompts alone, so no timeout was extended. The daemon also reports Session-first control-plane state: current writer claims, and a bounded credential-free view of unresolved ledger mutations with the age of the oldest, since an unresolved entry is what blocks a client from sending a new prompt. No credentials, tokens or prompt bodies.

## Validation

Two kinds of evidence, reported separately. **The real-harness run is the authoritative one.**

### Real local harness validation

Harnesses actually installed and exercised on the test machine:

| Harness | Transport | What was exercised |
|---|---|---|
| **Claude Code** (`claude-agent-acp` 0.63.0) | ACP stdio | Full end-to-end. Catalog (4 models: `default`/`sonnet`/`opus`/`haiku`), 2 concurrent Sessions in 2 Projects, 11 real prompts with real assistant replies, 3 model changes per Session, 5 cross-harness cycles, Stop, resume after Stop |
| **OpenCode** (`opencode-ai`) | Managed HTTP | Full control plane. Lazy start, runtime `/provider` inventory (582 models, 3 connected providers), Session create, prompt accept, transcript read, native error propagation. Model **inference** is blocked by this container's network policy (`opencode.ai` not allowlisted) and surfaced correctly as a native message error without breaking the Session |
| **PI**, **OMP**, **Codex** | — | **Not installed on this machine, therefore NOT validated.** The OMP *code path* (profile, `thinking` variant config, history loader) was exercised against a protocol-faithful ACP peer over real stdio, which is instrumentation, not harness validation |

The new `bridge/scripts/session-first-soak.mjs` runs the reproduction from #287 directly — `open A -> catalog -> change model -> send -> open B -> change model -> switch harness -> return A -> change model -> send` — and passed **58/58 checks** on Claude + OpenCode: catalogs stayed per harness with no shared model across all 5 cycles (claude=4, opencode=582 every time), every prompt was answered by the Session it was addressed to via echoed markers, one user turn per prompt with no duplicates, Stop left the interrupted turn visible and the Session still writable.

Diagnostics before / after that run:

| | before | after |
|---|---|---|
| claude adapter listeners | 7 (stopped) | 8 (running) — stable 8→8 across cycles |
| claude catalog adapter listeners | 1 | 1 |
| opencode listeners | 2 | 2 |
| pending ACP requests | 0 | 0 |
| in-flight Session loads | 0 | 0 |
| queued prompts | 0 | 0 |
| subscribers | 1 | 1 |
| transcript cache | 0 entries | 2 entries, 0 evictions |
| ledger | 0 | 14 accepted, 0 pending, 0 uncertain, **0 unresolved** |

Same probe against the OMP profile: ALL CHECKS PASSED, listeners exactly stable (7→7, 8→8, 1→1), 26 turns served, **zero** model/variant ordering violations, and the variant verified as reaching the harness.

PI, OMP and Codex still need to be run on a machine that has them. The probe is parameterized for exactly that: `HR_PRIMARY=pi HR_SECONDARY=opencode node bridge/scripts/session-first-soak.mjs`.

### Automated validation

All six PR checks are green on head `a0e0e4c`: Type-check and regressions, Bridge tests (macOS), Bridge tests (Windows), Application menu on macOS, Chromium product smoke, Debug APK.

Also run directly on this tree before pushing: bridge 401/401 on Linux, web type-check and production build, all 23 web regression commands from the `pr-checks` list, and all 7 Chromium product smokes including the new model-switch smoke.

### Tests added

- `bridge/test/acp-service-model-variant.test.js` — model applied before its variant; a variant the adapter never advertised is refused; a prompt queued behind a running turn defers *both* to dequeue
- `bridge/test/native-session-claim-lifecycle.test.js` — an adapter exit drops that agent's claims and only that agent's; the ledger reports unresolved mutations without exposing signatures
- `bridge/test/native-session-model-routing.test.js` — the corrected ordering contract; a Session stays usable when discovery fails for a non-catalog reason; a model the catalog says is gone is still rejected; sending does not wait out a cold catalog
- `web/src/native-session-model-lifecycle.test.mjs` — a refused delivery does not block the next prompt after a model change; an ambiguous one still does and still reuses its request id; ambiguous records expire; a wedged PI Session does not make an OMP Session look blocked
- `web/scripts/native-session-model-switch-smoke.mjs` — the full sequence in Chromium against the production build, plus a refused delivery followed by a model change and a reload

The browser smoke has teeth: with the pending-delivery fix reverted it fails at exactly the reported point — `Timed out waiting for prompt HTTP request for "B-SEND-AFTER-REFUSAL"`, i.e. no HTTP request is even attempted after the model change.

## Known gaps

- PI, OMP and Codex are unvalidated here. They are the remaining real-harness gate, and the reason this is not yet a release candidate.
- OpenCode inference could not be exercised in this environment, only its control plane.
- OMP performance was deliberately left alone, per #287: correctness first, then measurement, then a separate performance change.
- The `native-session-v3-adapter` still patches `api.loadMessagePage`, `getWorkThread`, `continueTask` and `cancelWorkThread` globally. Only the `listAgentModels` patch — the one that altered mature semantics — was removed. The rest stays thin and transitory as intended, but it is the natural next reduction.
- The installed `claude-agent-acp` 0.63.0 does advertise an `effort` config option, which the `claude` profile does not list as a model variant. That looks like a real, honest capability worth adding, but it is a product change rather than a lifecycle fix, so it was left out of this PR.

Refs #197
Refs #287
Refs #297
Refs #298

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_019icjiT3JyP84HtUz4itqA2


giuliastro:
## Real PI now exercised, and it found two more defects

The previous round could not test PI, OMP or Codex. PI and Codex turned out to be installable from npm at the versions the profiles pin, so `resolveLaunchPlan` detects them and the daemon runs them for real. New head `f2e9157`.

Detected harnesses on the test machine: `["pi","claude","codex","opencode"]`.

**Real PI works fully** — `@automatalabs/pi-acp@0.5.0`, authenticated through `pi-stored-credentials`, 114 base models plus per-model `thinkingLevel` variants for 632 catalog entries, cold discovery in 1.6s and warm in 2-6ms.

### Two defects only real PI could reveal

**1. A variant valid for the chosen model was refused.** PI advertises a *different* `thinkingLevel` range per model. Measured directly against the adapter across a 9-model sample: `meta.llama4-maverick` offers only `["off"]`, `nova-2-lite` stops at `high`, `anthropic.claude-opus-4-7` goes to `max`. Nine of nine sampled models differed from the session default, so per-model probing is genuinely required — I had hypothesised the control was session-scoped and the experiment refuted it, which is why the existing probe loop was left alone.

`AcpService.setModel` discarded the `configOptions` the adapter returns when the model changes, so the Session kept describing the *previous* model and the variant was validated against the wrong range. A real soak run rejected `zai.glm-4.7-flash` + `thinkingLevel=high` as HTTP 500. The Session now adopts the options reported for the model it holds; a level the current model genuinely does not offer is refused before it reaches the adapter, names the values that model does offer, and is reported as a 409 conflict about the user's choice rather than a server fault.

**2. Switching between Sessions re-read the harness for no reason.** The transcript cache is bounded by a 24MB memory budget *and* by an entry cap. The cap was 8 — sized for conversation-first, where a user works in a handful of Conversations. Session-first inverts that: Home lists every native Session and invites hopping between them, so opening a ninth evicted the first, and going back re-read that harness's journal or re-ran `session/load`. Eviction also clears `#loaded`, so the reload is a full one.

Measured on real PI: **46 evictions across 12 Sessions while the cache held 39KB of its 24MB budget.** The entry cap was doing all the evicting with the memory bound essentially unused. The memory budget now governs, the entry cap sits above ordinary navigation, and evictions record which bound forced them so cap thrash is visible in diagnostics instead of looking like ordinary cache pressure.

### Diagnosability for "the model list is unavailable"

Discovery now reports its phase while still loading, and a discovery timeout names that phase, its source and the last error. Demonstrated live: `claude models=0 loading=true phase=creating-session`. Previously every case looked identical, so a harness that will not start was indistinguishable from discovery that is merely slow.

### Harness isolation, verified

An unauthenticated Codex fails in **0.3s** with a clear reason (`spawn xdg-open ENOENT` — it wanted a browser for ChatGPT auth) and, immediately afterwards, PI still returned 632 models and OpenCode 582. One broken harness does not contaminate the others.

### Real local harness validation

| Harness | Result |
|---|---|
| **PI** (`pi-acp` 0.5.0, ACP) | Soak **68/68**: 2 Sessions, 3 model changes each, 5 cross-harness cycles with OpenCode, a `thinkingLevel=max` turn, Stop, and all listeners/in-flight/ledger state back to baseline. Control plane only — this machine's bedrock credentials are rejected for inference, so turns end in a native error, which the probe now reports explicitly instead of counting as a pass |
| **Claude Code** (ACP) | Soak **67/67** with real inference and echoed-marker routing proof, after these changes |
| **OpenCode** (managed HTTP) | Full control plane, 582 runtime models; inference blocked by this container's network policy |
| **Codex** (`codex-acp` 1.1.14) | Starts and initializes; cannot authenticate here (no API key, and ChatGPT auth needs a browser). Verified only that its failure is fast, clear and isolated |
| **OMP** | Still not validated. There is no public `omp` CLI package that provides the `omp acp` entry point the profile expects, so it cannot be installed here |

The probe gained `HR_ALLOW_TURN_ERRORS=1` so a reachable harness whose provider will not serve inference can still be soaked — and the summary then states plainly that it covered the control plane only, with a count of the turns that ended in a native error.

### Automated validation

bridge **405/405**, web type-check and production build, all 23 `pr-checks` web regressions, all 7 Chromium smokes. One smoke (`native-opencode-browser-smoke`) failed once inside a back-to-back run and passed both in isolation and on a clean re-run of the full sequence — its preview server uses a fixed port that the previous script had not yet released.

### Still open

OMP remains the one harness with no coverage. If it is installed under a different command on your machine, `--acp-command`/`--acp-arg` will point the OMP profile at it and the soak will run against it unchanged:

```
HR_PRIMARY=omp HR_SECONDARY=pi HR_ALLOW_TURN_ERRORS=1 node bridge/scripts/session-first-soak.mjs
```

---

giuliastro:
## OMP session lookup, and the final measured state

Head `b2ec7ff`.

**OMP was walking its session tree once per Session opened.** `locateSession` recursively walks the OMP session tree, and that walk already enumerates every Session file — but only the single file being looked for was kept. Each new Session therefore paid its own full walk, so a machine with a lot of OMP history spent O(Sessions) tree walks finding files it had already seen, and every successive Session open was slower than the last.

The listing the walk produced is now retained and searched in memory; only a miss against a listing older than a second walks the tree again, so a Session created moments ago is still found. Files are matched by suffix rather than by recovering an id from a file name, because OMP Session ids may contain underscores.

This is deliberately **not** the previously reverted OMP index experiment: nothing is persisted, no second source of truth appears, and the data is exactly what the existing walk already read. The new regression measures it: 40 Sessions share one walk instead of taking one each, and a late-created Session is still found. The loader now also reports its walk count, listing size and resolved-Session count through the ACP service diagnostics, so "opening Sessions gets slower" is measurable rather than only felt.

OMP itself is still unvalidated end to end — there is no public `omp` CLI package providing the `omp acp` entry point — so this is a fix to the OMP code path backed by a unit regression, not a real-OMP run.

### Final measured state, one daemon, real PI + Claude + OpenCode + Codex

```
pi        proc=running  listeners=8  pending=0  catModels=632  phase=ready
claude    proc=running  listeners=8  pending=0  catModels=4    phase=ready
codex     proc=stopped  listeners=3  pending=0  catModels=0    phase=idle
opencode  proc=running  listeners=2  pending=0  catModels=582
```

Transcript cache after the full soak on each harness — the number the cache change is about:

| | entries | evictions | weightEvictions | hits | misses | maxEntries |
|---|---|---|---|---|---|---|
| pi | 8 | **0** | 0 | 48 | **0** | 64 |
| claude | 2 | **0** | 0 | 102 | **0** | 64 |

The same shape of workload previously produced 46 evictions on PI while the cache held 39KB of its 24MB budget. Every revisit is now a cache hit, so no Session is re-read from its harness for want of a cache slot.

Operation ledger: 30 accepted, 0 pending, 0 uncertain, **0 unresolved**. No pending ACP requests, no in-flight loads, no queued prompts on any harness.

Soaks on this exact tree: **PI 68/68**, **Claude 67/67 with real inference and echoed-marker routing proof**. bridge **409/409**.

---

giuliastro:
This backend-reuse fix branch is fully contained in the current release candidate. Closing the obsolete draft.

PR #306, opened 2026-08-26
Session-first: finalize transcript convergence and stop false model-change notices

giuliastro:
## Scope

Follow-up to draft PR #304, based exactly on its current head `5615f4d136f60e54ea364fa4680b600fe9947899`.

This PR addresses the two remaining real-harness regressions without touching `main`, `v3/taskdesk`, the bridge protocol, or adding a second renderer.

### 1. Mounted transcript can remain shorter than durable native history

- Keeps the newest-page merge monotonic across **live ACP -> native journal identity swaps**, not only same-message-id updates.
- Stabilization is deliberately limited to the final assistant reply of the same final user turn, with text/prefix agreement and no tools/errors, so unrelated/repeated historical replies cannot be collapsed.
- Treats ACP `session.updated` as a selected-transcript lifecycle edge as well as index/detail state.
- Reuses #304's existing coalesced tail refresher and bounded lifecycle settle read; no permanent fast polling is added.

This targets the known PI/OMP live-to-JSONL boundary and also gives every ACP-backed harness a final authoritative tail reconciliation.

### 2. False `Model changed` when continuing an existing Native Session

A Native Session projection currently carries recovered/requested model metadata but **does not persist provenance proving that the user explicitly changed the picker**. Adjacent `Run.model` comparison therefore confuses model/default enrichment with user intent.

The timeline now fails closed for Native Session projections: it does not fabricate a `Model changed` lifecycle line from model metadata alone. Actual model selection/sending is unchanged. Legacy Task-backed Runs retain their established model-change event behavior.

A future Native Session model-change annotation should only be restored once explicit user-choice provenance is carried by the projection; it must never be inferred again from enrichment.

## Regression coverage

- same-id stale journal prefix cannot shorten a streamed answer
- different live/journal assistant IDs cannot shorten or duplicate the final reply
- cross-ID stabilization is scoped to the same final user turn
- ACP `session.updated` must refresh selected transcript and schedule one settle read
- Native Session model metadata differences cannot fabricate a model-change lifecycle line

## Validation gate

Keep this PR **DRAFT / UNMERGED** until real-harness validation is complete, especially PI, OMP and OpenCode, followed by Codex and Claude smoke tests.


giuliastro:
Superseded by the later checkpoint work and current release candidate. This divergent experimental branch is retained only as history and is no longer a merge candidate.

PR #307, opened 2026-08-26
Claude: clear stale Working when no prompt is in flight

giuliastro:
Regression follow-up to #305. Real-device testing showed the previous fix preserved `busy` when Claude had no live `session/prompt`, which is exactly the stale historical-Session case. This PR starts with a regression that reproduces that condition, then will keep `Working` only while Claude actually has an in-flight prompt. No changes to main, v3/taskdesk, PI, OMP, Codex or OpenCode.


giuliastro:
## Proposed fix for what this PR does not reach

Real-device feedback: the Sessions are not the ones stuck on `Working` — **individual Activity sections inside finished Claude Sessions are**, and only some of them. That symptom cannot be fixed by `corroborateClaudeSessionStatus` at all, because an Activity section's status is never derived from the Session status.

### Root cause

`web/src/conversation-parts.ts` computes an Activity section's status purely from the parts inside it:

```ts
if (parts.some((part) => part.type === "tool" && part.state?.status && part.state.status !== "completed" && part.state.status !== "error")) return "running"
if (parts.some((part) => part.type === "reasoning" && part.time?.start && !part.time.end)) return "running"
```

and `taskdesk-message-content.tsx` prints `running` as **`Working`**. So one part the bridge never closes pins that one section on `Working` forever — which is exactly why only *some* sections are affected. Two independent producers in `bridge/src/acp-service.js` leave parts open:

1. **Tool calls with no closing `tool_call_update`.** The `tool_call_update` branch is the only thing that ever moves a tool part off `running`. A call still open when the turn ends, is cancelled or fails — and a call replayed out of `session/load` history — keeps `state.status: "running"` for good, and it is written into the snapshot, so it survives restarts and reopens identically every time.
2. **Reasoning that a tool call interrupts.** Only the message-chunk path stamps `time.end` on a preceding reasoning part (`previous?.type === "reasoning" && partType !== "reasoning"`). The `tool_call` branch pushes its part without that, so *thinking → tool call* — Claude's ordinary shape — leaves a reasoning part with a start and no end forever. On its own that is enough to keep the section on `Working`.

Both survive `#active` being cleared, both survive the snapshot, and neither is visible at Session level: hence "the Session reads idle but the section still says Working".

### Proposed change — branch `claude/claude-sessions-activity-sections-30lmzf`

Bridge (source of truth):
- `settleUnfinishedActivity()` closes a Session's open tool calls and reasoning once the turn behind them is no longer live: end of turn, `abort`, snapshot restore, after a history replay, and when the public status layer concludes the Session is idle while the service still holds its busy flag (`settleReportedIdleActivity`, wired into `publicSessionStatus`).
- A call with no reported outcome closes as `incomplete` — not `completed` (a success that never arrived) and not `error` (accusing a tool that most likely ran). A later `tool_call_update` still overwrites the part, so settling can never lose a real outcome.
- Historical reasoning closes at its own `start`, so a snapshot reopened days later does not bill the outage as thinking time.
- The `tool_call` branch now closes a preceding open reasoning part, which also fixes the duration shown for it.

Web (presentation, robust on its own):
- `activityStatus` reads live tool states from an allow-list (`pending`, `running`, `in_progress`) instead of "anything that is not completed or error", so no unknown or settled state can hold a section on `Working` again.
- `incomplete` gets its own presentation in both conversation UIs: a closed row with no result, rather than the running ellipsis or the failure mark.

Tests: 425 pass in `bridge` (10 new, including the Claude adapter that opens a tool call it never closes, abort mid-call, replayed history, and a snapshot written mid-turn) and 11 in `conversation-parts`. Every runnable web suite passes; `test:native-session-home` needs `vite`, which is not installed in that environment.

### On the change already in this PR

I kept it rather than reverting: it does clear a genuinely stale `#active` flag, and reverting would risk bringing back the Session-level symptom you observed on device. Two things to be aware of, neither introduced by the fix above:

- While a queued prompt waits between turns, `service.status()` is busy but there is no in-flight `session/prompt`, so the Session now reads **idle** with work still outstanding.
- `#startTurn` emits `session.updated` just before the ACP request is registered, so a status read in that window also reports idle.

Both are transient presentation flickers. If you want them gone, the narrow fix is to let the corroboration see the queue (busy-because-queued is not a stale flag) rather than to loosen the staleness rule again. Say the word and I'll open this branch as a PR into `fix/v3-claude-stale-working-status-v2-2026-08-26`, or push it straight onto that branch.

No changes to PI, OMP, Codex or OpenCode behaviour beyond the shared `AcpService` settling, which is correct for every harness: nothing a finished turn started is still running.

---

PR #308, opened 2026-08-26
fix(session-first): make OMP continuation follow native branch truth

giuliastro:
## Scope

OMP-only Session-first reliability follow-up on the **final merged head of #304** (`7dddd1c9edc6fdba1fb32bce8c4fc43c00febb71`).

This PR does **not** target `main`, does not modify `v3/taskdesk`, and does not add a second chat renderer or a parallel Session model. It remains **DRAFT / UNMERGED** until real OMP validation passes.

## Failures reproduced during real validation

The first pass fixed the normal continuation path, but real-device testing isolated additional failures specifically after OMP errors/interruption:

1. normal continuation keeps the actual model instead of regressing to `Harness default`;
2. every continuation gets a fresh Activity boundary instead of marking the previous completed Activity `Working` again;
3. normal leave/reopen follows the same active JSONL branch;
4. **after one or more red failed turns**, leave/reopen can lose those errors, recover the wrong model, or show strange/duplicate replies;
5. the mounted OMP reply can occasionally stop a few words early, while leave/reopen immediately reveals the complete final text;
6. changing Session while `Oh My Pi is getting started` can restart that presentation timer. This is recorded but intentionally not changed here until it is proven OMP-specific rather than a shared mount/presentation issue.

### 1. Production OMP must never guess the active branch

OMP history is a tree. `inferLatestTerminalLeaf()` is useful only for standalone/history-only fallback and can choose a newer abandoned sibling after a failed/retried turn.

The original #308 made **paged** production reads require extension-provided `activeSessionLeaf`, but its full-history fallback still called `inferLatestTerminalLeaf()` when the extension state was temporarily unavailable. That contradicted the intended architecture and explains why error/interruption cases could reopen on a different terminal sibling.

The same OMP loader object now treats `pageRequiresActiveLeaf = true` as a strict contract for **both full loads and pages**. If authoritative extension state is unavailable, it returns no guessed branch and lets `AcpService` retain/retry its already-known observational state instead of replacing it with an arbitrary sibling.

Standalone loader consumers keep the previous inference fallback.

### 2. Failed attempts remain visible without resurrecting abandoned answers

A retry can create a successful assistant sibling of an earlier failed assistant for the same user node. Following only the selected ancestry made the red failure disappear on reopen; merging every sibling would instead recreate duplicate answers.

OMP history now linearizes exactly:

- the authoritative selected branch; plus
- assistant records with a real `errorMessage` whose parent user prompt belongs to that branch and whose journal record existed no later than the selected leaf.

Normal assistant siblings are never added. Later abandoned error/success siblings beyond the selected leaf are also excluded. Journal append order is retained, so visible history remains `prompt -> failed attempt -> retry/success`.

### 3. Model continuity survives a failed visible attempt

OMP `MessagePage.model` remains native truth for the projection. Model reconstruction now also accepts provider/model metadata from a visible failed assistant attempt tied to the selected prompt. This covers the real case where a failure used the selected model but the later successful journal envelope omitted provider/model metadata; reopen must not fall back to an older/default model just because the first attempt failed.

While an HR-originated OMP turn is live, a lagging journal page still cannot overwrite a concrete current selection with the previous branch model. Late model enrichment still fills a current Run whose model was unknown.

### 4. Continuation / Activity boundary remains established before ACP mutation

A logical projection Run is created **before** writer acquisition and prompt delivery. OMP can emit `session.updated` during claim/model/prompt setup; without the provisional Run that event can reactivate the preceding completed Activity.

The provisional Run is promoted to the durable daemon `clientRequestId` after acceptance. A definite refusal removes it; an ambiguous delivery preserves the durable pending identity without claiming the turn is still Working.

### 5. OMP final-tail convergence is bounded but longer than one read

The previous #308 did an immediate transcript read plus one settle read at 900 ms after `session.updated`. Real validation proved that OMP's JSONL can still lag that second read, producing a reply whose final words appear only after navigation forces a fresh read.

Other harnesses are unchanged: they keep the existing single 900 ms settle.

Only OMP now gets a finite convergence sequence after the latest lifecycle edge: **350 ms, 900 ms, 1.8 s, 3.2 s**, in addition to the immediate selected-tail read. A newer lifecycle event coalesces/restarts the sequence; leaving the selected Session cancels the old chain by Session identity. There is no permanent fast polling.

## Regression coverage added

- production OMP full history and paging do not infer a terminal sibling when `activeSessionLeaf` is unavailable;
- a failed sibling attached to the selected user prompt remains visible after reopen;
- abandoned successful siblings are not resurrected;
- abandoned errors journalled after the selected leaf are not resurrected;
- the OMP newest page keeps `prompt -> red failure -> selected final answer`;
- model recovery can use the visible failed attempt when the selected final envelope lacks model metadata;
- existing authoritative-branch pagination still keeps opaque cursors and exact limits;
- OMP alone receives the bounded multi-read lifecycle convergence window;
- a settle sequence captured for Session A cannot refresh Session B after navigation.

Existing #308 coverage remains for model reconciliation, provisional Run ordering, live -> journal assistant identity, native prompt idempotency and OMP active-leaf paging.

## CI status for the latest follow-up

The previous #308 head passed the full PR workflow. The new recovery commits are on head `1bf5b530819e578ce67e4030bac8b651f29b5ee4`.

GitHub did not dispatch a new Actions run for these connector-authored synchronize/reopen events, even after temporarily toggling the PR Ready and pushing a test-only comment. Do **not** treat the old green run as validation of this new head. Real OMP validation remains mandatory before merge.

## Real validation gate

Keep this PR **DRAFT / UNMERGED** until all of these pass on a real OMP Session:

- continue an existing clean OMP Session without changing model: picker remains on the actual model, never `Harness default`;
- continue repeatedly: every user turn creates its own Activity boundary and an older completed Activity never flips back to Working;
- deliberately produce a red failed turn, then retry successfully: the red error and the one selected successful answer both remain visible;
- after that failure + success, leave/reopen 3-4 times: no additional replies appear, no error disappears, model remains the actual last model;
- repeat with two failures before a successful retry;
- change model explicitly, fail once, retry successfully, leave/reopen: the selected model remains correct;
- wait for a long answer to finish without leaving the Session: its final words must appear by themselves and must match what is shown after reopen;
- Stop once, leave/reopen, then continue once;
- only after OMP passes, smoke PI, OpenCode, Codex and Claude for non-regression.

### DIFF
--- a/bridge/src/harness-profiles.js
+++ b/bridge/src/harness-profiles.js
@@ -122,6 +122,7 @@ export const HARNESS_PROFILES = {
     // explicitly and invoke its published binary instead of relying on npx package-spec inference.
     args: ["--yes", "--package=@agentclientprotocol/codex-acp@1.1.14", "codex-acp"],
     adapterCommand: "codex-acp",
+    permissionMode: "allow",
     // The adapter offers `api-key` before `chat-gpt`; the former demands CODEX_API_KEY or
     // OPENAI_API_KEY, while a `codex login` leaves ChatGPT credentials the `chat-gpt` method
     // reads from disk. Prefer the login, exactly like the generic default already avoids
--- a/bridge/src/harness-profiles.js
+++ b/bridge/src/harness-profiles.js
@@ -18,6 +18,12 @@ const COMMON_CAPABILITIES = {
   sessionDelete: false
 }
 
+const OMP_HISTORY_LOADER = createOmpHistoryLoader()
+// OMP's JSONL file is a tree, not a linear transcript. The newest terminal leaf can belong to an
+// abandoned sibling branch, so paging must ask the OMP extension for the authoritative active leaf
+// before choosing a branch. AcpService falls back to its full load path if that state is unavailable.
+OMP_HISTORY_LOADER.pageRequiresActiveLeaf = true
+
 export const HARNESS_PROFILES = {
   omp: {
     id: "omp",
@@ -25,7 +31,7 @@ export const HARNESS_PROFILES = {
     command: "omp",
     args: ["acp"],
     permissionMode: "allow",
-    historyLoader: createOmpHistoryLoader(),
+    historyLoader: OMP_HISTORY_LOADER,
     // OMP exposes thinking as a real ACP config option. We probe only ids the running adapter
     // actually advertises; this list is a routing hint, never a source of invented variants.
     modelVariantConfigIDs: ["thinking"],
@@ -116,7 +122,6 @@ export const HARNESS_PROFILES = {
     // explicitly and invoke its published binary instead of relying on npx package-spec inference.
     args: ["--yes", "--package=@agentclientprotocol/codex-acp@1.1.14", "codex-acp"],
     adapterCommand: "codex-acp",
-    permissionMode: "allow",
     // The adapter offers `api-key` before `chat-gpt`; the former demands CODEX_API_KEY or
     // OPENAI_API_KEY, while a `codex login` leaves ChatGPT credentials the `chat-gpt` method
     // reads from disk. Prefer the login, exactly like the generic default already avoids
--- a/bridge/src/omp-session-history.js
+++ b/bridge/src/omp-session-history.js
@@ -1,10 +1,8 @@
 import { createReadStream } from "node:fs"
-import { open, readdir } from "node:fs/promises"
+import { readdir } from "node:fs/promises"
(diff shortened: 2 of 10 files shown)


giuliastro:
Superseded by the OMP rebuild merged in #314 and the current release candidate. This divergent experimental branch is retained only as diagnostic history and is no longer a merge candidate.

PR #313, opened 2026-08-26
fix(v3): make OMP native Session read/resume reliable without undo extension

giuliastro:
## Scope

OMP reliability fixes on top of `checkpoint/v3-session-first-working-2026-08-25`. `main` is untouched. PI is the lifecycle baseline: OMP follows the same ordinary ACP Session flow and diverges only for OMP-specific branching/actions.

## Root causes found against real OMP 18.x

- OMP 18.x journals start with a fixed-width `title` slot and a logical `type:"session"` header. The old loader treated the session header as a branch node because it has an `id`, creating a fake terminal leaf and making normal linear Sessions look ambiguous.
- OMP creates a fresh Session JSONL lazily. A cached negative directory lookup could therefore miss the file created by the first prompt and keep the first completed answer invisible until reopen.
- OMP persisted branch state was incorrectly treated as ambiguous whenever the JSONL contained sibling entries. OMP itself rebuilds its Session index in file order and selects the last non-header entry as the persisted leaf, so read-time ACP replay was both unnecessary and stateful.
- Harness Remote kept OMP titles only in its own snapshot while Session-first discovery reads OMP native metadata. OMP 18.x exposes `/rename` over ACP, so create-with-name and manual rename now persist through OMP itself and survive discovery/reopen.
- a selected model was being sent as a mutation again on every prompt. When the requested provider/model/variant already matches native Session state, the prompt now omits the redundant model mutation, avoiding a false same-model change notification.
- old OMP v1 journals have no `id` / `parentId` tree until OMP migrates them on native load. Harness Remote previously discarded those entries, which made old Sessions appear empty until a prompt accidentally caused native migration.
- OMP JSONL was not marked authoritative after writer acquisition, so the UI could keep serving a shorter ACP live cache even after the complete assistant reply was already durable on disk.

## Read / resume behavior

- existing OMP Sessions do not depend on `omp-undo-redo` to become readable;
- title/session metadata are excluded from the parentId branch graph;
- persisted OMP branch selection follows OMP's own `SessionEntryIndex.rebuild()` semantics: the last non-header journal entry is the persisted leaf;
- no transcript read uses native ACP `session/load`; sibling branches, failed attempts, model metadata and paging are resolved directly from JSONL;
- old v1 journals are migrated in memory for read purposes only, with stable Harness Remote message ids that survive OMP's later native v1 -> v3 rewrite;
- the optional undo/redo extension may refine the current live leaf when available, but it is never a prerequisite for reading;
- OMP model recovery stays journal-native; it never forces a long native `session/load` merely to populate the model picker;
- a recovered OMP model is written back to the open projection and latest historical Run, so continuing after reopen with the same model neither mutates OMP nor emits a fake `Model changed` event;
- the last verified native model is retained as lightweight per-Session memory across UI unmount/remount, so returning to a still-working OMP Session cannot transiently show `Harness default` just because busy list/page metadata omitted the model;
- OMP JSONL remains transcript authority while the Session is working and after Harness Remote owns the writer. Live ACP output is merged only as transient enrichment;
- when the journal already contains the complete final assistant while ACP still holds a shorter prefix, the complete persisted tail wins immediately without a duplicate;
- writer acquisition uses OMP's native ACP `session/resume`, which opens the stored Session without replaying its whole transcript. `session/load` is no longer part of ordinary OMP read/resume/send;
- a transient JSONL read failure fails closed on the observational read path: Harness Remote keeps the cached transcript/error signal and never falls back to `session/load`;
- the shared Working -> Ready tail refresh remains as reconciliation, but correctness no longer depends on UI timing because the bridge itself has the correct transcript authority.

## Rename / create

- OMP profile uses native ACP `/rename` and prefers OMP-listed titles;
- a newly created OMP Session with an explicit name immediately persists that name through the native rename path because `session/new` itself has no title parameter;
- rename does not edit OMP JSONL by hand.

## Second-turn live projection fix

- Harness Remote records an accepted prompt immediately with a temporary browser/bridge identity, while OMP persists the same user turn under its journal id before the assistant is durable.
- The newest-page merge previously kept both ids when that transition happened before an assistant existed. The work-thread timeline then saw two identical user turns: the accepted Run bound to the first orphan while OMP reply/activity bound to the second, so the current turn looked blank or stayed on starting until remount.
- The current persisted user turn now keeps the already-visible prompt identity while it is still non-terminal. Once a prior turn is terminal, assistant compatibility is required so genuinely repeated prompts are not collapsed.
- Persisted OMP assistant records now carry durable completion metadata because OMP writes them to JSONL only at message_end.
- The native v3 projection uses that completed OMP transcript envelope to clear forced Working for the exact current prompt, without timers or navigation.

## Regression coverage

- realistic OMP 18.x title slot + session header fixture;
- lazy JSONL creation after an earlier negative lookup;
- first and second ordinary linear turns stay journal-native without `session/load`;
- sibling branches follow OMP's persisted last-entry leaf with zero ACP replay;
- v1 Sessions are readable before native migration and keep stable public message ids across a v1 -> v3 rewrite;
- an HTTP transcript read performs zero `session/load` and zero `session/resume`; the first writer claim performs exactly one `session/resume` and still zero `session/load`;
- native OMP `/rename` persistence;
- extension absent and stale-extension cases;
- error + retry/success sibling history;
- stable branch paging and model recovery;
- live ACP id -> JSONL convergence without duplicate/cut tail, including a complete journal assistant replacing a shorter live ACP prefix;
- selected-tail event refresh plus one lifecycle-driven Working -> Ready persisted-tail read;
- reopen + continue with the recovered same model sends no model mutation and keeps adjacent Run model metadata equal;
- leave + return while OMP is still busy restores the known model even when transient native metadata omits it;
- long OMP model enrichment does not call the session-scoped model endpoint / force ACP replay.

The frozen checkpoint also contained stale UI/smoke assertions that no longer matched its already-frozen product code, including old scroll preservation, old Project grouping, pre-machine-scoped New Session expectations, and an OpenCode create selector that treated the Project select as the harness select. These were aligned in tests/smokes only; no unrelated product behavior was changed to satisfy them.

Keep this PR in draft until validated against a real OMP installation.


giuliastro:
Superseded by #314, which rebuilt the OMP ACP path cleanly from the checkpoint and passed the real OMP validation. This draft is no longer a merge candidate.

PR #317, opened 2026-08-27
Add LRU eviction and lightweight listing to prevent session memory leaks

Baylar55:
This fix addresses an unbounded memory accumulation issue in the bridge during long running daemon sessions or when handling prompts with large image attachments.

What was happening:
Each call to listSessions was loading and parsing full session snapshots from disk into memory for every session. Transcripts containing large base64 image data URLs stayed in memory permanently because the messages and todos maps never evicted entries. Additionally, unlisted or deleted sessions continued to linger in the session maps.

What this changes:
1. Switched session listing to only restore lightweight metadata so full message transcripts are no longer loaded into heap just to display the list.
2. Introduced an LRU cache for in memory session transcripts with a default cap of 8, configurable through the MAX_CACHED_TRANSCRIPTS environment variable or the CLI flag.
3. Active, queued, loading, or replaying sessions are pinned in memory so they are never dropped mid turn.
4. Evicted sessions seamlessly reload their transcript from disk snapshots on demand.
5. Deleted sessions clear transcript arrays from their tombstone files on disk to prevent storage leaks.
6. Stale owned sessions that disappear out of band from the daemon are pruned after 5 consecutive polls, cleaning up their state and disk files.

All 256 test cases in the test suite pass cleanly, including new test coverage for eviction, pinning, rehydration, and pruning.

### DIFF
--- a/bridge/src/acp-service.js
+++ b/bridge/src/acp-service.js
@@ -1,5 +1,5 @@
 import { randomUUID } from "node:crypto"
-import { mkdir, readFile, rename, writeFile } from "node:fs/promises"
+import { mkdir, readFile, rename, rm, writeFile } from "node:fs/promises"
 import path from "node:path"
 import {
   listExtensionActions,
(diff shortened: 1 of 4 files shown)


giuliastro:
Thanks Baylar55! I reviewed the changes and CI is fully green across bridge tests, type-check/regressions, macOS/Windows, and the debug APK. The memory-management fix is well scoped and has solid regression coverage. Merged, thank you!

Issue #330, opened 2026-08-28
V3 architecture simplification: make Native Session runtime explicit


giuliastro:
## Goal

Post-release architecture cleanup for Harness Remote 3.x.

The baseline is **current `main` / 3.x**. Do not revive the pre-release draft branches.

## Released state

The original highest-risk ambiguity has already been removed:

- #331 introduced the explicit `ConversationController` boundary.
- Native Sessions no longer monkey-patch shared `api` / `taskClient` methods.
- The released code uses `ConversationRuntime` / `ConversationTurn` as the neutral UI runtime.
- Native Sessions are not projected into `MachineTask/MachineTaskRun` to render the mature chat.
- The released invariant is:

```text
Native Session -> ConversationRuntime/ConversationTurn -> mature chat/timeline
```

The old draft #333 is closed as superseded because this invariant is already in the released code and that branch predates later PI, OMP, handoff and deletion fixes.

The old draft #332 is also closed as superseded. Its proposed standalone reconciliation module predates the final production PI live/journal and OMP live/replay behavior and must not be merged as-is.

## Remaining work

These are non-release-blocking 3.x maintainability improvements:

1. extract current harness-specific reconciliation rules from the production adapter into a deterministic, well-tested domain without changing behavior;
2. split large React controllers into controller/state and presentation components;
3. consolidate layered global CSS overrides into scoped styles/design tokens;
4. reduce legacy Task-backed compatibility only where current reachability proves it is unused;
5. replace brittle source-text regression guards with behavioral contract tests where practical;
6. keep per-harness capability differences explicit rather than moving them into generic UI conditionals.

## Constraints

- start from current `main`;
- preserve current 3.x behavioral contracts;
- no broad rewrite;
- no transcript semantic changes disguised as refactoring;
- every slice must have focused regression coverage and be independently reviewable;
- real-device behavior wins over green tests if they disagree.

This is continuous technical debt, not a release gate.

Refs #197
Refs #331
Refs #332
Refs #333

giuliastro:
## Phase 1 complete

PR #331 merged into `checkpoint/v3-session-first-working-2026-08-25` as `bb71ab075c83146c66f288fb95a97995ca0a1b00`.

Result:
- Native Session mount no longer monkey-patches shared `api` / `taskClient` methods.
- `WorkThreadConversation` now has an explicit `ConversationController` I/O boundary.
- Native Sessions inject a Session-scoped controller.
[...]

PR #332, opened 2026-08-28
refactor(v3): isolate native Session reconciliation semantics

giuliastro:
Second slice of #330, built on the explicit controller boundary from #331.

### What changes
- moves PI live/journal identity stabilization into `native-session-reconciliation.ts`;
- moves OpenCode transcript-completion matching into the same pure reconciliation domain;
- moves native handoff prompt normalization and transcript model authority there;
- keeps `native-session-v3-adapter.ts` responsible for projection state, writer ownership and applying state transitions only;
- adds focused behavioral tests for repeated PI answers, repeated OpenCode prompts, terminal metadata, handoff normalization and backend-scoped model authority;
- updates older source guards to point at the module that now owns those invariants.

### Why
Harness-specific reconciliation rules were mixed into the generic Session projection controller. That forced an agent fixing PI/OpenCode/model behavior to edit the same file that owns generic Session state and mutations. Separating pure semantics makes the change surface smaller and makes regressions easier to localize.

### Non-goals
- no protocol/ACP changes;
- no UI changes;
- no native Session lifecycle changes;
- no Task/Run removal yet;
- no changes to `main`.

Base: `checkpoint/v3-session-first-working-2026-08-25`.

Refs #330
Refs #197


giuliastro:
Closing as superseded by the final Harness Remote 3.0 stabilization.

The intent of this draft was to isolate PI/OpenCode/native-session reconciliation into a dedicated module. The released v3.0.0 line subsequently evolved those reconciliation rules substantially in the production adapter, including the final PI live/journal identity fix and OMP live/replay reconciliation.

The exact draft module/diff is therefore no longer safe or current, and merging it now would replace later validated behavior with an older architectural slice.

The remaining structural cleanup is tracked in #330 against current main; future refactoring should start from the released v3.0.0 code, preserve the behavioral contracts, and not revive this draft branch.

Issue #325, opened 2026-08-28
Session handoff UX: machine/harness/model choice, lineage and transferred-history visibility


giuliastro:
> **COMPLETED for the Harness Remote 3.0 release scope.**
>
> The validated same-machine cross-harness handoff UX shipped in v3.0.0. Cross-machine continuation is intentionally tracked separately in #347. The pre-release wording below is retained as historical context.

Follow-up for the current experimental "Continue with another harness" feature.

The functional handoff already works: it creates a real native Session on the target harness and carries bounded context forward. The remaining gap is product/UI clarity.

Future UX requirements:
- make it obvious that the newly created Session is a continuation of the previous Session, not an unrelated Session
- when continuing, allow explicit selection of machine, harness and model
- show the relationship/lineage between source and target native Sessions
- provide a way from the target Session to inspect the source Session history and/or exactly what context was transferred
- provide navigation back to the previous linked Session
- consider how a chain of multiple handoffs is represented
- preserve the product model: each step remains a real harness-owned native Session, with linkage metadata rather than introducing a synthetic Conversation identity
- design the flow for multi-machine handoff separately from the currently validated same-machine implementation

Do not implement this larger UX yet. For now the existing handoff action should be clearly marked Experimental in the UI.

giuliastro:
Implementation completed in PR #334 against `checkpoint/v3-session-first-working-2026-08-25`.

The final UX uses Machine / Harness / Model routing directly in the mature v3 composer. Changing destination is only a preview; the linked native Session is created when Send is pressed. Same-machine and cross-machine continuation, persistent bidirectional lineage, previous/next Session navigation, transferred-context inspection, canonical machine identity and mobile responsive layout are implemented.

Validation is fully green on PR checks run #1475, including type-check/build, full web regressions, bridge tests on Linux/macOS/Windows, Chromium product smoke and signed debug APK.

Head: `a977f7893939749264d6ff6dc0f991b42777277a`.

Note: PR #334 is still visually marked Draft only because the GitHub connector's Ready-for-review mutation currently errors on an invalid GraphQL field (`Repository.fullDatabaseId`). The implementation and CI are complete; no code or review blocker remains.

giuliastro:
PR #334 has been reworked to the same-machine scope requested for final audit.

Key changes:
- cross-machine routing remains deferred; no machine selector and no /v1/session-handoffs endpoint
- machine-level /v1/session-links is used for durable lineage reads
[...]

PR #334, opened 2026-08-29
feat(v3): add same-machine native Session handoff UX and lineage

giuliastro:
## Summary

Implements the Session handoff UX from #325 on the frozen v3 Session-first checkpoint, deliberately scoped to **same-machine continuation**.

- removes the separate header "Continue with another harness" action
- adds Harness + Model destination controls to the mature v3 composer
- changing Harness/Model is inert: no mutation occurs until Send
- if Send targets another harness, creates one linked real native Session on that harness
- keeps every hop harness-owned; no synthetic Conversation/Task identity is introduced
- source Session shows `Continues in`; target Session shows `Continued from`
- linked Sessions are navigable in both directions and naturally support handoff chains
- persists the exact bounded transferred-context packet in `SessionLinkStore`, so it remains inspectable after reopen/reload
- keeps the existing mature v3 transcript, controller, scrolling, reasoning/activity and error-rendering path
- uses canonical daemon machine IDs for native Session identity and migrates legacy browser state
- keeps ConversationStatePill visible on mobile and integrates the routing controls responsively

## Deliberate scope

Cross-machine handoff is **not** implemented in this PR.

The previous cross-machine prototype and `/v1/session-handoffs` endpoint have been removed. Multi-machine continuation needs a separate design for target project/path mapping, daemon capability/rolling-upgrade negotiation and recovery/reconciliation. This PR keeps the validated daemon handoff boundary strictly same-machine.

## Recovery / idempotency

Resource creation and prompt delivery are deliberately separate recovery domains:

- handoff Session creation persists its `clientRequestId` **before** network I/O
- if that recovery state cannot be persisted, resource creation is not started
- ambiguous Session creation has no blind client TTL; retries retain the original idempotency key regardless of age
- definite handoff 4xx releases the creation key and allows a genuinely new attempt
- an accepted creation does **not** immediately clear its creation key
- the returned target ref is first persisted as the durable route anchor; only then is the handoff creation acknowledged and its creation key released
- if the client crashes after route-target persistence but before acknowledgement, the durable route target is authoritative on restart and the stale creation key is then acknowledged
- once a target Session is known, the route record owns only that target identity (not prompt/model payload) and is never discarded by a TTL
- prompt idempotency/retry semantics remain exclusively in `native-session-prompt`
- an ambiguous first target prompt must follow normal prompt reconciliation, but can never cause another target Session to be created
- a definite first-prompt 4xx releases prompt delivery state while retaining the already-created route target; a changed prompt/model therefore retries the **same target Session**
- attachments are not transferred across harnesses in this iteration; existing attachments remain visible/removable and block routed Send until removed

## Canonical machine identity migration

The checkpoint previously keyed several browser records by the saved `WorkspaceMachine.id`. Native Session identity now uses `snapshot.machine.id`.

Before native Sessions are used, the app migrates:

- pending prompt
- pending command
- pending handoff
- handoff-context marker
- route continuation state
- native Session draft

Migration preserves stored payloads/clientRequestIds and does not overwrite a conflicting canonical record.

## Lineage persistence

`/v1/session-links` is a **machine-level** endpoint. Lineage reads use the unscoped workspace machine config, not the agent-scoped native Session config.

`SessionLinkStore` persists:

- source native Session ref
- target native Session ref
- creation timestamp
- bounded transferred context

This lets the target recover both lineage and the exact transferred context after navigation or a full reload.

## Safety / architecture

- base: `checkpoint/v3-session-first-working-2026-08-25`
- `main` is untouched
- daemon handoff remains same-machine
- destination changes do nothing until Send
- slash commands remain scoped to the open Session
- model-only changes on the current harness remain normal same-Session model changes

## Added validation

Coverage now includes:

- resource-creation idempotency retained even for a pending handoff older than 24h
- lost-response handoff retry reuses the same `clientRequestId`
- refusal to create a Session when durable client recovery state cannot be written
- accepted creation keeps its idempotency key until the returned target has been durably promoted to route state
- crash-safe recovery across the target-persisted / creation-acknowledgement boundary
- definite handoff 4xx permits a fresh creation attempt
- canonical machine-ID storage migration, including payload/clientRequestId preservation
- local `SessionLinkStore` persistence and transferred-context persistence
- machine-level lineage endpoint
- browser attachment blocking during harness change
- browser model-catalog isolation across harness changes
- browser handoff creation 409 followed by a successful new attempt
- browser **target created -> first target prompt 4xx -> retry with changed prompt/model on the same target**
- browser reload followed by durable lineage + transferred-context recovery

## Current validation status

Final validation head: `85345eb867d4e000093ac8c71f6f6d7c44595f7d`.

PR checks **#1535 are fully green**:

- TypeScript type-check + production web build ✅
- full web regression suite ✅
- native Session canonical machine-ID migration test ✅
- native Session handoff recovery/idempotency tests ✅
- bridge tests on Linux ✅
- bridge tests on macOS ✅
- bridge tests on Windows ✅
- Chromium portrait/landscape + general multi-machine smoke ✅
- PI native Session smoke ✅
- OpenCode native Session + real-regression smoke ✅
- same-machine model-switch + handoff smoke ✅
- target-created → first-prompt 4xx → same-target recovery ✅
- reload/reopen durable lineage + transferred-context recovery ✅
- complete controls + screenshot audit ✅
- Android debug APK build + signature verification ✅

Artifacts:

- `harness-remote-debug-v2.11.1-dcecf9d`
- `v3-browser-audit-dcecf9d0d55c8c4fbf307ff85b88432b46011456`

The PR intentionally remains **draft** for a final audit before merge.

## Rolling-upgrade compatibility

Opening an existing native Session must remain safe when the mobile/web client is newer than the daemon.

- `listNativeSessionLinks` now validates the runtime response shape instead of trusting the TypeScript type.
- Legacy or malformed `/v1/session-links` responses, including Android-normalized HTTP 204 `true`, are treated as an empty lineage list.
- The workspace defensively refuses to store a non-array lineage collection.
- Chromium smoke explicitly opens a Session while the fake daemon answers the first lineage GET with HTTP 204 and verifies that the composer remains usable.

This fixes the runtime crash `Cannot read properties of undefined (reading 'find')` seen when opening a Session against a daemon without the new lineage endpoint.


giuliastro:
Validation complete: PR checks #1475 is fully green, including Chromium product smoke and signed debug APK. I attempted to mark this PR Ready for review, but the GitHub connector currently fails that mutation because it queries the nonexistent GraphQL field `Repository.fullDatabaseId`. Draft status is therefore a connector limitation only; this PR is ready for manual review/merge into the checkpoint.

PR #357, opened 2026-09-02
fix: keep HR3 machine endpoint for single-harness startup

### LINKED ISSUE #356: HR3 launcher should always expose a Machine endpoint for single-harness setups
## Problem

Harness Remote 3 is machine-first: the current client adds connections through **Machines** and requires `/v1/machine` discovery.

The launcher still preserves the legacy single-backend path whenever exactly one supported CLI is detected. That creates two compatibility gaps:

- **OpenCode-only**: the launcher starts managed `opencode serve` directly. That server is not a Harness Machine endpoint, so HR3 rejects it in Machines.
- **Single ACP backend**: the standalone bridge does expose `/v1/machine`, so existing Sessions remain visible, but it does not expose the full daemon `/v1/projects` catalog. The HR3 Machine → Project → Session creation flow is therefore incomplete.

This also means the current quick-start claim that one detected CLI automatically gives the normal HR3 setup is too broad.

## Desired behavior

Without `--single`, `harness-remote` should expose the full HR3 Machine contract even when only one harness is installed, including OpenCode-only machines.

`--single` should remain the explicit legacy compatibility opt-out for users who intentionally want the old per-harness endpoint.

## Migration/documentation

README and `docs/QUICK_START.md` now include an HR2 → HR3 startup migration note explaining that legacy standalone endpoints are compatibility paths, direct 2.x-style `opencode serve` is not an HR3 Machine, and saved 2.x server profiles are not automatically converted to `workspaceMachines`.

giuliastro:
## Summary

Fixes #356.

Harness Remote is machine-first in v3, so normal `harness-remote` startup now keeps the full Machine contract even when only one harness is available.

### Behavior

- single ACP harness -> machine daemon, including `/v1/machine` and `/v1/projects`
- OpenCode-only -> machine daemon with OpenCode as the managed HTTP primary
- `--backend opencode` selects OpenCode as the Machine primary without dropping back to a direct server
- `--single` remains the explicit legacy per-harness compatibility opt-out

### Implementation

- allow the daemon configuration to represent an OpenCode primary without inventing an ACP backend
- make the primary ACP bridge optional in the machine daemon
- route an HTTP primary through the existing managed-host proxy, including unprefixed compatibility routes
- keep ACP startup and Session semantics unchanged
- update launcher help, README and quick-start migration docs

### Regression coverage

Adds focused tests for:

- one detected ACP using daemon mode
- OpenCode-only daemon mode
- explicit OpenCode primary
- `--single` legacy opt-out
- OpenCode daemon configuration without phantom ACP state
- unprefixed routing when the Machine primary is managed HTTP
- OpenCode-only machine wiring never constructs a phantom ACP bridge

No web/UI, transcript, Session reconciliation, deletion or handoff behavior is changed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -186,7 +186,7 @@ See [REFERENCE.md](REFERENCE.md) for detailed security and backend notes.
 
 ## Quick start
 
-> **Upgrading from Harness Remote 2.x?** HR3 uses a **Machine daemon** as its normal connection contract, not the old per-harness server profile model. Old standalone ACP bridge commands still start and can expose native Sessions, but they are a legacy compatibility path and do not provide the complete HR3 Machine → Project → Session workflow. A direct 2.x-style `opencode serve` endpoint is **not** an HR3 Machine endpoint. Saved 2.x server profiles are also not automatically imported into the new Machines list, so add the machine again after upgrading. For HR3, prefer the launcher/daemon setup below and connect through **Machines → Add machine**.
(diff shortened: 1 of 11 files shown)


giuliastro:
Superseded by #374. The final #357 file state was ported onto current `main` after verifying none of its 11 touched files had changed since the old merge base. #374 then passed the current full PR suite, including Chromium product smoke and Debug APK, and was merged. Closing this stale draft to avoid two sources of truth.

PR #363, opened 2026-09-04
fix(web): drive machine refresh from live event stream

### LINKED ISSUE #362: Machine snapshots poll on a 10s timer while a live event stream is already connected
### Summary

`NativeSessionsWorkspace` refreshes machine snapshots on a fixed timer even when every machine already has a healthy live event stream open. The stream is subscribed for transcripts, but nothing in the workspace listens to it for machine/session state, so an agent that finishes a turn can take up to 10 seconds to show up in the rail. From the outside this looks like the app polls instead of streaming.

### Where

- `web/src/components/standalone-universal-workspace.tsx:56-57` — `MACHINE_NORMAL_POLL_MS = 10_000`, `MACHINE_RECONNECT_POLL_MS = 1_500`
- `web/src/components/standalone-universal-workspace.tsx:445-459` — the only thing that bumps `revision` (and therefore re-runs discovery at `:381-439`) is `window.setInterval` plus `visibilitychange`
- `web/src/taskdesk-live-events.ts:59` — `subscribeTaskDeskLiveEvents` already exists and is used elsewhere; the workspace never calls it
- `web/src/opencode-events.ts:89-193` — the fetch transport already does reconnect with exponential backoff (1s → 30s), a 30s stall watchdog, and publishes `connected` / `reconnecting` / `connection-error` / `closed` status

The transport comment at `opencode-events.ts:87` says it "deliberately keeps the UI on its existing polling fallback if the stream is unavailable" — which is the right design, but the other half (letting the stream drive the refresh while it *is* available) is missing at the call site.

### Suggested change

Let the stream drive the refresh and demote the timer to the fallback it was designed to be:

1. Subscribe to each machine's stream in the workspace; on any event, request a refresh through a leading-edge burst limiter (one turn emits many part/message events, so a refresh per event would hammer `/v1/machine`).
2. On `connected`, refresh once — a stream that just recovered was silent for the whole outage, so the current snapshot is stale.
3. Choose the timer interval from stream health: flapping machine → `1_500`; every machine streaming → a long safety net (60s); any machine without a stream → the current `10_000`. Partial coverage must keep the normal cadence, since the machines without a stream are exactly the ones whose changes would go unnoticed.

Keying the subscription effect on endpoint identity rather than the `machines` array identity matters: the parent rebuilds that array on every persist, and re-subscribing per render tears down healthy streams.

### Measured effect

I implemented the above against `v3.0.1` and drove the real built app in Chromium against a machine daemon I could control (real `/v1/machine` payload, scriptable SSE):

| Scenario | Before (by inspection) | After (measured) |
| --- | --- | --- |
| Idle 25s, stream connected | ~2 `/v1/machine` reads | **0 reads** |
| Agent event arrives | up to 10s late | **48 ms** to the refresh |
| Stream forcibly killed | n/a | timer takes over, 9995 ms gaps |
| Stream restored | n/a | transport reconnects on its own, refresh fires immediately on `connected` |
| Burst of 25 events | 1 read per poll tick | 1 leading + 1 trailing read |

`tsc -b` is clean, `src/v3-workspace-navigation.test.mjs` and `src/workspace-runtime-merge.test.mjs` still pass, and the policy (interval selection + burst collapsing) has unit tests.

Happy to open a PR with the patch if the approach looks right to you — in particular whether the 60s safety net and the 400ms burst window are the numbers you'd want, and whether a refresh should be scoped to the machine that emitted the event rather than refreshing all of them (the current discovery cycle is all-machines, so I kept that).

andrebrait:
Resolves #362

### Summary
`NativeSessionsWorkspace` refreshed machine snapshots on a fixed timer (`10_000ms`) even when connected machines already had an active, healthy live event stream open. This caused agent turn completions to take up to 10 seconds to reflect in the session rail.

This PR wires `subscribeTaskDeskLiveEvents` into `NativeSessionsWorkspace` so events immediately trigger a refresh, demoting the polling timer to a safety net/fallback.

### Changes
1. **Live event subscription**:
   - Subscribes to each machine's live event stream via `subscribeTaskDeskLiveEvents`.
   - Keys the subscription on endpoint identity (`host:port:user`) rather than the `machines` array reference, preventing unnecessary reconnects during parent component re-renders.
   - Triggers an immediate refresh on `connected` to reconcile any turns that completed while disconnected.

2. **Burst limiting (`web/src/machine-live-refresh.ts`)**:
   - Added a 400ms leading-edge burst limiter so the flood of events emitted during an active agent turn coalesces into at most one leading and one trailing `/v1/machine` read.

3. **Dynamic interval selection**:
   - When all machines have a healthy event stream: timer is demoted to a 60s safety net.
   - When any machine is missing an active stream: falls back to the standard 10s poll cadence.
   - When any machine is flapping/reconnecting: uses the 1.5s fast reconnect poll.

4. **Tests**:
   - Added unit tests in `web/src/machine-live-refresh.test.mjs` covering burst collapsing, cancel semantics, and interval selection under all machine states.
   - Added `machine-live-refresh.test.mjs` to `test:v3-product-ui` in `web/package.json`.

### Verification
- `npm run test:v3-product-ui` passes (91/91 tests).
- `npx tsc -b` passes cleanly.
- Production build (`npm run build`) builds with 0 errors.
- End-to-end verified on live daemon + headless browser rig:
  - Turn completion latency to UI refresh dropped from up to 10s to **48ms**.
  - Idle machine reads dropped to **0** while connected.
  - Stream drop immediately falls back to 10s timer.


web/src/components/standalone-universal-workspace.tsx
```diff
@@ -442,9 +450,48 @@ function NativeSessionsWorkspace({
...
+      limiter.cancel()
+      for (const subscription of subscriptions) subscription.close()
+    }
+  }, [streamTargets])
+
   useEffect(() => {
     if (!loaded) return
-    const interval = reconnectingCount > 0 ? MACHINE_RECONNECT_POLL_MS : MACHINE_NORMAL_POLL_MS
+    const interval = machinePollIntervalMs({
+      reconnecting: reconnectingCount > 0,
```
giuliastro: `reconnectingCount` tracks discovery failures, not SSE stream recovery. A healthy `/v1/machine` endpoint with a reconnecting event stream therefore gets the normal 10 second cadence instead of `MACHINE_RECONNECT_POLL_MS`. Please keep a reconnecting stream state and pass it here, with a test covering the transition.

giuliastro (changes requested):
Thanks for taking this on, Andre. The stream driven refresh and burst limiting are a nice improvement, and the focused tests make the intended policy easy to follow.

I found one integration gap before merging. The reconnecting branch in `machinePollIntervalMs` is never driven by the event stream status. The workspace passes `reconnectingCount`, which only counts `/v1/machine` discovery errors. When a stream reports `reconnecting`, `connection-error`, or `closed`, the code only marks that machine as not live, so the timer falls back to 10 seconds instead of the intended 1.5 seconds during stream recovery. Please wire the stream status into this input and add coverage for that path.

Once that is fixed, I will be happy to merge it. Thanks again for the thoughtful implementation.

andrebrait:
@copilot review

web/src/components/standalone-universal-workspace.tsx
andrebrait: Fixed. Added `reconnectingStreams` state in `NativeSessionsWorkspace` tracking `isStreamReconnecting(status.type)` (`reconnecting`, `connection-error`, `closed`) and wired it to `machinePollIntervalMs`, with unit tests covering recovery transitions.

giuliastro (approved):
Thanks, Andre. I re-checked the updated implementation and the reconnecting-stream gap from the previous review is now addressed: stream recovery states drive the 1.5s reconnect polling path, with focused coverage for the transition back to a healthy stream. The change remains scoped to the web session-refresh path and does not touch ACP/backend behavior. Looks good to merge — thanks for the fix.

Issue #367, opened 2026-09-06
Make PR checks required, then de-duplicate post-merge CI


giuliastro:
## Why

Audit after the red-main incident found that PR #363 was merged while its pull-request workflows were red. That means `main` is not currently protected by the checks we rely on for validation.

Because of that, the regression suites repeated inside Pages, Android APK/AAB and Desktop packaging are currently acting as a post-merge safety net. Removing them first would make the repository faster but less safe.

## Current repository state — 2026-09-06

Repository metadata currently reports:

- `main`: **not protected**;
- required-status enforcement: **off**;
- repository rulesets: **none**.

The connected GitHub App has contents/workflow/PR write permissions but not repository administration, so the final ruleset/branch-protection switch cannot be performed through the current connector.

There is also a prerequisite before enabling required checks: `.github/workflows/pr-checks.yml` currently has a workflow-level `pull_request.paths` filter limited to `web/**`, `bridge/**` and the workflow file itself. If its job names were made globally required as-is, docs-only or other out-of-scope PRs could fail to produce the required contexts and become unmergeable.

## Phase 0 — make the required gate safe for every PR

Before changing repository settings:

- ensure the required gate is instantiated for **every pull request**;
- either run the canonical checks on every PR initially (simplest/safest), or introduce an always-present lightweight gate that evaluates path-specific jobs without making required contexts disappear;
- do not rely on a workflow-level `paths` filter for a context that will be globally required;
- keep path-scoped advisory workflows such as `Verify desktop application menu` out of the globally required set unless they are changed to always emit a terminal status.

## Phase 1 — enforce the pre-merge gate

Protect `main` with a ruleset / branch protection that requires pull requests and required status checks before merge.

Recommended required checks once Phase 0 is complete:

- `Type-check and regressions`
- `Bridge tests (macos-latest)`
- `Bridge tests (windows-latest)`
- `Chromium product smoke`
- `Debug APK` if installability should block merges; otherwise keep it advisory because it is the slowest job.

For maximum release safety during the current stabilization phase, prefer making `Debug APK` required as well and optimize the gate later rather than weakening it first.

## Phase 2 — simplify CI after enforcement

Only once bypassing red PR checks is no longer possible:

- make `PR checks` the canonical full validation suite;
- stop manually repeating generic web regressions inside Pages/APK/AAB/Desktop packaging;
[...]

PR #375, opened 2026-09-06
ci: centralize regression test tiers

giuliastro:
Part of #367.

This centralizes the duplicated web regression command lists into named CI tiers in `web/package.json` without reducing coverage:

- `test:ci:baseline` preserves the current Android APK/AAB suite;
- `test:ci:pages` preserves the current Pages suite (baseline + attachments);
- `test:ci:desktop` preserves the current Desktop suite including platform/Electron coverage;
- `test:ci:full` preserves the complete PR regression suite.

The workflows will call these tier scripts instead of maintaining independent copied command lists. This makes future test additions/removals explicit and prevents CI drift while keeping the existing post-merge safety nets in place until required-check enforcement can be configured for `main`.

No runtime product behavior changes.

### DIFF
--- a/.github/workflows/android-aab.yml
+++ b/.github/workflows/android-aab.yml
@@ -35,14 +35,7 @@ jobs:
         working-directory: web
 
       - name: Run web regression tests
-        run: |
-          npm run test:i18n
-          npm run test:config
-          npm run test:ui
-          npm run test:settings
-          npm run test:model
-          npm run test:events
-          npm run test:profiles
+        run: npm run test:ci:baseline
         working-directory: web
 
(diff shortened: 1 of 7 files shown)


giuliastro:
Direction looks good, but the latest diff has unrelated churn that should be removed before merge. In particular, it deletes a number of useful explanatory comments from Android/Desktop workflow files and introduces `recordss the reasoning` in CONTRIBUTING.md. Please keep the CI-tier centralization and the intended command substitutions, but restore unrelated comments/copy from current main and fix the typo. I would not merge this version as-is even if CI is green.

PR #381, opened 2026-09-06
fix(web): use native Session model options

dnhrdt:
## Problem

An existing native Session can accept a different model option set from the machine-wide catalog used for a fresh technical Session.

I reproduced this with a real Claude ACP Session: the global catalog exposed a decorated Fable model id, while the exact Session's `/config/providers?sessionID=...` response exposed the bare id it accepted. The native Session surface populated its picker from the global catalog, so choosing the visible Fable entry failed client-side as unavailable before the adapter received a model mutation. Choosing `Default` instead reached the adapter but selected its actual default (Opus), so it was not an equivalent fallback.

The native observer currently passes an empty catalog scope, and the shared conversation controller also switches to the empty scope whenever routing is available, even when the selected harness has not changed.

## Change

- carry native `sessionID` and `directory` through `AgentModelScope` and its cache key;
- load `/config/providers` for that exact existing Session;
- route the request with the explicitly selected agent rather than a potentially stale profile agent id;
- retain the machine-wide catalog only when a cross-harness action will create a fresh Session;
- add a behavioral regression that proves both request paths, plus the existing source-level UI guards.

This keeps model authority with the native Session, consistent with the Session-first compatibility rules and the catalog work tracked in #287.

## Verification

- Reproduced the differing global and Session-specific catalogs against `claude-agent-acp` 0.75.1.
- `cd web && npm run build`
- `cd web && npm run build:electron`
- `cd web && npm run test:ci:full`
- `cd bridge && npm test`: 472 of 473 passed; the unchanged Windows symlink-boundary test could not create its fixture (`EPERM`) before reaching its assertion.

I did not install a locally signed Android build. The shared client production build and full mobile regression tier pass; the real-harness verification covered the model-catalog endpoints and the original rejection path.


giuliastro (approved):
Reviewed the full diff. The change stays in the web client/model-selection path and regression tests; it does not modify ACP services, harness adapters, daemon/runtime behavior, or workflows. Session-scoped model discovery is routed through the explicitly selected agent and the bridge already supports `/config/providers?sessionID=...` for ACP-backed native Sessions. Cross-harness handoff still uses the fresh-session machine catalog. I don't see a regression blocker here. Thanks for the focused fix.

giuliastro:
Important follow-up after merge: #381 was reverted from `main` via #383 because manual validation exposed a serious cross-harness model-routing regression. In an existing Codex Session, the model request could combine a Codex agent path with a different machine/backend routing identity, causing another harness's model catalog to appear. That is a high-severity regression for this project because Session/harness isolation must be reliable.

Please do not treat the merged #381 as still accepted. Its feature is currently out of `main`.

If you want to reintroduce this behavior, please open a new PR based on the current `main` and validate it much more aggressively before asking for merge. At minimum, the replacement PR must prove with end-to-end tests that:

- the currently opened Session always uses a coherent `agentId` + `backend` identity;
- a machine whose primary harness is OMP/PI/Claude/OpenCode can open an existing Codex Session without leaking that primary harness's model catalog;
- cross-harness continuation still uses the correct fresh-session catalog;
- existing Sessions for Codex, Claude Code, OMP, PI and OpenCode continue to open and operate normally;
- model selection, transcript loading, new Session creation and Session switching are all regression-tested.

Please also perform real-harness validation yourself before resubmitting. We should not rely on maintainer manual testing to discover a regression of this severity after merge.

PR #382, opened 2026-09-07
fix: preserve bounded native session pagination

dnhrdt:
## Summary

- preserve the opaque ACP `session/list` cursor through the bridge and web API
- keep the recurring Session rail refresh bounded to the first page
- add explicit incremental loading for older pages, with deduplication and cursor reset after adapter changes
- retain owned Session metadata and status enrichment without loading transcript snapshots
- keep an in-flight Session deletion visible until DELETE succeeds, then suppress that exact identity from cached pages

## Why

Codex ACP returns native Sessions in pages of 25. Harness Remote currently drops `nextCursor`, so older Sessions are invisible. Eagerly flattening the complete catalog fixes the data loss but makes the 30-second rail refresh expensive: the observed 794-Session response was about 317 KB and took about 10.5 seconds with a warm adapter.

This change keeps ordinary discovery cheap. Page one refreshes automatically; each `Load older Sessions` action advances one page for the applicable harnesses and keeps the accumulated rows in the client.

The Chromium gate also exposed browser fixtures that still served the machine-wide model endpoint after #381 switched existing Sessions to their Session-scoped catalog. Those fixtures now exercise the merged endpoint contract. The same gate caught premature cached-row suppression during an in-flight DELETE; pending and successful deletion states are now distinct.

## Verification

- `node --test test/acp-client.test.js test/server.test.js` — 70/70
- `npm run test:ci:full` — pass
- `npm run build` — pass
- `npm run build:desktop` — pass
- `npm run test:desktop` — 20/20
- all five native Session Chromium product scenarios — pass locally and in GitHub CI
- live Codex adapter smoke — 25 rows on page one, 25 non-overlapping rows on page two, next cursor retained (about 5.0s cold / 0.55s warm)
- GitHub CI — type/regressions, Bridge on Windows and macOS, Chromium product smoke, macOS application menu, and Debug APK all pass

The full bridge suite passes 474/475 on this Windows host. The unchanged `project-catalog` symlink fixture receives `EPERM` before reaching its assertion because this account cannot create symlinks.

No Android package or running Harness daemon was replaced for this test.

### DIFF
--- a/bridge/src/acp-client.js
+++ b/bridge/src/acp-client.js
@@ -236,10 +236,17 @@ export class AcpClient extends EventEmitter {
     this.#child.stdin.write(`${JSON.stringify({ jsonrpc: "2.0", method, params })}\n`)
   }
 
-  async listSessions() {
+  async listSessionPage(cursor) {
     await this.start()
-    const result = await this.request("session/list", {})
-    return result.sessions ?? []
+    const result = await this.request("session/list", cursor ? { cursor } : {})
+    return {
+      sessions: result.sessions ?? [],
+      ...(typeof result.nextCursor === "string" && result.nextCursor ? { nextCursor: result.nextCursor } : {})
+    }
+  }
+
+  async listSessions() {
+    return (await this.listSessionPage()).sessions
   }
 
   close() {
--- a/bridge/src/acp-prompt-echo-filter.js
+++ b/bridge/src/acp-prompt-echo-filter.js
@@ -42,6 +42,7 @@ export class AcpPromptEchoFilter extends EventEmitter {
   get processID() { return this.#acp.processID }
 
   start(...args) { return this.#acp.start(...args) }
+  listSessionPage(...args) { return this.#acp.listSessionPage(...args) }
   listSessions(...args) { return this.#acp.listSessions(...args) }
   notify(...args) { return this.#acp.notify(...args) }
   close(...args) { return this.#acp.close?.(...args) }
--- a/bridge/src/acp-service.js
+++ b/bridge/src/acp-service.js
@@ -565,6 +565,25 @@ export class AcpService {
     })
   }
 
+  /**
+   * Seed lightweight ACP listing metadata without restoring a transcript snapshot.
+   *
+   * The paged Session index can reveal an older Session before any stable endpoint or mutation
+   * addresses it. Remembering that metadata lets a later open/claim resolve the native identity
+   * without turning every recurring index refresh into an eager walk of the full history.
+   */
+  rememberListedSessions(sessions) {
+    return sessions.map((session) => {
+      const known = this.#sessions.get(session.sessionId)
+      const updatedAt = this.#preserveListedTimestamps && known?.updatedAt
+        ? known.updatedAt
+        : session.updatedAt ?? known?.updatedAt ?? new Date().toISOString()
+      const normalized = { ...session, updatedAt }
+      this.#sessions.set(normalized.sessionId, normalized)
+      return normalized
+    })
+  }
+
   async createSession({ directory, title, model }) {
     await this.#acp.start()
     const result = await this.#acp.request("session/new", { cwd: directory, mcpServers: [] })
@@ -1746,17 +1765,8 @@ export class AcpService {
   async #refreshSessions() {
     if (!this.#sessionListing) {
       this.#sessionListing = this.#acp.listSessions().then((sessions) => {
-        const listed = new Set()
-        const refreshed = sessions.map((session) => {
-          listed.add(session.sessionId)
-          const known = this.#sessions.get(session.sessionId)
-          const updatedAt = this.#preserveListedTimestamps && known?.updatedAt
-            ? known.updatedAt
-            : session.updatedAt ?? known?.updatedAt ?? new Date().toISOString()
-          const normalized = { ...session, updatedAt }
-          this.#sessions.set(normalized.sessionId, normalized)
-          return normalized
-        })
+        const listed = new Set(sessions.map((session) => session.sessionId))
+        const refreshed = this.rememberListedSessions(sessions)
         for (const [sessionID, session] of this.#sessions) {
           if (this.#ownedSessions.has(sessionID) && !listed.has(sessionID)) refreshed.push(session)
         }
--- a/bridge/src/codex-session-history.js
+++ b/bridge/src/codex-session-history.js
@@ -198,7 +198,10 @@ export function createCodexHistoryLoader(sessionRoot = path.join(homedir(), ".co
 
   loadCodexHistory.page = async (sessionID, options = {}) => {
     const file = await locateSession(sessionID)
-    if (!file) return { messages: [], before: null, hasMore: false }
+    // Absence is not proof of an empty Session: a rollout may live outside this installation's
+    // local tree while ACP session/load can still replay it. Let AcpService distinguish that from a
+    // located, genuinely empty rollout and fall through to the protocol source.
+    if (!file) return undefined
     return readCodexPage(file, sessionID, options)
   }
 
--- a/bridge/src/server.js
+++ b/bridge/src/server.js
@@ -227,13 +227,18 @@ export function createBridgeServer({ config, acp, serviceOptions, machineRegistr
   // Live activity is tracked separately so an active Session still sorts to the top without reading
   // its transcript. Invalid or missing harness timestamps stay at zero instead of pretending to be
   // freshly updated on every poll.
-  const listVisibleSessionMetadata = async (directory) => {
-    const [sessions, deletedSessionIDs] = await Promise.all([
-      acp.listSessions(),
+  const listVisibleSessionMetadata = async (directory, cursor) => {
+    const [page, deletedSessionIDs] = await Promise.all([
+      typeof acp.listSessionPage === "function"
+        ? acp.listSessionPage(cursor)
+        : cursor
+          ? { sessions: [] }
+          : acp.listSessions().then((sessions) => ({ sessions })),
       service.deletedSessionIDs()
     ])
+    service.rememberListedSessions(page.sessions)
     const visible = new Map(
-      sessions
+      page.sessions
         .filter((session) => !directory || sameListedDirectory(session.cwd, directory))
         .filter((session) => !hiddenSessionIDs?.has(session.sessionId))
         .filter((session) => !deletedSessionIDs.has(session.sessionId))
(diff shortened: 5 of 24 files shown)


giuliastro (dismissed):
Thanks — the pagination approach is well motivated and the current CI is green, but I’m holding this PR from merge because it does not satisfy the required Codex-only safety scope.

The change modifies shared ACP/native-Session paths used by multiple harnesses, not only Codex: `AcpClient`/`AcpPromptEchoFilter`, `AcpService`, `createBridgeServer`, and the generic Native Session discovery/home flow. In this repository the same `AcpClient` infrastructure is used by OMP, PI, Claude Code, and Codex, so these changes increase the regression surface outside Codex even if the current tests pass. The PR also updates OpenCode smoke fixtures.

Please scope the pagination behavior explicitly to Codex, or introduce a capability/backend gate so OMP, PI, Claude Code, OpenCode, and the existing generic ACP behavior remain unchanged. Keep the bounded first-page refresh / explicit older-page loading for Codex and the existing regression coverage. Once the shared-harness surface is isolated, this should be straightforward to re-review.

giuliastro (commented):
Correction to my previous review: the shared implementation is acceptable and potentially desirable. ACP `session/list` pagination is protocol-level behavior, so this should not be forced into a Codex-only branch if other adapters can benefit from the same path.

The merge criterion is now regression safety across every affected harness. Please keep/extend coverage so OMP, PI, Claude Code, OpenCode, and Codex all preserve their existing Session behavior when `nextCursor` is absent, while adapters that paginate can expose bounded first-page refresh plus explicit older-page loading. Codex already has the strongest live evidence; the remaining concern is proving the shared path is neutral for the others rather than isolating it from them.

dnhrdt:
Thanks for the correction — agreed that pagination belongs at the ACP protocol level.

I added commit 8f2b767 with the requested regression-safety coverage:

- The bridge now exercises the no-nextCursor Session index and status behavior for OMP, PI, Claude Code, and Codex. Each backend keeps its existing single-page result, emits no cursor header, and never follows an invented cursor.
- The shared web discovery path now exercises the same no-nextCursor behavior for OpenCode, OMP, PI, Claude Code, and Codex, preserving backend/transport/status and exposing no Load older cursor.
- The existing tests still cover opaque cursor forwarding through AcpClient/bridge and explicit older-page loading in the Session rail.

Local verification: the focused bridge pagination/index tests pass (3/3) and web test:ci:full passes. The full bridge suite is 475/476 locally; its only failure remains the pre-existing Windows EPERM symlink-fixture setup failure. The pushed CI run will provide the clean cross-platform result.

giuliastro (changes requested):
Manual Codex validation found two merge-blocking regressions on this branch.

1. **Existing Codex Sessions open with an empty transcript.** The Session detail falls through to the empty-state UI (`Start the conversation…`) instead of showing the persisted Codex history. The current Session UI also swallows initial transcript-read failures, so a failed/unreadable history is visually indistinguishable from a genuinely empty Session. Please reproduce this against a real Codex installation and verify both a page-one Session and an older paginated Session open with their actual transcript. If a transcript read fails, surface the failure rather than rendering the empty-conversation state.

2. **The Codex model picker can show models from another harness.** I traced a concrete routing bug introduced by the Session-scoped model-catalog path from #381 and exposed on this branch. `WorkThreadConversation` uses the selected machine's `config` as `destinationConfig` even when the route has not changed. The Session-scoped catalog path then calls `api.listModels({ ...config, agentId }, ...)`: `agentId` becomes `codex`, but `backend` can remain the machine/primary harness. The request therefore has a `/v1/agents/codex/...` path with a mismatched `X-Harness-Backend`; `routedAgentForBackend()` intentionally corrects that mismatch by rerouting to the backend named in the header. Result: a Codex Session receives another harness's catalog.

For the current native Session, the model-catalog request must use the exact Session/agent config (matching both `agentId` and `backend`). The machine-wide config is appropriate only when a real cross-harness route is creating a new Session. Please add a regression where the machine primary/backend differs from the opened Codex agent and assert that the existing Codex Session still reads only Codex's Session-scoped catalog.

The pagination work itself is still reasonable and the new cross-harness no-cursor tests are useful, but listing pagination is not sufficient validation: this cannot merge until real Codex **open transcript + model catalog** behavior is green as well. The misplaced empty-state icon is secondary to the failed history state, but the failed-read path should not present that state as a valid empty conversation.

giuliastro:
Update: #381 has now been reverted from `main` via #383 after manual Codex validation exposed the Session-scoped model-routing regression. Please rebase this PR onto the new `main` before further validation so #382 does not accidentally reintroduce #381 through its branch ancestry.

For #382 itself, keep the bounded pagination work, but any Session-scoped model-catalog behavior from #381 must either be dropped or reintroduced only with the routing fix: the current Session request must carry a coherent agent/backend identity (e.g. Codex path + Codex backend), with a regression where the machine primary harness differs from the opened Codex harness. Also retain the requested real Codex checks for opening page-one and older paginated Sessions with non-empty persisted transcript, and surface transcript read failure instead of the valid-empty `Start the conversation…` state.

dnhrdt:
Rebased onto the post-#383 `main` (`f89be04`) and force-pushed the clean branch at `b281c18`.

The #381 Session-scoped catalog behavior is **not** reintroduced: `taskClient`, `NativeSessionObserver`, and the model-scope behavior now remain exactly on the reverted upstream baseline. I also dropped the two fixture-only commits that existed solely to support #381. The browser regression now deliberately saves OMP as the machine primary, opens Codex Sessions, and verifies that the picker receives only the Codex machine catalog and never an OMP routing override.

The Codex transcript blocker is fixed in two layers:

- The native rollout reader now understands current `event_msg:item_completed` records (`UserMessage`, `AgentMessage`, and `Reasoning`) while retaining the legacy lowercase event format and excluding response-item/tool payloads.
- A missing local rollout is no longer reported as authoritative empty history; paged reads fall through to ACP replay. A real read failure is surfaced as an alert in the transcript and cannot render the valid-empty “Start the conversation…” state.

Real Codex validation after the fix, using `@agentclientprotocol/codex-acp@1.1.14` plus the actual bridge HTTP message endpoint:

- page-one Session: 43 visible messages, both user and assistant roles, model metadata present;
- older Session from Session-list page 2: 44 visible messages, both roles, model metadata present.

No transcript content was printed during that validation.

Local verification:

- focused Codex history + bridge server tests: 67/67 passed;
- full web `test:ci:full`: passed;
- production web build: passed;
- Chromium Codex navigation/model-routing/failed-history smoke: passed;
- full bridge suite: 477/478 passed; the sole local failure remains the pre-existing Windows `EPERM` while creating the symlink fixture, before that test body can run.

The new GitHub checks are running now.

giuliastro:
Thanks for the follow-up fixes. The branch is materially better, but given the severity of the regressions found during the previous manual validation, this PR is not ready to be handed back to the maintainer for exploratory testing.

Before asking for merge, please complete the end-to-end validation yourself and report the results clearly. In addition to CI, please verify against real harnesses where applicable:

- Codex page-one Session opens with its full persisted transcript;
- an older Codex Session loaded through pagination opens with its full persisted transcript;
- Codex model picker never receives models from a different machine-primary harness;
- Session switching A -> B -> A remains correct and does not retain the previous transcript/catalog;
- create, prompt, model change and delete still work for Codex;
- OMP, PI, Claude Code and OpenCode keep their existing Session behavior, especially where `nextCursor` is absent;
- failures to read persisted history are surfaced as errors and never masquerade as a valid empty Session.

Please include the exact real-harness versions used and the observed results. The maintainer should not have to act as the primary QA layer for changes touching shared Session/ACP behavior after we already found cross-harness routing and transcript regressions on this PR lineage. Once you have completed that validation, we can re-review for merge.

dnhrdt:
Completed the requested owner-side E2E validation on rebased head `adeb2ab` (base `515b0ae`). All current GitHub checks are green.

Environment:

- Harness Remote web `3.0.1`, bridge `0.1.7`, Node `24.18.0`, npm `11.3.0`
- Codex CLI `0.153.4`, `@agentclientprotocol/codex-acp` `1.1.14`, its actually cached embedded `@openai/codex` `0.147.0`
- Claude Code CLI `2.1.263`, `@agentclientprotocol/claude-agent-acp` `0.63.0`, its actually cached embedded `@anthropic-ai/claude-agent-sdk` `0.3.220`
- OMP, PI, and OpenCode were not installed/detectable on this host. I am not claiming live-harness coverage for those three; their no-`nextCursor` behavior is covered by the automated bridge/web matrix.

Observed results:

- **Codex page one:** the native endpoint returned 2 persisted records (1 user, 1 assistant) and the UI rendered exactly those 2 messages, including the expected assistant marker, with no transcript error or empty state.
- **Older paginated Codex Session:** three explicit `Load older Sessions` actions reached Session-list page 4. The native endpoint returned 10 persisted text records (2 user, 8 assistant, `X-Has-More: 0`); all 10 texts were present in the UI, grouped into 4 logical turn bubbles. No page-one marker or catalog leaked into it.
- **Different machine primary:** I deliberately saved `claude` as the machine primary, then opened the real Codex Session. The picker contained only provider `codex` and its 6 base GPT model families; no Claude model appeared.
- **A -> B -> A switching:** Codex A showed its 2-message transcript and Codex catalog, Claude B showed its 2-message transcript and 5-model Claude catalog, and returning to A restored only A's transcript/catalog. No stale transcript, catalog, empty state, or page error remained.
- **Codex create / prompt / model change / delete:** I created exactly one disposable Session through the UI, selected `gpt-5.4-mini · low`, sent a marker prompt and received the exact marker; changed to `gpt-5.6-luna · low`, sent a second marker prompt and received the exact marker. The native endpoint then contained exactly 4 records and reported `{"providerID":"codex","modelID":"gpt-5.6-luna","variant":"low"}`. UI deletion completed asynchronously and removed that exact Session from both the list and the API.
- **Real Claude no-cursor behavior:** the real Claude list returned 24 Sessions and no next-cursor header. An existing Session opened with its complete 2-message transcript and Claude-only catalog. Filtering to Claude exposed no `Load older Sessions`; switching back to all harnesses restored the Codex older-page action.
- **Persisted-history failure:** I held an existing, previously unopened Codex rollout under an exclusive read lock and opened it through the UI. The transcript showed `Session history could not be loaded` plus the concrete internal read failure, rendered 0 messages, and did not show the valid-empty `Start the conversation` state. I released the lock immediately; after a settled switch away and back, the same Session recovered all 4 persisted messages and the error was gone.

Verification on the rebased head:

- `npm run test:ci:full` in `web`: pass
- `npm run build` in `web`: pass
- full bridge `npm test`: 477/478 locally; the only failure is the unchanged Windows `EPERM` while creating the symlink fixture, before that test body runs. The GitHub Windows bridge job passes.
- GitHub: type/regressions, Windows and macOS bridge tests, Chromium product smoke, macOS application menu, and Debug APK all pass.

The rebase takes upstream PR #386's current Codex rollout parser as its baseline. This branch retains the missing-local-rollout ACP fallback, explicit initial-read error, bounded pagination, delete tombstone, and no-cursor matrix. `adeb2ab` also updates `docs/DEPENDENCIES.md` to describe both current `item_completed` and legacy rollout records.

No transcript content, local paths, or Session identifiers are included in this report.

giuliastro:
Correction: no need to install or live-test harnesses you do not use. The real Codex and Claude validation you reported is useful and addresses the primary regressions we asked you to reproduce. OMP, PI, and OpenCode can be validated separately by other maintainers/testers; please do not install additional harnesses solely for this PR.

dnhrdt:
Understood — thank you for the clarification. I will not install additional harnesses solely for this PR. The unchanged head adeb2ab and the reported real Codex/Claude evidence are ready for formal re-review.

giuliastro:
Thanks for the thorough follow-up and careful validation. Much appreciated.

PR #390, opened 2026-09-08
fix: scope model routing to the selected harness

dnhrdt:
## Summary

This fixes native model routing without making historical ACP Sessions authoritative for picker membership.

- Model discovery always uses the selected harness's current catalog. The shared client boundary now keeps `agentId` and `backend` coherent, including desktop routing when a machine's primary harness differs from the selected one.
- Existing Session model state and per-turn history remain runtime/timeline metadata; they do not widen the selectable catalog.
- Picker discovery never calls the exact-Session `/config/providers` route and therefore does not `session/load` or `session/resume` an inactive Session merely to populate models.
- Current harness catalog variants/reasoning controls remain attached to catalog-resolved models for both prompt and command dispatch.
- Claude catalog normalization exposes a lone canonical `claude-…[1m]` model such as Fable as a stable bare ID when no bare sibling exists. At the actual mutation boundary, `AcpService` maps that stable ID to the exact option advertised by the target Session. Alias-style entries such as `opus[1m]`, and genuine bare/decorated sibling pairs, remain distinct.
- The pinned official Claude ACP adapter moves from `0.63.0` to `0.75.1` (embedded Claude Code `2.1.257`) so the currently advertised Fable model is executable.

This retains the v3 backend-audit boundary from #287: the harness owns current selectable membership, while the Session owns current and historical model state. It also retains #382's bounded pagination behavior unchanged.

## Real-adapter validation

With `@agentclientprotocol/claude-agent-acp@0.75.1`, a fresh prompt-less technical Session advertised:

- `default`
- `opus[1m]`
- `claude-fable-5-1[1m]`
- `sonnet`
- `haiku`

The adapter accepted `claude-fable-5-1` through `session/set_config_option` and returned `claude-fable-5-1[1m]` as the canonical current option. The normalized current catalog therefore exposes the accepted stable ID `claude-fable-5-1`, while mutation still respects the target Session's exact current option spelling.

## Automated verification

- `web`: `npm run test:ci:full` — pass
- `web`: `npm run build` — pass
- `web`: `npm run build:electron` — pass
- focused bridge routing/catalog/variant/adapter tests — 47/47 pass
- `bridge`: 481/482 pass locally; the only failure is the unchanged Windows `EPERM` while creating the symlink fixture, before that test body runs
- `git diff --check` — clean

The pushed head is `a7b3bd0`. All five GitHub checks pass on this head.


dnhrdt:
CI follow-up on current head `384576a`:

The first Chromium run exposed a product-smoke fixture gap rather than a product failure. Its fake daemons still served only the machine-wide `/models` endpoint. The corrected client intentionally requested the exact existing Session's `/config/providers` endpoint, received the fixture's 404, and therefore kept the composer disabled.

Commit `384576a` updates all five native product-smoke daemons to distinguish exact-Session `/config/providers` responses from fresh-Session `/models` responses. The mismatched-primary Codex navigation smoke now also asserts the Session id, directory, Codex URL path, and `X-Harness-Backend: codex` together. No production code changed in this follow-up.

The rerun is fully green: type-check/regressions, bridge tests on Windows and macOS, the complete Chromium product smoke, and the Debug APK build all pass.

giuliastro (changes requested):
Thanks for digging into the real Claude/Codex routing problem and for the extensive validation. The bug is real, but I don’t think the current architectural direction is acceptable as-is.

The key issue is that this PR makes the exact historical ACP Session’s `configOptions` authoritative for model membership. That conflicts with a decision we already made during the v3 backend audit (#287): historical ACP Sessions can retain model options from when they were created, including provider models that have since been removed. We explicitly changed discovery so that persisted/historical Sessions are never used as authoritative model catalogs; ACP discovery starts from a fresh prompt-less technical Session per adapter lifetime, precisely to reflect the harness’s current runtime model catalog.

In other words, Harness Remote should conceptually keep these concerns separate:

- the harness owns the current selectable model catalog;
- the Session owns its current model / per-turn model history;
- a technical ACP Session is only an implementation mechanism for querying the harness when ACP does not expose a better harness-level catalog API.

A historical Session must therefore not become the source of truth for what the picker offers today. If a Session still carries a model that no longer exists, that may be useful as Session state/history, but it should not reappear as a currently selectable model merely because the Session retained it.

I think the real problems you found should be fixed one layer lower instead:

1. Make sure model discovery is always routed to the exact selected harness/backend on the machine (the #381/#383 regression showed why a mismatched `{agentId, backend}` pair is dangerous).
2. Fix ACP catalog normalization so the fresh/current harness catalog exposes IDs that the harness actually accepts (including the Claude Fable decorated-vs-bare ID issue), rather than switching authority to the historical Session.
3. Preserve adapter-advertised variants/reasoning controls from the current harness catalog. The current exact-Session `/config/providers` path also drops the variant metadata that the machine-wide catalog probes and exposes for Codex/PI/OMP.
4. Do not `session/load`/`session/resume` an inactive Session merely to populate the picker. For Codex this can acquire/hold the single-writer lock before the user performs any mutation, violating the Session-first read-only ownership contract.
5. Keep Session model/history information separate from picker membership. If we want to show model changes across a Session, that belongs in the Session timeline/runtime metadata, not in the selectable catalog.

This is also very close to the direction of #381, which we had to revert in #383 after real multi-harness regressions. I would prefer this PR to be substantially reworked rather than merged in its current form. The underlying routing/Claude ID bugs are worth fixing, but without reintroducing Session-scoped catalog authority.

dnhrdt:
Thanks — agreed with the boundary you described. `a7b3bd0` substantially reworks the PR around the current harness catalog instead of the historical Session.

Addressing the five points directly:

1. The model-catalog client now normalizes a selected route to a coherent `{ agentId, backend }` pair at the shared request boundary. A desktop regression starts from a Claude-primary profile and verifies that selecting Codex sends both the Codex path and Codex backend route.
2. ACP catalog normalization now presents a lone canonical Claude `claude-…[1m]` option as the stable bare ID when there is no bare sibling. A real 0.75.1 adapter probe confirmed that `claude-fable-5-1` is accepted and canonicalized back to `claude-fable-5-1[1m]`. At mutation time only, the service maps the stable picker ID to the exact option advertised by the target Session.
3. Prompt and command dispatch again resolve through the current catalog and retain its adapter-advertised variant config/value. The raw/exact-Session variant fallback is gone.
4. Native Session pickers use only `/v1/agents/:agent/models`. All `/config/providers` picker routing and its browser fixtures have been removed, so opening an inactive Session does not load/resume it for discovery or acquire its writer lock.
5. Historical/current Session model information stays in runtime/timeline metadata and never contributes picker membership.

The pinned Claude adapter update remains. Locally, the full web suite and both builds pass; focused bridge coverage is 47/47. The full bridge run is 481/482, with only the unchanged Windows symlink-fixture `EPERM` before that test body starts. I also replaced the PR title/body so they no longer describe the rejected exact-Session architecture.

Please re-review the current head when convenient.

giuliastro (approved):
Re-reviewed current head `a7b3bd0`. This substantially addresses the architectural concerns from the previous review.

The picker is back on the selected harness's current `/v1/agents/:agent/models` catalog; historical Session config no longer contributes selectable membership, and opening an inactive Session no longer needs `/config/providers` discovery or an ACP load/resume. The shared catalog boundary now keeps the selected agent/backend coherent, including the Claude-primary -> Codex desktop regression. Prompt/command dispatch again resolves model variants through the current harness catalog, preserving adapter-advertised reasoning controls.

The Claude Fable handling is now scoped to catalog normalization/mutation translation: the picker exposes the stable bare id while `AcpService` maps it back to whichever exact value the running target Session advertises. The tests preserve genuine bare/decorated sibling pairs and alias-style entries such as `opus[1m]`, so this does not collapse distinct selectable lanes.

CI is green on head across Windows/macOS bridge tests, type-check/regressions, Chromium product smoke, and Debug APK. The Claude ACP pin remains the main broader dependency change, but it is tied to the reproduced Fable compatibility issue and has direct adapter validation in this revision.

Approved. I would still keep the historical-Session/current-catalog boundary covered by the new regression tests going forward.

PR #492, opened 2026-09-13
feat(bridge): detect mimocode as an OpenCode-compatible backend

MattiaPARRINELLO:
## Why

[mimocode](https://mimocode.dev) is a native fork of OpenCode and speaks the same HTTP server contract: `mimo serve` answers `/global/health` and `/session` exactly like `opencode serve`, so the managed OpenCode host (`ManagedOpenCodeHost`) drives it unchanged.

The launcher only missed it because detection matches an executable literally named `opencode`, while the fork installs as `mimo`. On such a machine `detectBackends()` returns `[]` and startup fails with:

```
No supported agent CLI was found on PATH. Install/select omp, pi, claude, codex, or opencode, then re-run with --backend if needed.
```

even though the agent is installed. The workaround was to pass `--backend opencode` *and* `HARNESS_REMOTE_OPENCODE_COMMAND=<path-to-mimo>` by hand.

## What

- `BACKEND_EXECUTABLES.opencode` now lists `["opencode", "mimo"]`, so `mimo` satisfies the existing `opencode` backend.
- New `resolveOpenCodeCommand()` picks whichever alias is actually on `PATH` (upstream name first) and is used both by the machine daemon and the single-host path, instead of the hardcoded `"opencode"` default.
- `HARNESS_REMOTE_OPENCODE_COMMAND` and `--opencode-command` keep priority.

The `opencode` id, labels and HTTP model catalog are unchanged; mimocode just becomes another executable that can back them.

## Verification

- `cd bridge && node --test` → **494 passing, 0 failing** (2 new tests: detection via `mimo`, and command resolution preferring `opencode` then falling back to `mimo`, with env override).
- Against a real harness: `mimo serve --hostname 127.0.0.1 --port 4099` returned `GET /global/health` → `{"healthy":true,"version":"local"}` (HTTP 200) and `GET /session` → HTTP 200, i.e. the exact contract `ManagedOpenCodeHost` polls.

Not a network-service change: no auth, `--root`, or CORS behavior touched.


MattiaPARRINELLO:
Deleted for reopening with updated code (2 clean commits, isOpenCodeLike helper).

PR #494, opened 2026-09-14
feat(bridge+web): add mimocode backend support + fix external session detection

MattiaPARRINELLO:
## Why

Mimocode is an OpenCode fork with an identical `/global/health` and session API. This PR adds first-class support so Harness Remote detects mimocode automatically and the Android app correctly tracks sessions started from an external terminal.

### fix(web): detect external busy sessions in refreshStatus
- `refreshStatus()` in native-session-v3-adapter.ts now fetches `?directory=` (mimo's `/session/status` is directory-scoped — without it the response is `{}`)
- Adopts externally busy sessions (sets `forcedStatus = "running"` even when the app didn't start the turn)
- New test: `native-session-external-status.test.mjs` (busy adopt + idle settle)

### feat(bridge+web): add mimocode backend support
- **Bridge**: `BACKEND_EXECUTABLES` adds `"mimo"`, `resolveOpenCodeCommand()` prefers opencode then mimo, new "mimocode" harness profile, dynamic managed backend detection
- **Web**: `BackendKind` adds `"mimocode"`, `isOpenCodeLike()` helper, whitelist updates across ~15 files, mimocode.svg icon, i18n (en/it/zh)
- **Retrocompatible**: if the real `opencode` is detected, everything works as before

## How to verify
- `cd bridge && node --test` → 494 pass
- `npx tsc -b` → 0 errors
- `node --experimental-strip-types web/src/session-first-regression.test.mjs` → pass
- `node --experimental-strip-types web/src/native-session-external-status.test.mjs` → 2/2 pass
- Tested on real mimo serve (daemon mode, `/v1/machine` returns `{id:"mimocode",backend:"mimocode"}`)

## Test env
- Garuda Linux (Arch), mimo `~/.mimocode/bin/mimo` (fork OpenCode, version "local")
- daemon `--backend claude` → Claude ACP + managed mimocode


MattiaPARRINELLO:
All 6 Copilot comments addressed in b80dbcf2:

1. **Electron allowlists** (profile-registry, request-transport, event-transport) — added `mimocode` to all 3 backend sets
2. **Double detection** — removed `mimo` from `BACKEND_EXECUTABLES.opencode` (now only in `mimocode`)
3. **Backend-aware command** — `resolveOpenCodeCommand()` accepts `{backend}` hint, prefers `mimo` when `backend=mimocode`
4. **machinePayload agentId** — uses `config.backend` instead of hardcoded `"opencode"`
5. **cancelled guard** — busy adoption now preserves `forcedStatus === "cancelled"`
6. **Launcher test** — updated to expect `["mimocode"]` (not `["opencode","mimocode"]`) when only mimo is present

All tests pass: bridge 494, tsc, session-first-regression, settings-regression, external-status 2/2, machine-payload, i18n.

MattiaPARRINELLO:
Re-review requested after addressing all Copilot review points:

**Done (commit e435fb3b):**
1. ✅ Dynamic label —  now prints 'Mimocode is ready' or 'OpenCode is ready' based on backend
2. ✅ BACKEND_KINDS consolidation — single  constant in , imported by all 6 whitelist locations (electron: profile-registry, request-transport, event-transport; web: taskClient, native-session-discovery, work-thread-conversation)
3. ✅ Cancelled guard test —  now covers 
4. ✅ Directory doc —  semantics documented in test header
5. ✅ Detection order test —  covers both-detected, single-mimocode, and upstream-first priority
6. ✅ Electron allowlist test —  validates all backends accepted by profile registry

**Test results:** bridge 498 pass, tsc clean, external-status 4/4, electron allowlist 3/3

giuliastro (dismissed):
Grazie, il supporto Mimocode è ben avviato e gli ultimi commit hanno risolto i rilievi Copilot principali. Prima di integrare restano però tre blocchi:

1. **Rebase sull'integration branch.** Ho retargettato la PR a `codex/development-2026-09-11`; GitHub la segnala ora `mergeable_state: dirty`. Va ribasata/aggiornata su `b22db691dd9183a69ee76f04b0f462c16efa21fd` preservando le recenti fix Native Session/adapter già presenti lì.

2. **`--opencode-command` viene ancora ignorato nel daemon.** `parseDaemonOptions()` continua correttamente a valorizzare `openCodeCommand` da CLI/env, ma `main()` lo rinomina `defaultOpenCodeCommand` e poi ricalcola sempre `resolveOpenCodeCommand(process.env, { backend: managedBackend })`. Quindi `harness-remote-daemon --opencode-command /custom/mimo` può avviare un binario diverso da quello esplicitamente richiesto. L'override esplicito deve restare autoritativo; la risoluzione OpenCode/Mimo va usata solo come default automatico.

3. **Non reintrodurre `/session/status` nel normale idle/pre-Send OpenCode.** La PR #351 ha intenzionalmente escluso quel read dal path ordinario perché può essere lento/inaffidabile e ritardare il primo Send. Il nuovo `refreshStatus()` ora chiama `api.listStatuses(...)` anche per una Session OpenCode-like interna e idle. L'adoption del busy esterno è utile, ma va limitata alle Session realmente esterne (ad es. `entry.target.external`, o stato equivalente) mantenendo l'early-return storico per le normali Session interne quando non sono `forcedStatus === "running"` e non c'è recovery watch. Aggiungerei un contratto che provi entrambe le cose: external idle/busy legge status con `directory`, internal idle/pre-Send non chiama affatto `listStatuses`.

Quando questi punti sono risolti, rifaccio la review sul branch di integrazione e facciamo passare l'intera CI (Chromium, desktop e Debug APK) prima del merge.

giuliastro (dismissed):
Thanks, the Mimocode support is off to a good start and the latest commits have addressed the main Copilot findings. Before integration, three blockers remain:

1. **Rebase onto the integration branch.** I retargeted the PR to `codex/development-2026-09-11`; GitHub currently reports it as `mergeable_state: dirty`. Please rebase/update it onto `b22db691dd9183a69ee76f04b0f462c16efa21fd`, preserving the recent Native Session/adapter fixes already present there.

2. **`--opencode-command` is still ignored by the daemon.** `parseDaemonOptions()` still correctly captures `openCodeCommand` from CLI/env, but `main()` renames it to `defaultOpenCodeCommand` and then always recomputes `resolveOpenCodeCommand(process.env, { backend: managedBackend })`. This means `harness-remote-daemon --opencode-command /custom/mimo` can start a different executable from the one explicitly requested. The explicit override must remain authoritative; OpenCode/Mimo resolution should only be used as the automatic default.

3. **Do not reintroduce `/session/status` into the normal idle/pre-Send OpenCode path.** PR #351 intentionally excluded that read from the ordinary path because it can be slow/unreliable and delay the first Send. The new `refreshStatus()` now calls `api.listStatuses(...)` even for an internal, idle OpenCode-like Session. External busy adoption is useful, but it should be limited to genuinely external Sessions (for example `entry.target.external`, or equivalent explicit state), while preserving the historical early return for normal internal Sessions when `forcedStatus !== "running"` and no recovery watch is active. I would add a contract that proves both sides: external idle/busy reads status with `directory`, while internal idle/pre-Send OpenCode does not call `listStatuses` at all.

Once these points are resolved, I will re-review the PR on the integration branch and we should run the full CI gate (Chromium, desktop, and Debug APK) before merge.

giuliastro (changes requested):
Thanks. I have retargeted this PR back to `main`. Your branch is already based on the current `main` (`21ce6db49af708c4c7c3f96ef6a50f62dced8dab`), so **no rebase is needed** at this point.

Two blockers remain before this can be considered for merge:

1. **Preserve the explicit `--opencode-command` override.** `parseDaemonOptions()` still captures the CLI/env override, but `main()` renames it to `defaultOpenCodeCommand` and then recomputes `resolveOpenCodeCommand(process.env, { backend: managedBackend })`. That can start a different executable from the one explicitly requested. The explicit CLI/env value must remain authoritative; OpenCode/Mimocode command discovery should only be the automatic default when no override was supplied.

2. **Do not reintroduce `/session/status` into the normal internal idle/pre-Send OpenCode path.** PR #351 deliberately removed that read from ordinary idle OpenCode reconciliation because the endpoint can be slow/unreliable and delay prompt delivery. The current `refreshStatus()` change calls `api.listStatuses(...)` for an internal idle OpenCode-like Session as well as an external one. External busy-session adoption is useful, but it must be scoped to genuinely external Sessions (for example `entry.target.external`, or equivalent explicit state), while preserving the historical early return for normal internal OpenCode Sessions when `forcedStatus !== "running"` and no recovery watch is active.

Please add executable regression coverage proving both sides of that lifecycle boundary: an external OpenCode/Mimocode Session reads directory-scoped status and can adopt busy state, while a normal internal idle/pre-Send OpenCode Session does **not** call `listStatuses` at all. Please also keep the existing OpenCode recovery semantics from #351 intact.

Once these are fixed, we can validate the PR independently against `main` with the full regression/Chromium/desktop/Debug APK gate. If it is accepted there, we will handle integration into the separate development branch on our side.

MattiaPARRINELLO:
Addresses the two blockers from your review:

**1. Preserve explicit `--opencode-command` override**
`parseDaemonOptions` now tracks `openCodeCommandExplicit` (set to `true` when `--opencode-command` CLI flag or `HARNESS_REMOTE_OPENCODE_COMMAND` env is provided). In `main()`, the explicit value is used as-is instead of being overwritten by `resolveOpenCodeCommand(process.env, { backend: managedBackend })`. Regression tests added in `daemon-cli.test.js` proving the flag is set correctly for CLI, env, and auto-resolved cases.

**2. Scope `/session/status` to external sessions only**
Restored the #351 early-return guard: `if (openCode && !entry.target.external && entry.forcedStatus !== "running" && !openCodeRecoveryWatchActive) return`. The directory-scoped `listStatuses` call now only happens for genuinely external sessions (`entry.target.external === true`). Added a regression test proving an internal idle/pre-Send OpenCode session does **not** call `listStatuses` at all. The test is wired into `test:ci:full`.

All bridge tests (501/501) pass. Web tests: `session-first-regression`, `native-session-external-status` (5/5 including new test), `native-session-observer`, and `native-session-model-lifecycle` all pass.

giuliastro:
I found one additional reproducible blocker while validating the updated PR locally.

`web/src/serverConfig.ts` now imports a runtime value from `./types.js`:

```ts
import { isOpenCodeLike, type ServerConfig } from "./types.js"
```

On `main`, that import is type-only:

```ts
import type { ServerConfig } from "./types.js"
```

The difference matters because `test:config` runs directly under Node with `--experimental-strip-types`. With the new runtime import, Node tries to resolve `web/src/types.js`, but the source tree only contains `types.ts`, so `npm run test:ci:full` fails immediately with:

```text
Error [ERR_MODULE_NOT_FOUND]: Cannot find module '.../web/src/types.js' imported from .../web/src/serverConfig.ts
```

I reproduced this on Node 22.23.2, matching the repository CI major version. The production build still passes because Vite resolves the source graph differently.

The diagnosis is also confirmed by running the same regression test through the existing Vite-based test runner:

```bash
cd web
node scripts/run-vite-test.mjs src/server-config.test.mjs
```

which passes successfully.

Recommended fix: keep `serverConfig.ts` compatible with the existing Electron/NodeNext `.js` import convention, but run `test:config` through the repository's existing Vite test runner instead of invoking the test directly with Node strip-types. For example, update the `test:config` script to use `scripts/run-vite-test.mjs` for `src/server-config.test.mjs`.

Please re-run the full web regression suite after the change (`npm run test:ci:full`). I would keep this PR blocked until that suite is green.

MattiaPARRINELLO:
Addresses the `test:ci:full` blocker.

`serverConfig.ts` now imports `isOpenCodeLike` (runtime) from `./types.js`, but `test:config`, `test:profiles`, and `test:machine-payload` ran under `node --experimental-strip-types` which cannot resolve `.ts` → `.js` — causing `ERR_MODULE_NOT_FOUND`.

**Fix:** Switched those three scripts to the existing Vite-based test runner (`scripts/run-vite-test.mjs`) which handles the source graph correctly. Production builds are unaffected (Vite resolves the graph differently).

**Verified:**
- `npm run test:ci:full` — all green
- `cd bridge && node --test` — 0 failures
- `npm run tsc -b` — clean (unchanged)

Commit: `8e8ced5a`

giuliastro:
Thanks for the work on this PR and for addressing the previous review findings.

After testing the latest version on a real machine with both OpenCode and Mimocode installed, I’m going to close this PR without merging it.

The main reason is that the core product goal is not actually achieved in the normal multi-harness setup. With both `opencode` and `mimo` available on `PATH`, the launcher detects both, but the daemon registers only one OpenCode-like managed backend and always prefers OpenCode. As a result, Mimocode does not appear as an available or active harness in Harness Remote.

We also tested the external-session behavior. The implementation is useful in the narrower case of a terminal attached to the same managed OpenCode/Mimocode server, but it does not provide general live synchronization for externally-running sessions, so the resulting behavior is more limited than the PR title and description initially suggested.

There are still useful ideas in this work — notably the OpenCode-like abstraction, Mimocode detection, and some of the status handling — and we may reuse those concepts in a future implementation.

However, reaching the intended behavior would now require another architectural iteration rather than a small follow-up fix, and I don’t want to keep asking for incremental corrections and repeated manual validation on this PR.

Thanks again for the contribution and the time spent iterating on it.

Issue #372, opened 2026-09-06
Roadmap P3 — ACP Provider Kit and harness ecosystem expansion


giuliastro:
## Why

Harness Remote should support more coding agents without turning every adapter into a bespoke product project or flattening harness-specific capabilities into fake common behavior.

The scalable path is a small ACP-oriented provider kit plus explicit capability extensions.

### ACP Provider Kit

Define reusable adapter primitives and contract tests for the common ACP lifecycle:

- detection / startup;
- Native Session list / lookup / create / load-resume;
- prompt dispatch and streaming updates;
- Stop/cancel;
- questions / permissions when advertised;
- model and variant/config discovery;
- reconnect/restart reconciliation;
- diagnostics and bounded resource ownership.

The kit must allow harness-specific extensions. It must not force every agent into capabilities it does not actually advertise.

### Capability-driven integration

- Keep `V3_HARNESS_CAPABILITY_MATRIX.md` as the explicit runtime contract.
- Separate generic transport/lifecycle behavior from per-harness reconciliation rules, aligned with #330.
- Add reusable adapter contract fixtures and real-harness validation gates.
- Make adding a standards-compatible ACP harness primarily a profile/adapter task rather than a generic UI rewrite.

### Expansion order

Investigate and integrate in this order unless real-world protocol/support constraints change:

1. GitHub Copilot CLI / ACP;
2. Cursor CLI / ACP;
3. Google Antigravity CLI if its stable control/session surface is suitable;
4. generic/community ACP integrations such as Kimi Code CLI;
5. additional community agents (for example Hermes Agent) after the provider-kit contract is proven.

Each addition must document which capabilities are implemented, which are advertised by the running harness, and which were verified against a real installation.

### Parallel work after federation is stable

Add explicit opt-in worktree/isolated parallel work only after the Native Session federation model is stable. Always expose path, branch/worktree and lifecycle; never silently move normal work into hidden daemon-managed worktrees.

## Exit criteria

- A new standards-compatible ACP harness can be added without modifying generic conversation semantics or adding harness-name conditionals throughout the UI.
- Adapter contract tests catch create/resume/stream/Stop/model/reconnect regressions before merge.
- Existing OpenCode, Codex, Claude, OMP and PI behavior remains unchanged while shared ACP infrastructure is extracted.
- At least one new major harness is shipped through the provider-kit path and passes real-harness validation.
- Harness-specific capability differences remain visible and honest.

## Related

#330

## Depends on

#368
[...]

PR #504, opened 2026-09-14
daemon: support custom ACP command for built-in primary with multiple detected CLIs

monperrus:
A custom primary passed via `--acp-command` is not part of the launcher executable catalog. Once two built-in ACP CLIs are installed, resolving the daemon launch plan rejected that custom `--backend` value and prevented startup.

Resolve built-in ACP hosts independently when the configured primary is custom, while retaining normal launcher resolution for built-in primaries.

Adds a regression test for a custom primary with Claude and Codex detected.


giuliastro (changes requested):
Thanks — the underlying startup problem is plausible, but this patch does not yet exercise or enable the real configuration path it claims to support.

Two blocking issues:

1. `parseConfig()` still resolves every `--backend` through `harnessProfile(value)`. A value such as `agentknit` therefore throws `Unsupported backend` before `resolveDaemonPlan()` can ever see it, even when `--acp-command` is supplied. The new unit test calls `resolveDaemonPlan()` directly with a custom backend and bypasses this real parser boundary.

2. Even if a custom backend somehow reached daemon startup, `main()` still executes `harnessProfile(config.backend)` and the ACP host loop also calls `harnessProfile(backend)`, so startup would still fail before registering the custom primary.

Please either define the intended supported contract as a built-in backend with a custom `--acp-command` override (and test that real `parseDaemonOptions()`/startup path), or add a coherent custom-primary profile/config model end-to-end. In either case the regression should exercise the actual daemon configuration/startup boundary rather than only the new helper.

I retargeted this PR to `codex/development-2026-09-11` so development never lands directly on `main`. Please update onto that integration branch before re-review.

giuliastro:
Post-v3.1.0 cleanup: retargeted this PR from the retired 3.1 integration branch to `main`. The functional change is still relevant and narrowly scoped (custom `--acp-command` primary + multiple detected built-in ACP CLIs). The previous red Debug APK job was unrelated to this patch: it failed because `android-actions/setup-android@v4` attempted to install the retired Android SDK `tools` package; the code/test jobs were green.

giuliastro:
Final post-release review: retargeting to `main` only removed the obsolete integration base; this PR is **not merge-ready yet**. The existing CHANGES_REQUESTED review is still valid on current `main`: `parseConfig()` still resolves a custom backend through `harnessProfile()`, and daemon startup still calls `harnessProfile(config.backend)`, so the patch's direct `resolveDaemonPlan()` test bypasses the end-to-end path that currently rejects the claimed custom primary. Keeping #504 open. It needs either (a) a clearly supported built-in backend + custom `--acp-command` contract tested through the real daemon configuration/startup boundary, or (b) a coherent custom-primary profile/config model end-to-end.

giuliastro:
Maintainer decision after the post-v3.1.0 review: keep this PR open, but narrow its contract to the small compatibility fix.

### Scope for #504

Support a custom `--acp-command` **only when `--backend` is an existing Harness Remote backend/profile** (for example `codex` or `claude`), including the case where multiple built-in ACP CLIs are detected.

This PR should **not** introduce arbitrary backend IDs such as `agentknit` and should not add a second generic provider/profile mechanism.

### Acceptance criteria

- A real configuration path such as `--backend codex --acp-command <custom executable>` parses successfully.
- Multiple detected built-in ACP CLIs do not cause the explicitly selected primary/custom executable to be rejected or replaced.
- The regression test exercises the real `parseDaemonOptions()` / daemon startup seam, not `resolveDaemonPlan()` in isolation.
- The startup path continues to use the existing built-in `harnessProfile(config.backend)` contract; no `harnessProfile(<arbitrary custom id>)` path is required here.
- Existing daemon behavior for normal built-in primaries remains unchanged.

True arbitrary/custom ACP provider IDs belong to #372 (ACP Provider Kit), where profile/config/capability/lifecycle support can be designed coherently end-to-end.

The existing `CHANGES_REQUESTED` review therefore remains in force until this narrower path is implemented and tested.

giuliastro:
Closing without merge.

After narrowing the intended contract to an existing Harness Remote backend/profile plus a custom `--acp-command`, current `main` already supports the behavior: `parseConfig()` preserves the explicit ACP command, `resolveLaunchPlan()` accepts the selected built-in primary with multiple detected CLIs, and daemon startup uses `config.acpCommand` for that primary.

PR #532 adds a regression test for that supported path without changing production runtime code. Its Bridge tests on Windows/macOS and Type-check/regressions are already green.

The original broader behavior in this PR — arbitrary backend IDs such as `agentknit` — is intentionally not being merged through this launcher bypass. That belongs to #372 so profile/config/capabilities/lifecycle are added coherently end-to-end.

PR #512, opened 2026-09-15
fix(opencode): reconcile Session rail lifecycle after navigation

giuliastro:
## Problem

A real OpenCode regression exists in the Session rail:

1. Session A is Working.
2. The user opens Session B before A finishes.
3. A completes normally, but its rail row can remain Working indefinitely.
4. Reopening A immediately shows the completed reply and changes the row to Ready.

The stale presentation bridge introduced by `fc851bbd` is still needed to prevent a false Working ↔ Ready flicker across navigation, so simply clearing `presentationOverrides` would reintroduce the original defect.

## Root cause

The workspace live stream refreshes machine discovery, but Session lifecycle is not part of `/v1/machine`. `reuseList()` therefore correctly preserves structurally identical runtime identity, which means `NativeSessionHome` can keep the same `sources` identity and not re-read the Session index after a background turn completes.

OpenCode lifecycle events and `/session/status` are also separate reads, so an idle edge can briefly arrive before the lightweight status endpoint has converged.

## Fix

- publish a client-side Session-index invalidation revision only for bounded lifecycle events;
- keep token chunks (`message.part.delta`) out of global Session discovery;
- on the stable 3.0 shell, make the next structural workspace reconciliation publish one fresh list identity after such an out-of-band invalidation, without changing `MachineSnapshot`, server config, or daemon payloads;
- preserve fresh streamed `session.status` / `session.idle` for a bounded 15s grace window so a just-observed idle edge beats a briefly stale `busy` read;
- clear transient stream authority and invalidate discovery on reconnect;
- keep the existing `presentationOverrides` bridge and 30s discovery fallback.

## Safety boundaries

This does **not** change:

- the public `/v1/machine` / `MachineSnapshot` contract;
- OpenCode prompt/Send routing;
- the #351 rule that ordinary idle/pre-Send OpenCode must not depend on `/session/status`;
- transcript reconciliation or reply-settle semantics;
- Stop behavior;
- ACP writer ownership;
- polling cadence.

The live status override is bounded; durable Session discovery becomes authoritative again after the grace window.

## Regression coverage

Executable coverage proves:

- token chunks preserve structurally identical workspace sources;
- a Session lifecycle edge forces exactly the next stable workspace reconciliation to publish fresh sources;
- the invalidation subscriber observes the already-updated streamed status;
- OpenCode `session.status` normalization preserves `busy`/`idle`;
- a fresh streamed `idle` beats a briefly stale `busy` read;
- live status authority expires rather than pinning a row forever;
- reconnect clears transient status authority and invalidates Session discovery;
- Chromium reproduces the original product flow and proves Session A becomes Ready while Session B remains selected, before A is reopened.

## CI maintenance

Validation also exposed an unrelated repository CI break: `android-actions/setup-android@v4` currently defaults to the retired Android SDK `tools` package. The workflow skips that default package set and still installs the exact Android 36 platform/build-tools explicitly, so the signed Debug APK gate can execute again.

No merge should happen until the complete PR CI is green.


giuliastro:
Closing as superseded. The same Session-rail lifecycle fix was ported to the 3.1 integration line in #513 and merged as commit `0dadac745aa346caf4e9c57e2c1ed857e238303a`. The resulting behavior is included in Harness Remote v3.1.0, whose release notes explicitly include stable Session rail transitions between Working, Attention and Ready.

Issue #523, opened 2026-09-19
Mobile onboarding: simplify first-machine QR setup and post-scan flow


giuliastro:
## Problem

The first-run machine setup flow in the Android APK is too tortuous and not self-explanatory enough.

Observed manually on RC5:

1. On first launch, it is not obvious how to get to the QR scanner to add the first machine.
2. After successfully scanning a machine QR code, the "new machine" form remains open with the scanned machine data.
3. This leaves the user in an ambiguous state: it looks like setup is still incomplete even though the machine has already been added.

## Expected UX

For first-run onboarding:

- Make the primary action to connect the first machine obvious.
- QR scan should be a first-class action, not something the user has to discover indirectly.
- After a successful QR scan/add:
  - close/dismiss the machine creation form automatically;
  - show that the machine was added successfully;
  - offer a clear path to **Sessions**;
  - optionally expose a secondary action to **Add another machine / Scan another QR**.

For subsequent machine management:

- Do not keep the editable "new machine" form open after a successful add unless the user explicitly chooses to add another machine.
- The normal post-success state should be the machine list / workspace, with a clear **Add machine** or **Scan QR** action.

## Acceptance criteria

- Fresh APK install: user can understand how to connect the first machine without prior knowledge.
- QR scanner is directly reachable from the first-run setup surface.
- Successful scan never leaves the user stranded in the creation form.
- User can immediately continue to Sessions after adding the first machine.
- User can explicitly choose to add/scan another machine.
- Existing machine editing flow remains unchanged.
- Verify on physical Android device, including back navigation after scan.

## Context

Found during manual RC5 validation. Functional connectivity works; this is an onboarding/state-transition UX issue rather than a transport or pairing failure.

giuliastro:
Resolved by #524 (`fix(android): simplify first-machine QR onboarding`) and shipped in Harness Remote v3.1.0. The released flow is scan-first, successful pairing closes the stale setup state, and the user gets an explicit **View sessions** action.

PR #524, opened 2026-09-19
fix(android): simplify first-machine QR onboarding

giuliastro:
Fixes #523.

## What changes

- exposes **Scan machine QR code** directly in the first-run Machines surface instead of requiring the manual editor first;
- keeps **Add manually** as a separate fallback;
- after a successful QR/manual add, shows an explicit completion state with **View sessions** as the primary next step;
- keeps **Scan another machine** / **Add machine** as secondary continuation actions;
- dismisses stale manual-machine draft/confirmation state after successful QR pairing;
- localizes the Android pairing status instead of showing hardcoded English;
- completes i18n parity across English, Italian, Traditional Chinese and Simplified Chinese;
- adds a regression guard that fails if any supported language is missing an English translation key;
- adjusts narrow Android layout so the completion and pairing actions remain usable.

## UX rationale

Pairing is treated as an onboarding step rather than the destination. After a successful add, Harness Remote confirms completion and presents an explicit path into Sessions instead of relying on the user discovering the bottom navigation or making machine rows ambiguously clickable.

This follows the established pattern used by mature device-onboarding products: complete the pairing/setup step, then offer a clear primary transition into the product while keeping “add another” secondary.

## Scope

Frontend / Android onboarding and i18n only. No bridge, ACP, daemon, Session routing or protocol changes.

## Validation target

Run the normal PR gates including i18n/regressions, Chromium product smoke and Debug APK. After CI is green, validate the generated APK on a physical Android device.


giuliastro:
CI on head `2f04c0d976a3763b793c7d1eb40f7a062178e93b` is green: type-check/regressions, bridge macOS/Windows, desktop Ubuntu/macOS/Windows, Chromium product smoke, OpenCode live Zen and Debug APK all passed.

Physical Android validation is intentionally still pending before merge/freeze. Artifact: `harness-remote-debug-v3.1.0-2f04c0d` from workflow run 35427341009.

giuliastro:
Updated Android onboarding after physical feedback.

Head: `86422a372cc86c69942561856124a6d48343a6f9`

Changes now include:
- explicit post-add **View sessions** primary CTA;
- Scan another/Add machine kept secondary;
- Android pairing status localized instead of hardcoded English;
- complete i18n key parity for en/it/zh-TW/zh-CN, with a regression test that fails on missing translations;
- previously missing Traditional Chinese command/connect translations and Simplified Chinese attachment labels filled.

Validation:
- type-check/regressions ✅
- bridge macOS/Windows ✅
- desktop Ubuntu/macOS/Windows ✅
- OpenCode live Zen ✅
- Chromium product smoke ✅ on same-SHA rerun (first attempt hit a transient cross-machine Project-selection assertion; rerun passed the full suite including that exact test)
- Debug APK ✅

APK artifact: `harness-remote-debug-v3.1.0-86422a3`, workflow run `35429357567`.

PR remains open pending physical Android validation.

PR #529, opened 2026-09-19
fix(npm): make first publish registry-safe

giuliastro:
Follow-up from the first npm publish attempt.

The workflow correctly proved that the package is not yet on npm and that this repository currently has no NPM_TOKEN. It also exposed npm registry normalization that would have removed the CLI bins.

This fixes that before any first publish:
- removes leading ./ from published bin paths;
- normalizes repository.url;
- strengthens the package contract test;
- makes the publish workflow fail early with the exact bootstrap requirement;
- keeps README on the working GitHub npx command until the npm package actually exists.


giuliastro:
Post-v3.1.0 cleanup: retargeted this PR from the old integration branch to `main`. The previous CI failure was caused by an unchanged daemon-bin regression still expecting the pre-normalization `./bridge/src/daemon-bin.js` value; that test is now aligned with the registry-safe package path. The package-contract test itself was already passing.

giuliastro:
Closing as superseded by #530 and #531, which are already merged into `main`. #530 contains the npm-11-safe bin normalization, repository metadata normalization and the daemon-bin/package contract updates; #531 replaces the publish flow with release-tag-driven Trusted Publishing and explicitly documents the one-time bootstrap requirement. Keeping #529 open would now reintroduce a competing older publish-flow variant and README state.

PR #539, opened 2026-09-21
fix(web): carry the live image capability into the routed agent

luccapinto:
## What this fixes

On the v3 Session-first surface the composer never offers the image picker — on any harness, including an OMP that advertises `promptCapabilities.image`. #121 shipped that control for 2.x and every piece of it survived the rewrite: `attachments.ts` is byte-identical, the composer still renders `accept="image/*"`, the daemon still validates and still forwards ACP `image` blocks. What did not survive is the capability reaching the gate.

## The trace

v3 has two capability sources, and only one of them can answer this question:

| Source | Origin | Has `attachments`? |
| --- | --- | --- |
| `/v1/machine` → `agents[].capabilities` | `harness-profiles.js`, a static table | **no** |
| `/v1/agents/<id>/v1/capabilities` | the ACP `initialize` handshake | **yes** |

Measured against a 3.1.0 daemon:

```
GET /v1/machine                     → 'attachments' in agents[].capabilities === false   (omp and claude)
GET /v1/agents/omp/v1/capabilities  → { …, "attachments": true }
```

`native-session-observer.tsx` fetches the second one, keeps it in `attachmentsSupported` and synthesizes an agent that carries it. `routableRoutes` then uses that synthesized agent only when the current agent is **absent** from the machine snapshot:

```ts
const current = available.some((candidate) => candidate.id === target.agentID)
  ? available                                       // ← the live answer is discarded here
  : [agent, ...available.filter((candidate) => candidate.id !== target.agentID)]
```

In the ordinary case the agent *is* listed, so `available` wins. `work-thread-conversation.tsx:404` takes `destinationAgents` from that same routed list, and line 1002 reads `attachments: undefined` → `attachmentsSupported` is false → the picker is never rendered. The fetch happens, answers `true`, and is dropped one line later.

`v3-v2-parity.test.mjs` asserted that the observer *calls* `api.capabilities(target.config)`. That assertion stayed true the entire time — which is exactly the gap the regression slipped through.

## The change

Merge the live value into the current agent's entry instead of taking the snapshot verbatim. Six lines, no protocol change, nothing touched in `bridge/`.

The gate stays per-harness and stays driven by the handshake: a harness that does not advertise `promptCapabilities.image` still reports `attachments: false` from the same endpoint and still gets no control. That derivation is unchanged and already covered by `bridge/test/server.test.js`, which exercises a fake with and without the capability.

## Test

One assertion added to `v3-v2-parity.test.mjs`, on the delivery rather than on the call:

```
before the fix:  not ok 7 - Session-first restores capability-gated image attachments…   (7 pass, 1 fail)
after the fix:   8 pass, 0 fail
```

## Verified against a real harness

`omp acp` 18.0.4 behind a 3.1.0 machine daemon on Ubuntu / Node 22 — not only the fakes:

- **before:** `document.querySelectorAll('input[type=file]').length === 0` on a live OMP Session, while `/v1/agents/omp/v1/capabilities` answered `attachments: true` in the same page load;
- **after:** `input[type=file][accept="image/*"]` plus the `Attach image` button, checked at 420px and 1280px;
- a 64×64 solid magenta PNG attached from the composer with the prompt *"reply with only the colour name"* → OMP answered **`Magenta`**. That is the only proof that matters: the image reached the model;
- the same flow from an Android phone against the same daemon — image delivered, described correctly by the model. The upload came back 849×1568, so the client-side resize is doing its job;
- `npm run test:v3-product-ui` 107 pass, `test:native-session-controller` pass, `tsc -b && vite build` clean.

Paperclip restored in the mobile composer, left of the send button

On the machine used for this, both `omp` and `claude` advertise image support through their adapters, so both get the control. Nothing in this change decides that — the harness does.

## Note

The behaviour is unreachable in any client built from current `main`, including the 3.1.0 Android APK, so this needs a release to reach phones. The bridge side needs nothing.


gervaso-assistant:
Thanks for catching this and for the focused fix and validation. You were right: image attachments had not made it through the v3 Session-first rewrite, even though the underlying transport support was still present. I added a behavioral regression test for the live capability routing and merged PR #539.

